PR #119, opened 2015-01-26
First steps for the librdkafka-based classes

yungchin:
(I thought I'd offer these up before we start renaming "samsa" everywhere :))

The most basic functionality on these classes is in place (refresh metadata, consume from a topic - although not yet at the expected offset! -, produce) but clearly there's more to come. What's needed soon is that we hook up the rd_kafka module in builds, which will allow us to add integration tests.


kbourgoin:
This looks fine. Better to get this merged asap so we can start renaming everything and start working in smaller increments.

PR #125, opened 2015-02-05
(Re)define produce() method

yungchin:
As discussed informally, this is a simple approach to enabling both keyed and unkeyed message produce-requests.

### DIFF
--- a/kafka/base.py
+++ b/kafka/base.py
@@ -152,4 +152,10 @@ class BaseProducer(object):
 
     @abc.abstractmethod
     def produce(self, messages):
+        """
+        Produce messages to topic
+
+        :type messages: Iterable of strings, or iterable of (key, value) tuples
+                        to produce keyed messages
+        """
         pass
--- a/kafka/rdkafka/producer.py
+++ b/kafka/rdkafka/producer.py
@@ -54,17 +54,18 @@ class Producer(base.BaseProducer):
 
     def produce(self, messages):
         """ Sync-producer: raises exceptions on delivery failures """
-        delivery_reports = [list() for _ in range(len(messages))]
+        delivery_reports = []
         # This ^^ list of lists is perhaps overly cautious and bulky.  It's
         # meant to provide thread-safety without locking, but maybe all
         # delivery callbacks will actually run on one thread.
 
-        for msg, dr in zip(messages, delivery_reports):
+        for msg in messages:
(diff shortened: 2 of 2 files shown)


yungchin:
A few things maybe still worth discussing:

* would we want to enable passing partition_ids (which would override the partitioner)? That might get slightly unwieldly if we'd stick to unnamed tuples
* when passing in a single message, should it be possible to just pass a single bytestring?
* is it a problem that passing two unkeyed messages or one keyed message looks very similar, syntactically?

kbourgoin:
:+1: Looks good. Don't worry about Travis, I've fixed this on my branch and will cherry-pick the change to master shortly.

kbourgoin:
Ah, forgot to answer the questions:

1. I've seen the partition ids in the Scala code too, where it isn't sent to Kafka. I think that's excessive, personally, and would just be confusing to use.  We can add it later to the partitoner code if someone needs it.
2. I think adding that would just complicate it. It's not unreasonable to ask someone to send an iterable instead of a message. Again, we can add it later if we're wrong. It won't be hard to change.
3. Ah, nuts. That is a problem, isn't it? Those two are exactly the same.  I'll make a ticket so we can address it.

kbourgoin:
@yungchin I don't the the last question is an issue, actually.  It only becomes an issue if the method accepts individual strings.  If we define it as always accepting an iterable of messages, it's fine.  It's the difference between this:

```python
producer.produce(['test', 'test'])
```
and
```python
producer.produce([('test', 'test')])
```

Mandating that it always be an iterable of messages removes the ambiguity.  This is how I implemented it on the python side: https://github.com/Parsely/pykafka/blob/feature/py_producer/kafka/pykafka/producer.py#L121-L126

yungchin:
Thanks!

Re 1: yeah, I wouldn't have any use for it either - librdkafka, like the scala driver, does have it, which is why I started wondering if we should. 

(In fact, we make use of librdkafka's facility for it here https://github.com/Parsely/pykafka/blob/30ce2a161ec5407e83b0a2abe4f0389ab08e6e16/kafka/rdkafka/producer.py#L67 - because the pykafka partitioner has already decided a partition at that point)

Re 3: sorry, that was confusing wording on my part! Indeed, I just meant that the difference is very small - a single pair of brackets - so I was thinking people with bracket-blindness might hate it. On second thought it's fine though.

PR #127, opened 2015-02-05
Feature/py producer

### LINKED ISSUE #122: Finish Producer Implementation
Finish these implementations.

kbourgoin:
Opening this up as a PR. There's no tests yet, so we'll have to figure those out later.

@yungchin can you take a look at this? In particular, the changes in `base.py`, `partitioners.py` and `common.py` since those might affect you.  There aren't a lot of changes in there, but some worth noting.

Also, the implementation will give you a good idea of how I was thinking to do partitioners in Python.

Fixes #122.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,9 @@ notifications:
 
 before_install:
     - sudo apt-get install -qq libev-dev libsnappy-dev
-    #- sudo apt-get install -qq zookeeper libev-dev libsnappy-dev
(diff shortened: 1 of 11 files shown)


yungchin:
Thanks, this should all work for me. In the librdkafka config, compression-types are set as strings, so I'll actually end up converting the enums back :)

kbourgoin:
Merging this in because it's 90% of the way there. I added a couple FIXME notes for error handling cases that won't work right now.

PR #133, opened 2015-02-20
[RFC] move pykafka.producer __init__ argspec to base producer

yungchin:
These are the changes needed to the base producer, in order to get the rdkafka.producer changes in the convert_config branch to work. They were a straight copy out of pykafka.producer, so I've taken the liberty to also deduplicate the ```__init__``` argspec there.

PR #134, opened 2015-02-24
[RDY] convert pykafka to rdkafka config

### LINKED ISSUE #129: Write adapter for rdkafka config options
We'll need to lay this out in such a way that it's easy to trace which rdkafka options are either a) configurable through some pykafka setting (and under what name) or b) fixed to a default/non-default choice that makes the most sense for pykafka.

(Looking at @kbourgoin's work, most of these will be translated from the Producer/Consumer __init__ options.)

yungchin:
This stuff is all ready for merging, but would depend on us being happy with pullreq #133.
Would close #129.


kafka/rdkafka/config.py
```diff
@@ -0,0 +1,85 @@
...
+    "refresh_leader_backoff_ms": "topic.metadata.refresh.fast.interval.ms",
+    "required_acks": "request.required.acks",
+    "topic_refresh_interval_ms": "topic.metadata.refresh.interval.ms",
+    }
+
+
+def convert_config(config_callargs, base_config={}):
+    """
+    Convert config_callargs to rd_kafka (config, topic_config) tuple
+    """
```
kbourgoin: Better formatting here is
```
"""Convert config_callargs to rd_kafka (config, topic_config) tuple"""
```
The extra newlines will mess up some code folding programs and don't follow PEP8.

kafka/rdkafka/config.py
```diff
@@ -0,0 +1,85 @@
...
+    "consumer_group": "group.id",
+    "max_retries": "message.send.max.retries",
+    "queued_max_message_chunks": "queued.max.messages.kbytes",
+    "refresh_leader_backoff_ms": "topic.metadata.refresh.fast.interval.ms",
+    "required_acks": "request.required.acks",
+    "topic_refresh_interval_ms": "topic.metadata.refresh.interval.ms",
+    }
+
+
+def convert_config(config_callargs, base_config={}):
```
kbourgoin: Having a mutable default argument is almost always a bad idea.  If that dict gets changed in the course of this function, then those changes will persist.  It'd be better to have `base_config=None` and then a line in the function like `base_config = base_config if base_config is not None else {}`

kafka/rdkafka/config.py
```diff
@@ -0,0 +1,85 @@
...
+    config = copy(base_config)
+    topic_config = {}
+    for key, val in config_callargs.iteritems():
+        if key in NOT_AVAILABLE or val is None:
+            logger.debug("Skipping config item: {}={}".format(key, val))
+            continue
+
+        # we'll do keys first:
+        try: renamed_key = TRANSLATE_NAMES[key]
+        except KeyError: renamed_key = key.replace("_", ".")
```
kbourgoin: It's unnecessarily terse to compress these lines. It makes it too easy to not realize what they are when you're just skimming the code.

kafka/rdkafka/config.py
```diff
@@ -0,0 +1,85 @@
...
+        # now vals:
+        try: converted_val = TRANSLATE_VALUES[key](config_callargs, val)
+        except TypeError: converted_val = TRANSLATE_VALUES[key][val]
+        except KeyError: converted_val = str(val)
+
+        # split them into two dicts:
+        if renamed_key in RD_TOPIC_CONF_NAMES: destination = topic_config
+        elif renamed_key in RD_CONF_NAMES: destination = config
+        else: raise ImproperlyConfiguredError("{}={} ?".format(key, val))
+        destination[renamed_key] = converted_val
```
kbourgoin: Same with the if/elif/else.  It'd be better to have these on separate lines. Otherwise, it's a big block of text and bad for reading.

kafka/rdkafka/consumer.py
```diff
@@ -20,19 +21,27 @@
 
 class SimpleConsumer(base.BaseSimpleConsumer):
 
-    def __init__(self, client, topic, partitions=None):
-        if isinstance(topic, basestring):
-            topic = client[topic]
-        self._topic = topic
-        self._partitions = partitions or self.topic.partitions # ie all
+    def __init__(self, *args, **kwargs):
```
kbourgoin: As annoying as it is, duplicating the argspec is a good idea here.  That way code-completion tools will know what's expected. You don't have to reproduce the docstring, but the copypasta for the args will make this much more usable for people.

kafka/rdkafka/consumer.py
```diff
@@ -20,19 +21,27 @@
...
+        """ For argspec see base.BaseSimpleConsumer.__init__ """
```
kbourgoin: IIRC, if you don't have a docstring at all, Sphinx will just use the one from the base class.

kafka/rdkafka/producer.py
```diff
@@ -16,18 +17,21 @@
 
 class Producer(base.BaseProducer):
 
-    def __init__(self, topic, partitioner=partitioners.random_partitioner):
-        self._topic = topic
-        self._partitioner = partitioner
+    def __init__(self, *args, **kwargs):
+        """ For argspec see base.BaseProducer.__init__ """
```
kbourgoin: Same here about argspec and docstring.

kbourgoin:
Most of this looks good. The feedback is mainly about making the code more readable or maintainable. I'm still not super familiar with python-librdkafka so reviewing that part is hard. I figure we'll catch anything missing when we go to write tests.

kafka/rdkafka/consumer.py
yungchin: Thanks, will change it. I was mostly worried about duplicating the default arguments (I think we'd notice if the argspec got out of sync between the py and rd_kafka implementations, through the test suite, but we might miss it when default arguments would get out of sync?).

This inspect.getcallargs pattern was also part of pullreq #133 that you merged - I'll create a patch reverting that part.

pykafka/rdkafka/consumer.py
```diff
@@ -20,18 +21,47 @@
...
+            auto_commit_enable=DEFAULTS["auto_commit_enable"],
+            auto_commit_interval_ms=DEFAULTS["auto_commit_interval_ms"],
+            queued_max_messages=DEFAULTS["queued_max_messages"],
+            fetch_min_bytes=DEFAULTS["fetch_min_bytes"],
+            fetch_wait_max_ms=DEFAULTS["fetch_wait_max_ms"],
+            refresh_leader_backoff_ms=DEFAULTS["refresh_leader_backoff_ms"],
+            offsets_channel_backoff_ms=DEFAULTS["offsets_channel_backoff_ms"],
+            offsets_commit_max_retries=DEFAULTS["offsets_commit_max_retries"],
+            auto_offset_reset=DEFAULTS["auto_offset_reset"],
+            consumer_timeout_ms=DEFAULTS["consumer_timeout_ms"]):
```
kbourgoin: Is there a real utility to having `DEFAULTS`?  It seems to only obscure what the actual value is while not really adding much.  If these defaults are used in more than one place it could be useful, but otherwise it just seems like an unnecessary complication.  Explicit is better than implicit here, for sure.

kbourgoin:
I took a quick look and this seems to be okay.  There's a lot I don't understand, though, so I'm going to make a more thorough pass soon.

pykafka/rdkafka/consumer.py
yungchin: So `DEFAULTS`, through `get_defaults_dict()` pulls the values from `base.py` - I was thinking I wouldn't have to remember updating this if we update `base.py` then. But happy to drop that if you prefer!

(Btw, given that `base.py` doesn't enforce anything now that it's not an abc, perhaps this might as well pull defaults straight from `pykafka.simpleconsumer` instead...)

yungchin:
We're most likely abandoning the cffi-based bindings, because of performance issues. Now trying a C extension instead. Some of this config-conversion work will re-appear though.

PR #135, opened 2015-02-24
[RDY] rdkafka: topic auto-create

yungchin:
Sorry if this is a bit dodgy - tests would pass after merging from #134, but then you'd get a massive diff of course.
That is, this is all ready once #134 has been merged.


kafka/rdkafka/__init__.py
```diff
@@ -1,6 +1,8 @@
 import logging
+from time import sleep
 
 from kafka import base
+from .config import default_config
```
kbourgoin: Looks like this line made Travis-CI fail. It was unable to import it for some reason.

kbourgoin:
Other than the tests breaking, it looks pretty okay to me.

yungchin:
We're most likely abandoning the cffi-based bindings, because of performance issues. Now trying a C extension instead.

Issue #115, opened 2015-01-26
SimpleConsumer Implementation


kbourgoin:
Implement a python version of the SimpleConsumer. Work in progress in [this pull request](https://github.com/Parsely/pykafka/pull/137).

  - [x] Rename `Consumer` to `BaseSimpleConsumer` in `base.py`
  - [x] Finish implementation that's been started in `pykafka/consumer.py`
  - [x] Unit test `SimpleConsumer`

yungchin:
- [ ] update rdkafka.SimpleConsumer to match interface

emmett9001:
Not sure what `abstrct.py` refers to from the first checklist item. I see there's already a `SimpleConsumer` in `pykafpa/consumer.py`. Is that the one this ticket is about?

yungchin:
Yes, that should be the one this is about - I believe it's still WIP. And abstract.py has been renamed base.py - it's meant to serve as the "interface definition" shared between kafka.pykafka and kafka.rdkafka.

emmett9001:
Some things I notice reading the [kafka-python implementation](https://github.com/mumrah/kafka-python/blob/master/kafka/consumer/simple.py):

* The kafka-python `SimpleConsumer` actually subclasses its `Consumer` class, allowing it to use `commit` and other functions from that class. The stub in our `SimpleConsumer` does not currently subclass the existing `Consumer` (in `kafka/pykafka/consumer.py`) - should it?
* More generally, what is the status of `kafka/pykafka/cosumer.py`? It appears to be defunct, or at least out of date - should I be adding new functionality to it, porting what's there to some other location, or ignoring it completely? The `SimpleConsumer` implementation in kafka-python uses the Offset Commit/Fetch API, but only by way of its extension of the base `Consumer` class and its `commit` function.

emmett9001:
Answers to the above from conversations with @kbourgoin 

The kafka-python `SimpleConsumer` does do roughly the same thing that we need ours to, but we don't need to stick to any design decisions in that library such as subclass/superclass relationships. We can think of our `SimpleConsumer` as roughly the combination of kafka-python [Consumer](https://github.com/mumrah/kafka-python/blob/master/kafka/consumer/base.py#L30) and [SimpleConsumer](https://github.com/mumrah/kafka-python/blob/master/kafka/consumer/simple.py#L68).

[...]

PR #136, opened 2015-03-12
Commit/Fetch API

emmett9001:
This pull request adds support for the [Offset Commit/Fetch API](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-OffsetCommit/FetchAPI) to pykafka. It implements the wire protocol in `pykafka/protocol.py` and uses the new request and response types in `pykafka/broker.py`. This pull request also adds a function `Cluster.get_offset_manager` in `pykafka/cluster.py` that implements the [Consumer Metadata Request](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ConsumerMetadataRequest).

Major changes:
* Adds binary encoders and decoders for all calls in the CF API
* Adds functions to commit and fetch offsets by consumer group to `broker.py`
* Adds `Cluster.get_offset_manager` for discovering the offset manager broker
* Adds unit tests for the binary protocol

Outstanding:
- [x] Merge the [SimpleConsumer](https://github.com/Parsely/pykafka/issues/115) PR.
- [x] Have the `SimpleConsumer` commit and fetch its offsets using the new functions in this PR.
- [x] Update commit/fetch protocol unit tests


kafka/pykafka/cluster.py
```diff
@@ -95,6 +97,27 @@ def _update_topics(self, metadata):
             else:
                 self._topics[name].update(meta)
 
+    def discover_offset_manager(self, consumer_group_name):
```
emmett9001: This function is in `Cluster` since it needs access to all brokers so it can figure out which to return. It's possible that it would be better for it to live elsewhere, but this is my best guess at the moment.

kafka/pykafka/broker.py
```diff
@@ -112,3 +116,53 @@ def request_offsets(self, partition_requests):
...
+                logger.debug("Offset commit failed.")
+            else:
+                break
+
+    def fetch_consumer_group_offsets(self, group, topic, partitions, retries):
+        """Fetch the offsets stored in Kafka with the Offset Commit/Fetch API
+
+        Based on Step 2 here https://cwiki.apache.org/confluence/display/KAFKA/Committing+and+fetching+consumer+offsets+in+Kafka
+        """
+        preqs = [PartitionOffsetFetchRequest(topic, partition.id)
```
emmett9001: The structure of this function will probably change a bit. I think it will make more sense to have the consumer assemble a bunch of `PartitionOffsetFetchRequest`s and then pass the list into this function.

kafka/pykafka/broker.py
```diff
@@ -112,3 +114,24 @@ def request_offsets(self, partition_requests):
...
+    ######################
+    #  Commit/Fetch API  #
+    ######################
+
+    def commit_consumer_group_offsets(self, group, preqs, retries):
+        """Commit the offsets of all messages consumed so far by this consumer
+            group with the Offset Commit/Fetch API
+
+        Based on Step 2 here https://cwiki.apache.org/confluence/display/KAFKA/Committing+and+fetching+consumer+offsets+in+Kafka
+        """
```
kbourgoin: :param: and :type: comments would be really useful here.

kafka/pykafka/broker.py
```diff
@@ -112,3 +114,24 @@ def request_offsets(self, partition_requests):
...
+        req = OffsetCommitRequest(group, partition_requests=preqs)
+        self.handler.request(req).get(OffsetCommitResponse)
```
kbourgoin: Why not return the result?

kafka/pykafka/cluster.py
```diff
@@ -95,6 +98,38 @@ def _update_topics(self, metadata):
             else:
                 self._topics[name].update(meta)
 
+    def get_offset_manager(self, consumer_group_name):
```
kbourgoin: In broker.py this is `group` and here it's `consumer_group_name`.  There's no ConsumerGroup type, so `_name` is needless specification.  It'd be best to standardize on simply `consumer_group`.

kafka/pykafka/cluster.py
```diff
@@ -95,6 +98,38 @@ def _update_topics(self, metadata):
...
+        while True:
+            try:
+                retries += 1
+                req = ConsumerMetadataRequest(consumer_group_name)
+                future = broker.handler.request(req)
+                res = future.get(ConsumerMetadataResponse)
+            except Exception:
+                logger.debug('Error discovering offset manager. Sleeping for {}s'.format(backoff))
+                if retries < MAX_RETRIES:
+                    time.sleep(backoff)  # XXX - not sure if this works here
```
kbourgoin: Don't see why this wouldn't work.

kbourgoin:
This all looks fine to me.  Not sure why Travis is failing, but once that's set I see no reason this can't be merged.  The protocol.py work is a bit much to sort through, but the tests look good.  As long as those are passing, I'm sure this is all good.

kafka/pykafka/broker.py
emmett9001: The result has nothing in it except error codes, which trigger exceptions in the protocol if present. There's no information to return.

kafka/pykafka/cluster.py
emmett9001: When I wrote the comment, I was concerned about how this function would be used with respect to threading and if there was some existing backoff functionality I should be reusing.

kafka/pykafka/broker.py
kbourgoin: :+1: should've checked that first.

tests/kafka/pykafka/test_protocol.py
```diff
@@ -177,5 +177,56 @@ def test_response(self):
...
+        preq = protocol.PartitionOffsetFetchRequest('testtopic', 0)
+        req = protocol.OffsetFetchRequest('test', partition_requests=[preq, ])
+        msg = req.get_bytes()
+        self.assertEqual(
+            msg,
+            bytearray(b'\x00\x00\x00.\x00\t\x00\x00\x00\x00\x00\x00\x00\x07pykafka\x00\x04test\x00\x00\x00\x01\x00\ttesttopic\x00\x00\x00\x01\x00\x00\x00\x00')
+        )
+
+    def test_offset_fetch_response(self):
+        # TODO - this is what I get back from kafka at the moment, clearly it's wrong
```
emmett9001: This is old, should be removed

kafka/pykafka/cluster.py
kbourgoin: Hm. We haven't added the threading yet. Could be a mistake to wait on it, but hopefully that sleep won't be much of an issue.

tests/kafka/pykafka/test_protocol.py
```diff
@@ -177,5 +177,56 @@ def test_response(self):
...
+        response = protocol.OffsetFetchResponse(
+            buffer('\x00\x00\x00\x01\x00\x0cemmett.dummy\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00')
+        )
+        self.assertEqual(response.topics['emmett.dummy'][0].metadata, '')
```
emmett9001: @kbourgoin One thing I noticed when testing is that I can never get an OffsetFetchResponse to return anything other than `''` in the `metadata` field.

kbourgoin: I'm not sure if that's correct of not. I've never used the API myself.

emmett9001:
I discovered here that I needed to use the new api version indicator (1 vs 0) to tell kafka to use a topic for offset management rather than zookeeper. I also discovered that some documentation I'd been reading was old, and that [this](https://github.com/apache/kafka/blob/0.8.2/core/src/main/scala/kafka/api/OffsetCommitRequest.scala#L61) was the cause of the corrupt offsets I was seeing earlier today. All fixed now, and I should be writing some unit tests for the `SimpleConsumer` on monday. I'll also leave the merge here until then in case @kbourgoin or @yungchin want to have another look.

PR #137, opened 2015-03-20
SimpleConsumer Implementation

emmett9001:
This pull request adds a `SimpleConsumer` to pykafka. It is based loosely on the [Scala implementation](https://github.com/apache/kafka/blob/0.8.2/core/src/main/scala/kafka/consumer/SimpleConsumer.scala) and influenced by the [kafka-python version](https://github.com/mumrah/kafka-python/blob/master/kafka/consumer/simple.py#L68).

Major changes:
* Standardizes naming between uses of the Offset API and the Offset Commit/Fetch API
* Adds a `SimpleConsumer` implementation

This pull request changes base.py, so @yungchin and @kbourgoin should weigh in here on those changes. It is also blocking the merge of the [CF API pull request](https://github.com/Parsely/pykafka/pull/136).


kafka/base.py
```diff
@@ -127,7 +127,7 @@ def latest_offset(self):
         """
         raise NotImplementedError
 
-    def earliest_offset(self):
+    def earliest_available_offset(self):
```
kbourgoin: Big +1 to name changes.  Irrespective of the needs of the Offset CF API, these are much more descriptive names.

kafka/pykafka/broker.py
```diff
@@ -77,7 +77,7 @@ def connect(self):
     def fetch_messages(self,
                        partition_requests,
                        timeout=30000,
-                       min_bytes=1024):
+                       min_bytes=1):
```
kbourgoin: Pretty sure I got 1024 from the Scala driver.  If it only returns 1 byte, that won't even cover the message header, so you couldn't decode anything meaningful anyway.

kafka/pykafka/broker.py
```diff
@@ -86,9 +86,10 @@ def fetch_messages(self,
         """
         future = self.handler.request(FetchRequest(
             partition_requests=partition_requests,
-            timeout=10000,
-            min_bytes=1,
+            timeout=timeout,
+            min_bytes=min_bytes,
         ))
+        # XXX - this call returns even with less than min_bytes of messages?
```
kbourgoin: Not sure about this comment. Are you asking if it's possible or have you seen it happen and we don't handle it correctly?

kafka/pykafka/broker.py
```diff
@@ -104,7 +105,7 @@ def produce_messages(self, produce_request):
             # Any errors will be decoded and raised in the `.get()`
         return None
 
-    def request_offsets(self, partition_requests):
+    def request_offset_limits(self, partition_requests):
```
kbourgoin: Another +1 to naming change. Nice improvement.

kafka/pykafka/broker.py
emmett9001: I'm not sure if this page is outdated, but I found 1 here https://kafka.apache.org/08/configuration.html (search for fetch.min.bytes)

kafka/pykafka/broker.py
emmett9001: The call does in fact return with less than min_bytes of messages. I suppose that points to a bug in the handler code?

kafka/pykafka/consumer.py
```diff
@@ -29,15 +133,63 @@ def topic(self):
...
+    def fetch_message_max_bytes(self):
+        return self._fetch_message_max_bytes
+
     def __iter__(self):
         while True:
-            yield self.consume()
+            message = self.consume(timeout=self._socket_timeout_ms)
+            if not message and self._consumer_timed_out():
+                raise StopIteration
+            yield message
```
kbourgoin: Oh, this isn't what I expected.  We'll need to document that iterating this isn't exactly an infinite iterator, but I think I like this solution better.

kafka/pykafka/consumer.py
```diff
@@ -29,15 +133,63 @@ def topic(self):
...
+
+    def _consumer_timed_out(self):
+        if self._consumer_timeout_ms == -1:
+            return False
+        disp = (time.time() - self._last_message_time) * 1000.0
+        return disp > self._consumer_timeout_ms
```
kbourgoin: Are there cases where `.consume` will return `None` other than timing out?  If not, this function and tracking the last_message_time can be avoided.

kafka/pykafka/consumer.py
```diff
@@ -29,15 +133,63 @@ def topic(self):
...
     def consume(self, timeout=None):
         """Get one message from the consumer.
 
         :param timeout: Seconds to wait before returning None
         """
+        owned_partition = self.partition_cycle.next()
+        message = owned_partition.consume(timeout=timeout)
+
+        if self._auto_commit_enable:
+            self._auto_commit()
```
kbourgoin: Ultimately, this needs to run on a separate thread.  It should be auto-committing regardless of whether or not the consumer is actively consuming.

kafka/pykafka/consumer.py
```diff
@@ -29,15 +133,63 @@ def topic(self):
...
+            self.commit_offsets()
+
+    def commit_offsets(self):
+        """Use the Offset Commit/Fetch API to commit offsets for this
+            consumer's topic
+        """
+        if not self.consumer_group:
+            raise Exception("consumer group must be specified to commit offsets")
+
+        self._last_auto_commit = time.time()
```
kbourgoin: Should be in the `_auto_commit` function so that a user calling this doesn't interfere with that functionality.

kafka/pykafka/consumer.py
emmett9001: In cases where there are no messages waiting in any of the `OwnedPartition` queues, `Consumer.consume` will spin and continue to return `None`. Given this, I don't see a way to avoid tracking this timeout and still do something with the `consumer_timeout_ms` kwarg.

kafka/pykafka/consumer.py
emmett9001: We can keep a reference to the autocommit worker thread (a `handlers.ThreadingHandler` instance?) around, start it in __init__ and just keep it running for the lifetime of the `Consumer`. It can run lines 161 and 162 in their own function. Does that sound reasonable?

kafka/pykafka/consumer.py
emmett9001: Good point

kafka/pykafka/consumer.py
kbourgoin: The consumer is going to need to get a `cluster.Cluster` instance in the constructor and then you can do `self.cluster.handler.spawn`.  The original intent of those handlers, carried forward from samsa, is that it'll let you switch between `threading` and `gevent` seamlessly.

kafka/pykafka/consumer.py
kbourgoin: Fair enough. It's not a big deal and I don't see a way around it either.

kafka/pykafka/consumer.py
```diff
@@ -29,15 +133,63 @@ def topic(self):
...
+
+        for broker, partitions in self._partitions_by_leader:
+            # XXX create a bunch of PartitionOffsetCommitRequests
```
kbourgoin: Is this a TODO?  I find `XXX` confusing. :-\

kafka/pykafka/consumer.py
```diff
@@ -47,9 +199,63 @@ class OwnedPartition(object):
...
+
+        try:
+            message = self._messages.get_nowait()
+            self._queued_chunks -= 1
+            self.last_offset_consumed = message.offset
+            return message
+        except Empty:
+            return None
+
+    def _fetch_last_known_offset(self):
```
kbourgoin: Many offsets are known in life.  `_fetch_committed_offset` would be more accurate, I think.

kbourgoin: PS: `last` and `latest` would be confusing as they're the "most recent" in terms of time, but may in fact not be further in the queue than previously committed offsets.  However, as far as the driver is concerned, there is only one committed offset for this consumer group/partition, so there's only one thing to ever be fetched.

kafka/pykafka/consumer.py
emmett9001: Yeah it's a TODO. I'll stick to that rather than XXX

kafka/pykafka/consumer.py
```diff
@@ -47,9 +199,63 @@ class OwnedPartition(object):
     Used to keep track of offsets and the internal message queue.
     """
 
-    def __init__(self, partition):
+    def __init__(self, partition, consumer):
```
kbourgoin: Adding `consumer` makes this a nice little reference cycle, which Python's GC can handle, but is a bit confusing.  It may be better to have `consumer_group=None` and `fetch_message_max_bytes=1` (or whatever) to avoid the cycle and make it a straighter object graph.

kafka/pykafka/consumer.py
emmett9001: That is a great catch, I didn't realize the naming confusion I was creating here.

kafka/pykafka/consumer.py
```diff
@@ -47,9 +199,63 @@ class OwnedPartition(object):
...
         self.partition = partition
+        self.consumer = consumer
         self._messages = Queue()
+        self.last_offset_consumed = 0
+        self.next_offset = 0
+
+        if self.consumer._auto_commit_enable and self.consumer.consumer_group is not None:
```
kbourgoin: I don't see why `_auto_commite_enable` has any bearing on whether or not you're fetching the committed offset.  Only the presence of `consumer_group` should matter.

kafka/pykafka/consumer.py
```diff
@@ -47,9 +199,63 @@ class OwnedPartition(object):
...
+            self.last_offset_consumed = self._fetch_last_known_offset()
```
kbourgoin: Also need to set `self.next_offset` here too.

kafka/pykafka/consumer.py
emmett9001: The thought was that there would not necessarily have been a non-automatic commit prior to this call if autocommit were not enabled, but that doesn't really make any sense as I look at it now.

kafka/pykafka/consumer.py
```diff
@@ -47,9 +199,63 @@ class OwnedPartition(object):
...
 
     def consume(self, timeout=None):
+        """Get a single message from this partition
+        """
+        if self._messages.empty():
+            self._fetch(timeout=timeout)
+
+        try:
+            message = self._messages.get_nowait()
+            self._queued_chunks -= 1
```
kbourgoin: Queued chunks as weird, and I think we want to avoid this setting.  My (admittedly poor) understanding of the Scala driver is that it has chunks of bytes queues and still serialized.  It then deserializes them on demand into messages and provides them to the user.  This seems odd to me because if you don't know the offset of the final message in a chunk (as it's still serialized), how do you know where to start fetching the next chunk?  It's all very confusing.

I talked with @yungchin about this a while back and IIRC librdkafka uses `max_messages_queued` which is simply the number of messages to keep in this per-partition queue.  It seems similar to how you've implemented `_queued_chunks`, so maybe all we need here is a name change.

@yungchin is my understanding of librdkafka here correct?  I remember us agreeing that the way there is better than the Scala driver

emmett9001: I'd be glad to drop this setting, as I'm also very confused by how it should actually work. `max_messages_queued` seems simpler.

kafka/pykafka/consumer.py
```diff
@@ -47,9 +199,63 @@ class OwnedPartition(object):
...
+        self.next_offset = 0
+
+        if self.consumer._auto_commit_enable and self.consumer.consumer_group is not None:
+            self.last_offset_consumed = self._fetch_last_known_offset()
 
     def consume(self, timeout=None):
+        """Get a single message from this partition
+        """
+        if self._messages.empty():
+            self._fetch(timeout=timeout)
```
kbourgoin: This is another thing which should exist on a separate thread, but I'm going to throw a monkey wrench in the code.  Each partition shouldn't be fetching from brokers for messages.  This would create one thread per partition and a lot of simultaneous requests to the broker.  Kafka 0.8 is designed such that a single requests to the broker can ask for data from multiple topics/partitons.  https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-FetchAPI

Instead of having on thread for each OwnedPartition, the SimpleConsumer should start a fetch thread which regularly polls all owned partitions for any where `len(op.messages.qsize()) < self._min_queued_messages`, organizes them by broker, and fires off the minimum number of requests to the broker.

Note: This may contradict something I told you before.  If so, you have my apologies. I kind of forgot I wanted to do this until right now.  The thread-per-partition thing is part of why samsa is a bit on the slow side when you have a ton of partitions.

kafka/pykafka/consumer.py
kbourgoin: Definitely check with @yungchin because i think it also has min_messages_queued.  That way, there's a window in there where the queue is mostly full, but not entirely full, and you don't ask for more messages.  This avoids the case where every single read of the queue causes a request to the broker.  Pretty sure librdkafka does that. If not, we won't do it either because it'd be difficult to do on the librdkafka side.

kafka/pykafka/consumer.py
emmett9001: Sounds good. Should this fetch thread be a `ThreadingHandler`?

kafka/pykafka/consumer.py
```diff
@@ -47,9 +199,63 @@ class OwnedPartition(object):
...
+                )
+                response = self.partition.leader.fetch_messages(
+                    [request], timeout=timeout,
+                    min_bytes=self._fetch_min_bytes
+                )
+
+                messages = response.topics[topic_name].messages
+
+                if self._queued_chunks >= self._queued_max_message_chunks:
+                    return
```
kbourgoin: When changing this to messages queued: this doesn't have to be precise.  If you're supposed to stop at 100 queued messages, but you've downloaded 10 extra, there's no harm in adding them to the queue. The temporary bulk is preferable to redownloading those messages.

kafka/pykafka/partition.py
```diff
@@ -36,20 +35,48 @@ def isr(self):
     def topic(self):
         return self._topic
 
-    def fetch_offsets(self, offsets_before, max_offsets=1):
+    def fetch_offset(self, consumer_group):
```
kbourgoin: `fetch_consumer_offset` would be preferable here to help differentiate from other offsets to be fetched.

emmett9001: I used `fetch_offset_limit` and `latest/earliest_available_offset` to differentiate the names in `Partition`.

kafka/pykafka/partition.py
```diff
@@ -36,20 +35,48 @@ def isr(self):
...
+        """Use the Offset Commit/Fetch API to set the current offset for this
+            partition for the given consumer group
+
+        :param consumer_group: the name of the consumer group for which to
+            fetch an offset
+        :type consumer_group: str
+        """
+        pass
+
+    def fetch_offset_limit(self, offsets_before, max_offsets=1):
```
kbourgoin: I was going to suggest a better name, but man this API is weird. I have no better suggestions, and this thing is rarely used anyway.

kafka/pykafka/topic.py
```diff
@@ -52,12 +58,13 @@ def fetch_offsets(self, offsets_before, max_offsets=1):
...
-        return Producer(self)
+    def earliest_available_offset(self):
+        """Get the earliest offset for each partition of this topic."""
+        return self.fetch_offset_limits(OffsetType.EARLIEST)
 
-    def latest_offsets(self):
+    def latest_available_offset(self):
         """Get the latest offset for each partition of this topic."""
-        return self.fetch_offsets(OFFSET_LATEST)
+        return self.fetch_offset_limits(OffsetType.LATEST)
```
kbourgoin: Should be `earliest_available_offsets` and `latest_available_offsets` since it returns multiple results.

kafka/pykafka/topic.py
```diff
@@ -36,11 +35,18 @@ def name(self):
...
 
-    def earliest_offsets(self):
-        """Get the earliest offset for each partition of this topic."""
-        return self.fetch_offsets(OFFSET_EARLIEST)
+    def get_producer(self):
+        return Producer(self)
+
+    def fetch_offset_limits(self, offsets_before, max_offsets=1):
+        """Use the Offset API to find a limit of valid offsets
+            for each partition in this topic
```
kbourgoin: Total nitpick, but the standard is to have the first line be a short, single-line description of the function.  The hanging indent is just confusing to me. :)

emmett9001: What to do when the short description takes more than a line to write? :D

kbourgoin:
First pass review is done, and things are looking good.  In addition to the comments I made, I'd really like to see some basic unit tests for the SimpleConsumer.  We don't have a good solution for integration testing yet, so just mocked unit tests ought to be sufficient.

kafka/pykafka/topic.py
kbourgoin: In the words of Mark Twain, eschew surplusage.

kbourgoin: And yes, i've been known to spend a few minutes finding just the right way to phrase a comment to find on a single, heavily-nested line.

emmett9001:
Sounds good. I'll address these comments tomorrow and write some tests, then merge. Hopefully I can do the same with the CF API pull request and use it in the `SimpleConsumer`.

Issue #116, opened 2015-01-26
BalancedConsumer Implementation


kbourgoin:
Implement a `BalancedConsumer` that manages one or more `SimpleConsumer`s and balances their use of partitions. The balancing algorithm has been implemented correctly in the old code, so that can be used again, or as a reference, for the new implementation.

We should only need one implementation of this for both the python and librdkafka sides of the driver.  If we use `SimpleConsumer`s to control which partitions we read from, it should be possible to use the same code for both paths.

kbourgoin:
The code for the previous implementation can be found [here](https://github.com/Parsely/pykafka/blob/b97401ce98cfbe2bccbf1dac95e65fea25f1a6b5/samsa/consumer/consumer.py#L214).

It may also be helpful to look at the description of the algorithm itself, which can be found [here](http://kafka.apache.org/documentation.html).  Search the page for "Consumer rebalancing algorithm".

emmett9001:
@kbourgoin It's not completely clear to me what you mean by "use SimpleConsumers to control which partitions we read from". Currently I'm imagining that I'll implement a function similar to the old `decide_partitions` in the new `BalancedConsumer` and use its output as the `partitions` kwarg of `SimpleConsumer`. Does this generally match what you had in mind?

kbourgoin:
Yeah, precisely that.

emmett9001:
Today I started the new BalancedConsumer implementation, largely based on the old samsa.consumer.consumer.Consumer. I've gotten it to the point where consumption is being correctly balanced between one to three consumers of a five-partition cluster (these are just the tests I've been doing).

emmett9001:
`BalancedConsumer` is working well. I've added support for a bunch of different configuration options from the kafka documentation and tested the consumer with up to 15 partitions shared between up to 8 consumer instances. It all works well.

emmett9001:
Time to start testing the BalancedConsumer. The branch is merged, so we'll soon being testing it in an 0.8.2.1 cluster.

PR #138, opened 2015-04-08
[WIP] Balanced Consumer

emmett9001:
This pull request adds a `BalancedConsumer` to pykafka. Related to [this issue](https://github.com/Parsely/pykafka/issues/116).

Since the minute details of actual message consumption are handled in `SimpleConsumer`, and since the `BalancedConsumer` maintains its own instance of one, the `BalancedConsumer` class essentially amounts to a 250+ line implementation of the consumer rebalancing algorithm detailed [here](http://kafka.apache.org/documentation.html).

Currently, this is the only piece of pykafka that uses a `KazooClient`.

# Major changes
* Implements `BalancedConsumer`
* Adds some configuration parameters to `SimpleConsumer`
* Makes messages fetching thread-safe in the `SimpleConsumer`
* Uses a single buffer allocation for kafka api responses
* Adds convenience methods to `Topic` for creating consumers
* Adds a few configuration parameters to `Cluster`
* Maintains a second socket connection for CF API communication
* Implements automatic offset resetting in the `SimpleConsumer`

# Todo
* [x] To properly support the `offsets.channel` config options, `Broker` should maintain two connections: one for non-offset-related requests and one for offset-related requests. This will allow the socket timeout to be configurable for commit requests independently of the timeout for other requests.
* [x] Support other remaining configuration options

### DIFF
--- a/kafka/pykafka/balancedconsumer.py
+++ b/kafka/pykafka/balancedconsumer.py
@@ -5,6 +5,7 @@ import itertools
 
 from kazoo.exceptions import NoNodeException
 from kazoo.client import KazooClient
+from kazoo.recipe.watchers import ChildrenWatch
 
 from kafka.pykafka.simpleconsumer import SimpleConsumer
 
@@ -34,6 +35,7 @@ class BalancedConsumer():
         self._id = "{}:{}".format(socket.gethostname(), uuid4())
 
         self._zookeeper = self._setup_zookeeper(zk_host)
+        self._add_self()
         self._consumer = self._setup_internal_consumer()
 
     def _setup_zookeeper(self, zk_host):
@@ -100,6 +102,55 @@ class BalancedConsumer():
                 pass  # disappeared between ``get_children`` and ``get``
         participants.sort()
         return participants
+
+    def _add_self(self):
+        """Add this consumer to the zookeeper participants.
+
+        Ensures we don't add more participants than partitions
+        """
+        participants = self._get_participants()
+        if len(self._topic.partitions) <= len(participants):
+            log.debug("More consumers than partitions.")
+
+        path = '{}/{}'.format(self._id_path, self._id)
+        self._zookeeper.create(
+            path, self._topic.name, ephemeral=True, makepath=True)
+
+        # Set all our watches and then rebalance
+        self._rebalancing = False
+        broker_path = '/brokers/ids'
+        try:
+            self._broker_watcher = ChildrenWatch(
+                self._zookeeper, broker_path,
+                self._brokers_changed
+            )
+        except NoNodeException:
+            raise Exception(
+                'The broker_path "%s" does not exist in your '
+                'ZooKeeper cluster -- is your Kafka cluster running?'
+                % broker_path)
+
+        self._topics_watcher = ChildrenWatch(
+            self._zookeeper,
+            '/brokers/topics',
+            self._topics_changed
+        )
+        self._rebalancing = True
+
+        # Final watch will trigger rebalance
+        self._consumer_watcher = ChildrenWatch(
+            self._zookeeper, self._id_path,
+            self._consumers_changed
+        )
+
+    def _brokers_changed(self, brokers):
+        pass
+
+    def _consumers_changed(self, consumers):
+        pass
+
+    def _topics_changed(self, topics):
+        pass
 
     def consume(self):
         """Get one message from the consumer
--- a/kafka/pykafka/balancedconsumer.py
+++ b/kafka/pykafka/balancedconsumer.py
@@ -116,6 +116,8 @@ class BalancedConsumer():
         self._fetch_wait_max_ms = fetch_wait_max_ms
         self._rebalance_backoff_ms = rebalance_backoff_ms
         self._consumer_timeout_ms = consumer_timeout_ms
+        self._offsets_channel_backoff_ms = offsets_channel_backoff_ms
+        self._offsets_commit_max_retries = offsets_commit_max_retries
 
         self._consumer = None
         self._consumer_id = "{}:{}".format(socket.gethostname(), uuid4())
@@ -169,7 +171,9 @@ class BalancedConsumer():
             num_consumer_fetchers=self._num_consumer_fetchers,
             queued_max_messages=self._queued_max_messages,
             fetch_wait_max_ms=self._fetch_wait_max_ms,
-            consumer_timeout_ms=self._consumer_timeout_ms
+            consumer_timeout_ms=self._consumer_timeout_ms,
+            offsets_channel_backoff_ms=self._offsets_channel_backoff_ms,
+            offsets_commit_max_retries=self._offsets_commit_max_retries
         )
 
     def _decide_partitions(self, participants):
--- a/kafka/pykafka/balancedconsumer.py
+++ b/kafka/pykafka/balancedconsumer.py
@@ -23,6 +23,10 @@ class BalancedConsumer():
                  socket_timeout_ms=30000):
         """Create a BalancedConsumer
 
+        Maintains a single instance of SimpleConsumer, periodically using the
+        consumer rebalancing algorithm to reassign partitions to this
+        SimpleConsumer.
+
         :param topic: the topic this consumer should consume
         :type topic: pykafka.topic.Topic
         :param cluster: the cluster this consumer should connect to
@@ -31,6 +35,14 @@ class BalancedConsumer():
         :type consumer_group: str
         :param zk_host: the ip and port of the zookeeper node to connect to
         :type zk_host: str
+        :param auto_commit_enable: if true, periodically commit to kafka the
+            offset of messages already fetched by this consumer
+        :type auto_commit_enable: bool
+        :param auto_commit_interval_ms: the frequency in ms that the consumer
+            offsets are committed to kafka
+        :type auto_commit_interval_ms: int
+        :param socket_timeout_ms: the socket timeout for network requests
+        :type socket_timeout_ms: int
         """
         self._cluster = cluster
         self._consumer_group = consumer_group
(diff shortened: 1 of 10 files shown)


emmett9001:
@kbourgoin this is ready for an initial review if you're interested. Either way, I'll be testing it tomorrow.

emmett9001:
Ready for a review from @kbourgoin and/or @yungchin.

kafka/pykafka/broker.py
```diff
@@ -16,7 +16,8 @@
 
 class Broker(base.BaseBroker):
 
-    def __init__(self, id_, host, port, handler, timeout):
+    def __init__(self, id_, host, port, handler, timeout,
+                 buffer_size=64 * 1024):
```
kbourgoin: This is a total formatting thing, so feel free to ignore, but when a parameter list has to go on two lines, I've found this style to be cleaner:

```python
def __init__(self,
             id_,
             host,
             port,
             handler,
             timeout,
             buffer_size=64 * 1024):
    """docstring blah blah blah"""
````

The logic in my head, which again ignore if you disagree since this is highly subjective, is that if it's one line you can scan it quickly. You can't really scan two lines as easily when reading. By having one on each line, you lose the ability to scan, which was lost already, but gain the ability to read because it's not a huge blob of text.  It also makes adding/removing parameters easier, especially if they're in the middle of the argument list.

I've seen this done both ways, but adding a couple more params to this feels very unwieldy to me:
```
def __init__(self, id_, host, port, handler, timeout,
             buffer_size=64 * 1024, socket_timeout_ms=1000 * 30,
             required_acks=4):

kafka/pykafka/cluster.py
```diff
@@ -96,11 +102,19 @@ def _update_topics(self, metadata):
...
         for name, meta in metadata.iteritems():
-            if name not in self._topics:
-                self._topics[name] = Topic(self._brokers, meta)
-                logger.info('Adding topic %s', self._topics[name])
-            else:
-                self._topics[name].update(meta)
+            if not self._should_exclude_topic(name):
+                if name not in self._topics:
+                    self._topics[name] = Topic(self, meta)
+                    logger.info('Adding topic %s', self._topics[name])
```
kbourgoin: Not your changes, but I just noticed that this really should be "Discovered topic" instead of "Adding".  We're adding it to the internal model, but it was in Kafka before we got here.

kafka/pykafka/connection.py
```diff
@@ -30,7 +30,8 @@
 class BrokerConnection(object):
     """A socket connection to Kafka."""
 
-    def __init__(self, host, port):
+    def __init__(self, host, port, buffer_size=64 * 1024):
+        self._buff = bytearray(buffer_size)
```
kbourgoin: This is a public repo, and this deserves public applause. Again, really good catch that we can eliminate a lot of memory allocations by simply re-using this buffer.  :clap: :clap: :clap:

kafka/pykafka/connection.py
```diff
@@ -47,7 +48,7 @@ def connect(self, timeout):
         """Connect to the broker."""
         self._socket = socket.create_connection(
             (self.host, self.port),
-            timeout=timeout
+            timeout=float(timeout) / 1000.0
```
kbourgoin: ```from future import __division__```?

kafka/pykafka/connection.py
```diff
@@ -74,10 +75,8 @@ def response(self):
         try:
             size = self._socket.recv(4)
             size = struct.unpack('!i', size)[0]
-            output = bytearray(size)
-            recvall_into(self._socket, output)
-            # TODO: Figure out if correlation ids are worth it
```
kbourgoin: This `TODO` is gone, did you ever look into how other drivers handle the coorelation ids?  This implementation doesn't get messages out of order, so matching them to the original request seemed needless.  Just wondering if you're thinking the same thing.

kafka/pykafka/simpleconsumer.py
```diff
@@ -102,21 +85,26 @@ def __init__(self,
...
 
         self._last_message_time = time.time()
-        self._consumer_timeout_ms = consumer_timeout_ms
 
         self._auto_commit_enable = auto_commit_enable
         self._auto_commit_interval_ms = auto_commit_interval_ms
         self._last_auto_commit = time.time()
 
-        self._queued_max_messages = queued_max_messages
+        self._running = True
```
kbourgoin: Small thing, but you might want to set this after the worker thread has started.

kafka/pykafka/simpleconsumer.py
```diff
@@ -240,18 +239,23 @@ def fetch(self):
...
-                    reqs, timeout=self._socket_timeout_ms,
+                    reqs, timeout=self._fetch_wait_max_ms,
                     min_bytes=self._fetch_min_bytes
                 )
                 for partition_id, pres in response.topics[self._topic.name].iteritems():
-                    partition = self._partitions_by_id[partition_id]
-                    partition.enqueue_messages(pres.messages)
+                    owned_partition = self._partitions_by_id[partition_id]
+                    owned_partition.enqueue_messages(pres.messages)
+                    owned_partition.lock.release()
```
kbourgoin: This section looks good, but multi-threaded code is so hard to reason about and debug. Have you tested this bit much?  I wonder if adding more fetch threads really speeds throughput. It looks like it should, I just want to know how it behaves in practice. That may have to wait until we have a full Kafka cluster set up, though.

kafka/pykafka/balancedconsumer.py
```diff
@@ -0,0 +1,371 @@
+import itertools
+import logging as log
+import signal
+import socket
+import sys
+import time
+from uuid import uuid4
+import weakref
```
kbourgoin: [nitpick] swap so `from` imports are after `import`

kafka/pykafka/balancedconsumer.py
```diff
@@ -0,0 +1,371 @@
+import itertools
```
kbourgoin: Noting before reading the file: can this go one level up? It'd be ideal if the same BalancedConsumer could handle the Python and C implementations.

kafka/pykafka/balancedconsumer.py
```diff
@@ -0,0 +1,371 @@
...
+
+        self._topic_path = '/consumers/{}/owners/{}'.format(self._consumer_group,
+                                                            self._topic.name)
+        self._consumer_id_path = '/consumers/{}/ids'.format(self._consumer_group)
+
+        def _close_zk_connection(signum, frame):
+            self._zookeeper.stop()
+            self._consumer.stop()
+            sys.exit()
+        signal.signal(signal.SIGINT, _close_zk_connection)
```
kbourgoin: I don't think we want to catch SIGINT. This is going to throw an exception if someone tries to run a consumer in the background thread.

kafka/pykafka/balancedconsumer.py
```diff
@@ -0,0 +1,371 @@
...
+                 fetch_wait_max_ms=100,
+                 refresh_leader_backoff_ms=200,
+                 offsets_channel_backoff_ms=1000,
+                 offsets_commit_max_retries=5,
+                 auto_offset_reset=OffsetType.LATEST,
+                 consumer_timeout_ms=-1,
+                 rebalance_max_retries=5,
+                 rebalance_backoff_ms=2 * 1000,
+                 zookeeper_connection_timeout_ms=6 * 1000,
+                 zookeeper_connect='127.0.0.1:2181'):
```
kbourgoin: Instead of zk hosts, would it make more sense to take a kazoo instance as an argument?  It'd let the user change any settings in zk that they wanted and make our use easier. It's how samsa currently works, fwiw.

kafka/pykafka/balancedconsumer.py
```diff
@@ -0,0 +1,371 @@
...
+    def _setup_zookeeper(self, zookeeper_connect, timeout):
+        """Open a connection to a ZooKeeper host
+
+        :param zookeeper_connect: the '<ip>:<port>' address of the zookeeper node to
+            which to connect
+        :type zookeeper_connect: str
+        :param timeout: connection timeout in milliseconds
+        :type timeout: int
+        """
+        hosts = zookeeper_connect.split(',')
```
kbourgoin: This isn't right. The kazoo expects a CSV of hosts, and the full string is used for zookeeper chrooting, too, so we definitely don't want to lose that.

kafka/pykafka/balancedconsumer.py
```diff
@@ -0,0 +1,371 @@
...
+        log.debug("Rebalance triggered by consumer change")
+        self._rebalance()
+
+    def _topics_changed(self, topics):
+        if self._setting_watches:
+            return
+        log.debug("Rebalance triggered by topic change")
+        self._rebalance()
+
+    def consume(self):
```
kbourgoin: Do we want a timeout on this?

kbourgoin: nvm, i see it's set on the consumer.

kafka/pykafka/connection.py
emmett9001: I didn't look into it actually, I just removed it because there are better places for that reminder.

kafka/pykafka/simpleconsumer.py
emmett9001: I've only tested it locally with two threads. Today I'll pump up the thread count and see what's affected. The threading logic is actually really simple as threading logic goes, though. It's just locking each partition to make sure it's ignored if another thread is already dealing with it.

I say it's simple and am probably wrong. :)

kafka/pykafka/balancedconsumer.py
emmett9001: I believe it can, but doesn't that make the c implementation more of a hybrid of python and c? The BalancedConsumer is python and the SimpleConsumer is c.

kafka/pykafka/balancedconsumer.py
emmett9001: Yes, I suppose this is more of a command line use case. I'll move the signal-catching to my test script.

kafka/pykafka/balancedconsumer.py
emmett9001: I think it makes sense to do both, actually. If the user doesn't want to manually create a KazooClient, this can still do it for them.

PR #139, opened 2015-04-14
Revert "move pykafka.producer __init__ argspec to base producer"

yungchin:
This reverts commit 0ebe693873815d0a9114673a2013265f7dd48694 partially:
as discussed in #134 explicit argspecs are to be preferred - the
inspect-based hack that I'm reverting here had already slipped into
master with #133.

### DIFF
--- a/kafka/pykafka/producer.py
+++ b/kafka/pykafka/producer.py
@@ -1,4 +1,3 @@
-import inspect
 import logging
 import time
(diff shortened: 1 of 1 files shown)


yungchin:
No longer needed as @kbourgoin did the same in ac4b7001759d5c14490f24c8079f3ecb576a9ba4

Issue #110, opened 2015-01-26
Handle Errors Seamlessly


kbourgoin:
When the cluster reconfigures, like a leader moving to a different broker, the protocol implementation will raise an error. We need to be certain that both the cluster model and the active consumer/producer are updated.

  - [ ] Make sure relevant errors are caught
  - [ ] Update cluster model to reflect changes
  - [ ] Reconfigure active consumer or producer

emmett9001:
What are some cases that we'll need to handle here? I imagine these changes will look a bit like [this](https://github.com/Parsely/pykafka/blob/dd93ef3e5650c7236ae86a46c7708d53defd4a2b/kafka/pykafka/partition.py#L75). If partitions are always updating their own leaders, what else is there to do here? What are specific error cases that need to be caught?

kbourgoin:
My plan was to go through the error list [here](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ErrorCodes) and handle the exception they'd throw in the broker.

The workflow I have in my head is:

  1. Try operation
  2. Catch exception we know how to handle
  3. Call `cluster.update()`
  4. Try again

This works great in a single-threaded environment. What's going to be hard is reasoning through where the locks need to be and what can go wrong in the worst cases.  I'm hoping a single cluster-wide lock to pause the whole thing while updating works, but I don't know for sure.  This hasn't gotten beyond the "vague idea" state for me.

emmett9001:
Related to https://github.com/Parsely/pykafka/issues/126

PR #142, opened 2015-04-17
Handle errors in consumer

emmett9001:
Related to https://github.com/Parsely/pykafka/issues/110

# Error Checklist
Some of these may be unnecessary, remove or comment if that's the case
- [x] Consumer OffsetOutOfRange
- [ ] Consumer InvalidMessage
- [x] Consumer UnknownTopicOrPartition
- [ ] Consumer InvalidMessageSize
- [ ] Consumer LeaderNotAvailable
- [ ] Consumer NotLeaderForPartition
- [ ] Consumer RequestTimedOut
- [ ] Consumer BrokerNotAvailable
- [ ] Consumer ReplicaNotAvailable
- [ ] Consumer MessageSizeTooLarge
- [ ] Consumer StaleControllerEpochCode
- [ ] Consumer OffsetMetadataTooLargeCode
- [ ] Consumer OffsetsLoadInProgressCode
- [ ] Consumer ConsumerCoordinatorNotAvailableCode
- [ ] Consumer NotCoordinatorForConsumerCode

### DIFF
--- a/pykafka/pykafka/protocol.py
+++ b/pykafka/pykafka/protocol.py
@@ -101,6 +101,11 @@ class Response(object):
                 clsname, response))
 
 
+class PartitionResponse(object):
+    def __init__(self, error):
+        self.error = error
+
+
 class Message(common.Message, Serializable):
     """Representation of a Kafka Message
 
@@ -605,7 +610,7 @@ class FetchRequest(Request):
         return output
 
 
-class FetchPartitionResponse(object):
+class FetchPartitionResponse(PartitionResponse):
     """Partition information that's part of a FetchResponse"""
     def __init__(self, max_offset, messages, error):
         """Create a new FetchPartitionResponse
@@ -613,9 +618,9 @@ class FetchPartitionResponse(object):
         :param max_offset: The offset at the end of this partition
         :param messages: Messages in the response
         """
+        super(FetchPartitionResponse, self).__init__(error)
         self.max_offset = max_offset
         self.messages = messages
-        self.error = error
 
 
 class FetchResponse(Response):
@@ -739,6 +744,18 @@ class OffsetRequest(Request):
                                  pnum, offsets_before, max_offsets)
                 offset += 16
         return output
+
+
+class OffsetPartitionResponse(PartitionResponse):
+    """Partition information that's part of an OffsetResponse"""
+    def __init__(self, offset, error):
+        """Create a new OffsetPartitionResponse
+
+        :param offset: The requested offset
+        :param error: The error code
+        """
+        super(OffsetPartitionResponse, self).__init__(error)
+        self.offset = offset
 
 
 class OffsetResponse(Response):
@@ -937,6 +954,16 @@ class OffsetCommitRequest(Request):
         return output
 
 
+class OffsetCommitPartitionResponse(PartitionResponse):
+    """Partition information that's part of an OffsetCommitResponse"""
+    def __init__(self, error):
+        """Create a new OffsetCommitPartitionResponse
+
+        :param error: The error code for this partition
+        """
+        super(OffsetCommitPartitionResponse, self).__init__(error)
+
+
 class OffsetCommitResponse(Response):
     """An offset commit response
 
@@ -1042,7 +1069,7 @@ class OffsetFetchRequest(Request):
         return output
 
 
-class OffsetFetchPartitionResponse(object):
+class OffsetFetchPartitionResponse(PartitionResponse):
     """Partition information that's part of an OffsetFetchResponse"""
     def __init__(self, offset, metadata, error):
         """Create a new OffsetFetchPartitionResponse
@@ -1050,9 +1077,9 @@ class OffsetFetchPartitionResponse(object):
         :param offset:
         :param metadata: arbitrary metadata that should be committed with this offset commit
         """
+        super(OffsetFetchPartitionResponse, self).__init__(error)
         self.offset = offset
         self.metadata = metadata
-        self.error = error
 
 
 class OffsetFetchResponse(Response):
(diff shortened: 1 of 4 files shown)


pykafka/pykafka/simpleconsumer.py
```diff
@@ -238,40 +247,49 @@ def fetch_offsets(self):
...
-        :type parts_by_error: dict {int: iterable}
+        :param res: a Response object containing partition responses
+        :type res: pykafka.protocol.Response
         :param handlers: mapping of error code to handler
         :type handlers: dict {int: callable(owned_partition, partition_response)}
         """
+        if handlers is None:
+            handlers = {}
+        error_handlers = dict(
+            self._default_error_handlers.items() + handlers.items())
```
kbourgoin: What happens if something is redefined?  Probably safer to do:
```python
error_handlers = self._default_error_handlers.clone()  # clone so updates don't suck
error_handlers.update(handlers)  # passed in ones override defaults
```

pykafka/pykafka/simpleconsumer.py
```diff
@@ -238,40 +247,49 @@ def fetch_offsets(self):
...
         for errcode, parts in parts_by_error.iteritems():
             for owned_partition, pres in parts:
-                if errcode in handlers:
-                    handlers[errcode](owned_partition, pres)
+                if errcode in error_handlers:
+                    error_handlers[errcode](owned_partition, pres)
+
+        self._reset_offsets(
+            [a[0] for a in
+             parts_by_error[OffsetOutOfRangeError.ERROR_CODE]])
```
kbourgoin: Why can't this be an error handler too?  Seems odd to have a dict of handlers and then this one as more special.

kbourgoin: If this can be made into a handler, then this whole function can be moved somewhere in pyafka/utils.  I'd happily use this in the producer stuff I've been working on.

pykafka/pykafka/simpleconsumer.py
```diff
@@ -292,9 +310,8 @@ def _reset_offsets(self, errored_partitions):
             reqs = [owned_partition.build_offset_request(self._auto_offset_reset)
```
kbourgoin: [nitpick] from a few lines above `owned_partitions_by_leader` is quite a mouthful.  It's not a long-lived variable so something simple like `by_partition` shortens the lines before and makes them easier to read through.

pykafka/pykafka/simpleconsumer.py
```diff
@@ -292,9 +310,8 @@ def _reset_offsets(self, errored_partitions):
...
                     for owned_partition in owned_partitions]
             response = broker.request_offset_limits(reqs)
-            for partition_id, offsets in response.topics[self._topic.name].iteritems():
-                owned_partition = self._partitions_by_id[partition_id]
-                owned_partition.set_offset(offsets[0])
+            error_handlers = {0: lambda op, pres: op.set_offset(pres.offset[0])}
```
kbourgoin: This isn't an error handler, it's a success handler.  I think that's an important distinction and might help clarify `_handle_partition_errors` if instead it's `_handle_partition_responses`.

pykafka/pykafka/simpleconsumer.py
```diff
@@ -317,14 +334,14 @@ def fetch(self):
...
                 error_handlers = {0: lambda op, pres: op.enqueue_messages(pres.messages)}
-                self._handle_partition_errors(parts_by_error, error_handlers)
-                self._reset_offsets(
-                    [a[0] for a in parts_by_error[OffsetOutOfRangeError.ERROR_CODE]])
+                self._handle_partition_errors(response, handlers=error_handlers)
             for owned_partition, _ in partition_reqs:
                 owned_partition.lock.release()
 
+    def _raise_error(self, error, info=""):
+        raise error(info)
```
kbourgoin: If the response handler is made generic, this can go with it to utils.

kbourgoin:
Overall this approach makes sense.  It still feels like a decent bit of indirection to me, though. I suppose if you're re-using a lot of handlers it ends up making sense.

pykafka/pykafka/simpleconsumer.py
```diff
@@ -238,40 +247,49 @@ def fetch_offsets(self):
...
-            owned_partition = self._partitions_by_id[partition_id]
-            partitions_by_error[pres.error].append((owned_partition, pres))
+        for topic_name in res.topics.keys():
+            for partition_id, pres in res.topics[topic_name].iteritems():
+                owned_partition = self._partitions_by_id[partition_id]
+                partitions_by_error[pres.error].append((owned_partition, pres))
         return partitions_by_error
 
-    def _handle_partition_errors(self, parts_by_error, handlers):
+    def _handle_partition_errors(self, res, handlers=None):
```
kbourgoin: What about: `def _handle_partition_response(self, res, success_callback, error_callbacks=None):`

Not sure I like naming them callback, but the point is to separate the success from error cases. This function can know that `0` is always success, but it's weird for the calling functions to have to know that as well.

pykafka/pykafka/simpleconsumer.py
emmett9001: This can't be  part of that handler dict because the handler function would involve appending these partitions to some shared sequence somewhere before calling `_reset_offsets`. More simply, `_reset_offsets` requires a list, which is at odds with the handler scheme. I suppose this could be moved into another function and the rest of `_handle_partition_errors` could be put in utils.

pykafka/pykafka/simpleconsumer.py
emmett9001: That's a fine point, though knowledge of specific error codes is sprinkled all over the place.

pykafka/pykafka/protocol.py
```diff
@@ -101,6 +101,11 @@ def raise_error(self, err_code, response):
                 clsname, response))
 
 
+class PartitionResponse(object):
```
emmett9001: This should go away. All of the classes that inherit from it should be converted to namedtuples. Functionally, it's the same thing, and it avoids the overhead of instantiating classes.

PR #143, opened 2015-04-17
Handle Cluster Reconfiguration

emmett9001:
Related to https://github.com/Parsely/pykafka/issues/110

# Error Checklist
Since the [official list](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ErrorCodes) doesn't specify which component should expect which error code, I've attempted to figure that out in this list. For what it's worth, [this](https://github.com/apache/kafka/blob/0.8.2/core/src/main/scala/kafka/common/ErrorMapping.scala#L30-L51) is the canonical source of all error codes Kafka is capable of throwing. Without digging deep on each one, though, I'm hesitant to support some of these others until we actually see them happening.

- [x] Consumer OffsetOutOfRange
- [x] Consumer UnknownTopicOrPartition
- [x] Consumer OffsetMetadataTooLargeCode
- [x] Consumer OffsetsLoadInProgressCode
- [x] Consumer ConsumerCoordinatorNotAvailableCode
- [x] Consumer NotCoordinatorForConsumerCode 
- [x] Producer UnknownTopicOrPartition
- [x] Producer InvalidMessage
- [x] Producer InvalidMessageSize
- [x] Producer MessageSizeTooLarge
- [x] Producer LeaderNotAvailable
- [x] Producer NotLeaderForPartition
- [x] Producer RequestTimedOut

Notes:
* Reading the code, it looks like the only place that a `RequestTimedOut` can be returned is [in response to a ProduceRequest](https://github.com/apache/kafka/blob/0.8.2/core/src/main/scala/kafka/server/DelayedProduce.scala#L51). I believe this means it doesn't need to be caught in the consumer.
* `LeaderNotAvailable` only needs to be [handled](https://github.com/apache/kafka/blob/0.8.2/core/src/main/scala/kafka/producer/async/DefaultEventHandler.scala#L180) during [metadata requests](https://github.com/apache/kafka/blob/0.8.2/core/src/main/scala/kafka/server/MetadataCache.scala#L60), which I think for pykafka means in `Cluster.update()`. The consumer does not need to explicitly react to this error code.
* `BrokerNotAvailable` is never surfaced to the client
* `OffsetsLoadInProgress` should be handled during an [offset fetch request](https://github.com/apache/kafka/blob/0.8.2/core/src/main/scala/kafka/consumer/ZookeeperConsumerConnector.scala#L412) by exponentially backing off
* `NotCoordinatorForConsumer` should be handled during an [offset commit request and an offset fetch request](https://github.com/apache/kafka/blob/0.8.2/core/src/main/scala/kafka/consumer/ZookeeperConsumerConnector.scala#L411) by calling `Cluster.get_offset_manager`.
* `ConsumerCoordinatorNotAvailable` should be handled during offset commit requests and offset metadata requests by exponentially backing off.


pykafka/pykafka/producer.py
```diff
@@ -69,43 +71,70 @@ def __init__(self,
...
+        try:
+            response = broker.produce_messages(req)
+
+            # Figure out if we need to retry any messages
+            to_retry = []
+            for topic, partitions in response.topics.iteritems():
+                for partition, (error, offset) in partitions.iteritems():
+                    if error == 0:
+                        continue  # All's well
+                    if error == UnknownTopicOrPartition.ERROR_CODE:
```
kbourgoin: Retrying here may be wrong. We need to figure out topic auto-creation. It was my understanding that retrying would give Kafka time to create the topic, but we should test and verify that this is the correct course of action.

emmett9001: I think it's fine for this pull request - we may end up realizing during testing that this is wrong.

emmett9001:
Depending on the state of the `update()` method's thread-safety, I believe this is ready to review and merge. @kbourgoin and @yungchin take a look at your leisure.

kbourgoin:
I vote merging sooner than later. I'll be on PyKafka this afternoon (hopefully) but the thread-safety stuff shouldn't hold this up.

Issue #128, opened 2015-02-05
Ensure auto-creating topics works


kbourgoin:
Both the producer and consumer should be able to handle topics that don't exist yet.

- [ ] python implementation
- [x] librdkafka implementation

yungchin:
Ack. On the rdkafka side, I'm planning to add some magic to rdkafka.cluster.topics for this.

kbourgoin:
What are you thinking there? I haven't done any work on the python side yet, so I'm not entirely sure where it would live. I was thinking to create essentially a VirtualTopic class which would handle things until it was replaced by a real `Topic` at some point.  Not really sure that works, though.

yungchin:
I'm not entirely sure yet, but I think it would work ok if cluster.topics
gets a __missing__ method that just creates a new Topic (so the first
access would be a little slow perhaps)?

kbourgoin:
Yeah, I think that's how it'll have to be done.  I need to verify that this even works in Kafka since auto-creating topics was broken on 0.8.0.  I'm assuming it's been fixed since then.

yungchin:
Yep, I've seen it work fine both on kafka 0.8.1 and 0.8.2 - one way to trigger creation is to do a metadata request on that specific topic name. That might be a nice call to stick in a __missing__() method.

PR #147, opened 2015-04-23
Add topic auto-creation.

kbourgoin:
Adds topic auto-creation to the Python implementation for #128.  I went with a very simple implementation with hard-coded retry and backoff values.  I'm not sure if we need to have these configurable, but could probably repurpose some of our settings if so.


pykafka/pykafka/cluster.py
```diff
@@ -3,16 +3,38 @@
...
+    """Dictionary which will attempt to auto-create unknown topics."""
+
+    def __init__(self, cluster, *args, **kwargs):
+        super(TopicDict, self).__init__(*args, **kwargs)
+        self._cluster = weakref.proxy(cluster)
+
+    def __missing__(self, key):
+        logger.info('Topic %s not found. Attempting to auto-create.', key)
+        # Auto-creating will take a moment, so we try 3 times.
+        for i in xrange(3):
```
emmett9001: There are about 15 different places where we do backoff logic like this. I'm wondering if there's a way we can standardize.

pykafka/pykafka/cluster.py
```diff
@@ -3,16 +3,38 @@
...
+            self._cluster.brokers[0].request_metadata(topics=[key])
```
emmett9001: I'm not sure what this does in the context of creating a new topic, but it seems important. Maybe it should be rolled up into a convenience method along with the next line.

emmett9001:
Nice and simple. Looks good!

pykafka/pykafka/cluster.py
kbourgoin: I agree, but I'm going to pass for the sake of this PR.  We could definitely use some standardization there.

pykafka/pykafka/cluster.py
kbourgoin: Yeah, that's a good point. I'll add it as a private method and have a docstring explaining why it does what it does.

PR #154, opened 2015-04-25
utils.socket: avoid overflow exceptions (fixes #151)

### LINKED ISSUE #151: "Buffer overflow in broker connection" upon SimpleConsumer init
I'm reading from a slightly synthetic topic here, with thousands of 1kB messages - still, not outlandish in size. Starting a consumer for that topic, I'm seeing
```
Exception in thread Thread-12:
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "pykafka/simpleconsumer.py", line 192, in fetcher
    self.fetch()
  File "pykafka/simpleconsumer.py", line 349, in fetch
  File "pykafka/broker.py", line 209, in fetch_messages
    return future.get(FetchResponse)
  File "pykafka/handlers.py", line 50, in get
    raise self.error
ValueError: Buffer overflow in broker connection (buffer size is 65536)
```
So now I'm trying different buffer sizes like so:
```
client = pykafka.KafkaClient("kafka:9092", ignore_rdkafka=True, socket_receive_buffer_bytes=1024*1024) 
```
... and find that I keep getting these overflow exceptions, even at 1MB, 4MB - it works at 64MB. The thing is, I'm not very clear what to choose here up front. So I'm wondering.... if we remove that check, and allow `recvall_into()` to write whatever (I think the bytearray would auto-resize?), would that be problematic? It would have me guess one less config variable (ie `socket_receive_buffer_bytes`).

(Maybe instead have `socket_receive_limit_bytes` with some big number, above which auto-resizing would not be allowed?)

yungchin:
We ran into issues with `recvall_into` because it limited the size of
messages that could be received to `socket_receive_buffer_bytes`, and
some messages (such as MessageSets) could be considerably bigger.

This changes how `recvall_into` reads from the socket: while still
reading at most `socket_receive_buffer_bytes`-sized chunks at a time,
it should now be able to assemble messages of any size.

As a downside, that reverts `BrokerConnection.response` to allocating a
fresh bytearray on every call, but to make up for that, this commit now
avoids the slicing/copying that happened in its return statement, as
well as the copies of `chunk` that happened in `recvall_into`.

### DIFF
--- a/pykafka/connection.py
+++ b/pykafka/connection.py
@@ -99,4 +99,4 @@ class BrokerConnection(object):
         size = struct.unpack('!i', size)[0]
         output = bytearray(size)
         recvall_into(self._socket, output, self._buffer_size)
-        return memoryview(output)[4:]  # skip past correlation-id
+        return buffer(output)[4:]  # skip past correlation-id
--- a/pykafka/connection.py
+++ b/pykafka/connection.py
@@ -45,10 +45,10 @@ class BrokerConnection(object):
             hold response data.
         :type buffer_size: int
         """
-        self._buff = bytearray(buffer_size)
         self.host = host
         self.port = port
         self._socket = None
+        self._buffer_size = buffer_size
 
     def __del__(self):
         """Close this connection when the object is deleted."""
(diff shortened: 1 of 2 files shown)


emmett9001:
This seems good to me. I think it could use a look from @kbourgoin before merging, though.

emmett9001:
After thinking through this a little further, we realized that it's easy enough to simply [avoid throwing an exception](https://github.com/Parsely/pykafka/commit/747cc18a9bd9360551b6895b41f50f29fb4d4fa3) and [remove the socket_receive_buffer_bytes parameter](https://github.com/Parsely/pykafka/commit/cf3f9776b43d67c2d8358266c2002e927a6a5f93). This nets the smallest amount of memory allocations out of all proposed solutions, while also getting rid of the need to warn about buffer overflows.

PR #159, opened 2015-05-02
move version.py

atarzwell:
fails to import when importing, moving the version.py to pykafka/ seems to resolve this. 

```python
Python 2.7.9 (default, Mar  1 2015, 12:57:24) 
[GCC 4.9.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pykafka
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/pykafka-1.0.0-py2.7.egg/pykafka/__init__.py", line 1, in <module>
    from version import version
ImportError: cannot import name version
>>> 
```

I am new to this project, so I may be doing something wrong. 

```
3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt7-1 (2015-03-01) x86_64 GNU/Linux
```


dan-blanchard:
Using `pykafka.version` will cause the whole package to get imported at setup time, which is a big problem if you don't already have the prerequisites installed.  We should just [do what we do with streamparse](https://github.com/Parsely/streamparse/blob/master/setup.py#L22-L26) and grab it with a regex.

kbourgoin:
I saw that, but I really don't care for using regex for that either.  I've seen it done a couple different ways, so I want to take a quick look around and see what the options are.  Otherwise, yeah, regex is the best option I've seen.

dan-blanchard:
@kbourgoin I've seen a few successful approaches to this in the past:

1.  Regular expressions (like we do with streamparse).
2.  Using `execfile` (we do this [with SKLL](https://github.com/EducationalTestingService/skll/blob/master/setup.py#L7)). 
3.  Storing it in a plaintext file and just reading that (although then you need to add something to `__init__.py` that loads from the file as well).
4.  Injecting a variable into `__builtins__` in setup.py that says "I'm running setup," and have `__init__.py` only import the subpackages when that is false.  This is what [scikit-learn does](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/__init__.py#L43-L49), although more for the reason that their C modules won't be built at setup time.

dan-blanchard:
Oh, and conda-build also switched recently to using [Python Versioneer](https://github.com/warner/python-versioneer), which grabs the version from your git tags.  It might be nicer.

kbourgoin:
I fixed this via the regex-based solution that @dan-blanchard mentioned. The commit is 7ce19f66d909331c4ca5e1a6b9700157022a3faf.

@emmett9001 can I get your opinion on the version number? I decided to go with Kafka's version number followed by our own counter.  Anyone else is welcome to comment as well, of course.  It's just a scheme that made sense to me, no particular attachment there.

emmett9001:
@kbourgoin I don't like tying pykafka's version number to kafka's - it seems like it could get confusing quickly. Personally I'd use a simpler versioning scheme, but I could be convinced either way.

kbourgoin:
I did it mostly because I was worried about protocol versioning, but maybe it's not really worth it.  If the protocol changes so much that we have to break compatibility, we can just do a major version of pykafka.

Then again, there's currently no way to interrogate Kafka to find its version, so it's not like we have a way of managing compatibility.  Let's stick with a 1.0.0 release next.

PR #162, opened 2015-05-11
Make sure retrying the rebalance also re-checks partition allocation.

kbourgoin:
This fixes a case where a balanced consumer is trying to acquire partitions it should have because it's balancing information is out of date.  To fix, we ensure that every time we retry rebalancing that we re-check which partitions we should have.


emmett9001:
Looks great to me, if it works in production.

kbourgoin:
Yeah, it's working.  We'll write real integration tests for it later.

emmett9001:
:wind_chime:

Issue #172, opened 2015-05-20
blocking consume on empty topic => 100% cpu utilization


ltucker:
First, glad to see this coming along. A meaningful python client with a balanced consumer is a very welcome sight!

One issue I've noticed is that the "blocking" consume() is a very busy loop currently (simple or a balanced-with-simple-inside) eg, the following causes 100% cpu utilization on the current trunk for me:

```
import pykafka

if __name__ == '__main__':
    client = pykafka.KafkaClient()
    topic = client.topics['some.empty-topic']
    consumer = topic.get_simple_consumer()
    for message in consumer:
        # never really even gets here...
        if message is not None:
            print message.offset, message.value
```

It would be interesting to see a select-like-behavior across the partitions with a timeout though I'm sure that's not exactly trivial. The block=False also appears has a very high rate of returning None even if one or more partitions have messages waiting so it is not easy to use non-busily either currently...

emmett9001:
Hi @ltucker, thanks for filing this. I'll be taking a look at it today. I'm wondering if you can elaborate on what you mean by "select-like-behavior".

ltucker:
Hey @emmett9001, mainly my expectation would just be a very quiet cpu when blocking and perhaps to be able to have a specifiable timeout for how long to block (even in the balanced case)

In referring to select-like-behavior, I mean the select/poll/epoll/kqueue family of calls that wait (quietly) for read/write readiness from a set of file descriptors and wake when such a condition arises or a timeout has expired.

I know that's not what's directly under the hood here (visiting and testing a lot of Queues and so forth) -- I can imagine perhaps some sleeping/signaling through threading.Event or Condition, but again probably not totally trivial in the way it's designed.

emmett9001:
I see, thanks for the detail. If I understand your comment about a blocking timeout, that actually already exists [here](https://github.com/Parsely/pykafka/blob/master/pykafka/simpleconsumer.py#L293). It may be slightly confusing that the timeout isn't a kwarg on `consume()`, but you can set it when you create the consumer:
```python
consumer = topic.get_simple_consumer(consumer_timeout_ms=1 * 1000)
```
This tells all subsequent calls to `consume()` with `block=True` (the default) to stop blocking after 1000 milliseconds. Maybe you can try this out and see if you can get some more acceptable behavior as a workaround.

ltucker:
[...]

PR #173, opened 2015-05-21
use a backported py3 Semaphore() to allow non-busy blocking

ltucker:
This branch is a possible solution to #172 

This branch uses a Semaphore to avoid busy-waiting in simpleconsumer.  This uses almost no cpu when blocking for messages.   I have attempted to preserve all other semantics.

All OwnedPartitions share a Semaphore with their owning SimpleConsumer.
When a message arrives into the Queue of an OwnedPartition, the semaphore is incremented.

When the SimpleConsumer wishes to consume a message, it decrements the semaphore (or sleeps until it can do so without making the count negative)

A verbatim backport of threading.Semaphore() from python 3 is used to allow for timeouts in Semaphore.acquire(...) replacing the internal tracking of _last_message_time (this could still be tracked if useful though)

### DIFF
--- a/pykafka/compat.py
+++ b/pykafka/compat.py
@@ -1,0 +1,94 @@
+import sys
+
+__all__ = ['PY3', 'Semaphore']
+
+PY3 = sys.version_info[0] >= 3
+
+
+if PY3:
+    from threading import Semaphore
+else:
+    from threading import Condition, Lock
+    # could use monotonic.monotonic() backport as well here...
+    from time import time as _time
+
+    # -- begin unmodified backport of threading.Semaphore from Python 3.4 -- #
+    class Semaphore:
+        """This class implements semaphore objects.
+
+        Semaphores manage a counter representing the number of release() calls minus
+        the number of acquire() calls, plus an initial value. The acquire() method
+        blocks if necessary until it can return without making the counter
+        negative. If not given, value defaults to 1.
+
+        """
+
+        # After Tim Peters' semaphore class, but not quite the same (no maximum)
+
+        def __init__(self, value=1):
+            if value < 0:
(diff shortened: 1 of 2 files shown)


emmett9001:
This looks great. It seems to me that the fundamental insight is that python's `threading.Condition` implements the sleeping and waking of threads in a non-busy way, and that `Semaphore` here is just a convenient wrapper around a counter and some of these `Condition`s. Huge thanks for checking this in.

I've tested it locally with a simple and a balanced consumer, and it works fine so far. I'll run a few more tests, but I don't see any reason why this shouldn't make its way into master.

One thing I'd ask you to change is to move the `compat` file into the `utils` directory. Once that's done, though, I think I'll be ready to merge it in.

emmett9001:
Here are some throughput numbers from a test against Parsely's internal topic:

This branch:
```
9290.16101693 messages per second
16691.8994126 messages per second
11514.1634643 messages per second
17355.3029494 messages per second
9688.64737541 messages per second
16467.9632443 messages per second
12575.0801395 messages per second
15365.8437233 messages per second
13834.7782301 messages per second
13275.5227708 messages per second
```

Current master:
```
2934.39524262 messages per second
8103.0306952 messages per second
8826.44237952 messages per second
7031.17096217 messages per second
14678.3146102 messages per second
9461.40953258 messages per second
9376.80358197 messages per second
12602.9718947 messages per second
15952.4283669 messages per second
8307.68101976 messages per second
```
This branch might actually make the overall throughput slightly higher, which would be awesome.

emmett9001:
This branch is currently running on a high-throughput test system in Parsely. I'll watch it for a bit to make sure things are working fine, but so far so good.

Issue #146, opened 2015-04-23
SimpleConsumer for librdkafka


kbourgoin:
Status: there's a working `RdkSimpleConsumer` which derives from `pykafka.SimpleConsumer` without changing the interface (see [here][0]). My `%timeit` results are somewhat noisy, but I generally see 4-7 microsecs per `consume` call vs 25-50 microsecs for the pure-python consumer (where me == @yungchin, as I've just edited this summary...).

[0]: https://github.com/Parsely/pykafka/blob/e475d9703f7903e1e9272503583f2ec45f8f21ce/pykafka/rdkafka/simple_consumer.py#L5

Still to do:
- [ ] passing `SimpleConsumer.__init__` config settings into the underlying rdkafka consumer
- [ ] wrap error conditions in the C module in meaningful exceptions (I've marked these `TODO` in the code)
- [ ] figure out if the python side needs more exceptions out of the rdkafka consumer: most cluster reconfiguration error responses and the like are absorbed as a matter of course, so I'm wondering if `KafkaClient.cluster` and co will need more explicit updates to stay informed?
- [ ] integration testing (there's a separate issue for this)
- [ ] hook up logging (also has a separate issue)
- [ ] take a shot at redefining `_rd_kafka.Message` to expose the payload through the buffer API rather than doing a string copy

kbourgoin:
This sounds like it may be as simple as making the interfaces line up, btw.

yungchin:
This is now underway in renewed form at https://github.com/Parsely/pykafka/tree/feature/rdkafka_extension - I'll update the description above to track what still needs doing.

yungchin:
Tracking this in #176 from now.

PR #176, opened 2015-05-31
librdkafka-backed C extension

yungchin:
This adds an optional C-extension which wraps [librdkafka](https://github.com/edenhill/librdkafka), and two wrapper classes deriving from `pykafka.Producer` and `pykafka.SimpleConsumer` respectively, but replacing those classes' worker-thread logic with calls into librdkafka.

The consumer still relies on `SimpleConsumer` logic for offset management (a choice made partly because this is hardly a performance-sensitive aspect of the consumer, and so the C work would be wasted on it, and partly because when work on this started librdkafka wasn't completely ready to store offsets on kafka 0.8.2 brokers).

The `TestCases` added here also derive from those which were written for the parent classes, which means we run the same tests against the child, ensuring a sufficient degree of interface compatibility between the rdkafka-backed and pure-python classes (there's currently only one exception to this - see the skipped test in the producer tests). We also repeat the `BalancedConsumer` integration tests with an rdkafka internal consumer.

I've struggled with pypy compatibility, and have given up on it for the time being. The tests are marked with `skipif` decorators to express this.

## Previously - Status as of 23 October 2015

This may need minor tweaks depending on how #269 (we already have producer futures in this pullreq, which should be kept compatible with those in #269) and #273 turn out (the latter needs fixing because it currently causes test failures and crashes on this branch - we're trying to consume while also restarting a consumer). When those two tickets have been sorted out, this should be otherwise ready for review.

## Previously - Status as of 10 October 2015

Work still pending before it's going to get a "ready" sticker:

  - [x] making the extension throw the same errors from `pykafka.exceptions` as the pure-python classes (where possible)
  - [x] moving the polling thread that's currently only in the producer, so that it exists in the consumer too (I realised while reading some librdkafka source that messages at "error" level won't make it into `logging` if we never poll the rdk-handle)
  - [x] decide what the public interfaces for switching between pure-py and rdkafka classes should look like (currently under consideration in #294 - which also adds some documentation)

## Original description

(Now that - as of c34e904 - this has a few tests passing against it, I'll track it in a PR, to gain the benefit of Travis watching it. The following was mostly copied from #146 which I'll close.)

Status: there's a working `RdkSimpleConsumer` which derives from `pykafka.SimpleConsumer` without changing the interface (see [here][0]). My `%timeit` results are somewhat noisy, but I generally see 4-7 microsecs per `consume` call vs 25-50 microsecs for the pure-python consumer.

[0]: https://github.com/Parsely/pykafka/blob/e475d9703f7903e1e9272503583f2ec45f8f21ce/pykafka/rdkafka/simple_consumer.py#L5

Still to do:
- [x] passing `SimpleConsumer.__init__` config settings into the underlying rdkafka consumer
- [x] wrap error conditions in the C module in meaningful exceptions (I've marked these `TODO` in the code)
- [x] figure out if the python side needs more exceptions out of the rdkafka consumer: most cluster reconfiguration error responses and the like are absorbed as a matter of course, so I'm wondering if `KafkaClient.cluster` and co will need more explicit updates to stay informed?
- [x] more comprehensive integration testing
- [x] handle calls to `reset_offsets` (for use as detailed in #194)
- [x] hook up logging
- [x] ~~take a shot at redefining `_rd_kafka.Message` to expose the payload through the buffer API rather than doing a string copy~~
- [x] extend coverage checking to c-extension

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -30,12 +30,11 @@ import weakref
 
 from kazoo.client import KazooClient
 from kazoo.exceptions import NoNodeException, NodeExistsError
-from kazoo.protocol.states import KazooState
 from kazoo.recipe.watchers import ChildrenWatch
 
 from .common import OffsetType
 from .exceptions import (KafkaException, PartitionOwnedError,
-                         ConsumerStoppedException, ZookeeperConnectionLost)
+                         ConsumerStoppedException, NoPartitionsForConsumerException)
 from .simpleconsumer import SimpleConsumer
 from .utils.compat import range, get_bytes, itervalues
 try:
@@ -213,7 +212,6 @@ class BalancedConsumer():
         self._owns_zookeeper = zookeeper is None
         if zookeeper is not None:
             self._zookeeper = zookeeper
-        self._zk_state_listener = None
         if auto_start is True:
             self.start()
 
@@ -262,8 +260,7 @@ class BalancedConsumer():
 
     @property
     def partitions(self):
-        return self._consumer.partitions if (
-            self._consumer and self._consumer._running) else None
+        return self._consumer.partitions if self._consumer else None
 
     @property
     def _partitions(self):
@@ -286,8 +283,6 @@ class BalancedConsumer():
                 self._setup_zookeeper(self._zookeeper_connect,
                                       self._zookeeper_connection_timeout_ms)
             self._zookeeper.ensure_path(self._topic_path)
-            self._zk_state_listener = self._get_zk_state_listener()
-            self._zookeeper.add_listener(self._zk_state_listener)
             self._add_self()
             self._running = True
             self._set_watches()
@@ -309,7 +304,6 @@ class BalancedConsumer():
             self._running = False
         if self._consumer is not None:
             self._consumer.stop()
-        self._zookeeper.remove_listener(self._zk_state_listener)
         if self._owns_zookeeper:
             # NB this should always come last, so we do not hand over control
             # of our partitions until consumption has really been halted
@@ -318,7 +312,7 @@ class BalancedConsumer():
             self._remove_partitions(self._get_held_partitions())
             try:
                 self._zookeeper.delete(self._path_self)
-            except:
+            except NoNodeException:
                 pass
         # additionally we'd want to remove watches here, but there are no
         # facilities for that in ChildrenWatch - as a workaround we check
@@ -373,20 +367,6 @@ class BalancedConsumer():
             reset_offset_on_start=reset_offset_on_start,
             auto_start=start
         )
-
-    def _suspend_internal_consumer(self):
-        """Suspend (ahem, stop) internal SimpleConsumer
-
-        This lets us temporarily suspend the internal consumer in situations
-        where we cannot assert ownership of our topic partitions.  Currently,
-        it actually just crudely stops it, because SimpleConsumer doesn't have
-        a suspend facility.  If this turns out a performance issue we could do
-        something more sophisticated here.
-        """
-        if self._consumer is not None:
-            log.debug(
-                "Suspending internal consumer ({})".format(self._consumer_id))
-            self._consumer.stop()
 
     def _decide_partitions(self, participants):
         """Decide which partitions belong to this consumer.
@@ -525,6 +505,7 @@ class BalancedConsumer():
             log.info('Rebalancing consumer %s for topic %s.' % (
                 self._consumer_id, self._topic.name)
             )
+
             for i in range(self._rebalance_max_retries):
                 try:
                     # If retrying, be sure to make sure the
@@ -536,15 +517,11 @@ class BalancedConsumer():
                         participants.append(self._consumer_id)
 
                     new_partitions = self._decide_partitions(participants)
-
                     if not new_partitions:
                         should_stop = True
                         log.warning("No partitions assigned to consumer %s - stopping",
                                     self._consumer_id)
                         break
-
-                    if new_partitions != self._partitions:
-                        self._suspend_internal_consumer()
 
                     # Update zk with any changes:
                     # Note that we explicitly fetch our set of held partitions
@@ -557,14 +534,13 @@ class BalancedConsumer():
                     self._remove_partitions(current_zk_parts - new_partitions)
                     self._add_partitions(new_partitions - current_zk_parts)
 
-                    # If suspended previously, restart:
-                    if self._consumer is None or not self._consumer._running:
+                    # Only re-create internal consumer if something changed.
+                    if new_partitions != self._partitions:
                         self._setup_internal_consumer(list(new_partitions))
 
                     log.info('Rebalancing Complete.')
                     break
                 except PartitionOwnedError as ex:
-                    self._suspend_internal_consumer()
                     if i == self._rebalance_max_retries - 1:
                         log.warning('Failed to acquire partition %s after %d retries.',
                                     ex.partition, i)
@@ -637,22 +613,6 @@ class BalancedConsumer():
             return False
         return True
 
-    def _get_zk_state_listener(self):
-        """Callback to suspend internal consumer when zk connection drops"""
-        ref = weakref.ref(self)
-
-        def listener(zk_state):
-            log.info("zk_state_listener: {}".format(zk_state))
-            self = ref()
-            if self is None:  # should never happen as we use remove_listener()
-                return
-            if zk_state != KazooState.CONNECTED:
-                # We don't handle the transition where the connection comes
-                # back: that's covered by the ChildrenWatch watches already
-                self._zookeeper.handler.spawn(self._suspend_internal_consumer)
-
-        return listener
-
     @_catch_thread_exception
     def _brokers_changed(self, brokers):
         if not self._running:
@@ -713,8 +673,8 @@ class BalancedConsumer():
                 return False
             disp = (time.time() - self._last_message_time) * 1000.0
             return disp > self._consumer_timeout_ms
-        if self._consumer is None:
-            raise ConsumerStoppedException
+        if not self._partitions:
+            raise NoPartitionsForConsumerException()
         message = None
         self._last_message_time = time.time()
         while message is None and not consumer_timed_out():
@@ -724,8 +684,6 @@ class BalancedConsumer():
             except ConsumerStoppedException:
                 if not self._running:
                     return
-                elif not self._zookeeper.connected:
-                    raise ZookeeperConnectionLost
                 continue
             if message:
                 self._last_message_time = time.time()
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -32,7 +32,7 @@ from .common import OffsetType
 from .exceptions import (KafkaException, PartitionOwnedError,
                          ConsumerStoppedException)
 from .simpleconsumer import SimpleConsumer
-from .utils.compat import range, get_bytes
+from .utils.compat import range, get_bytes, itervalues
 try:
     from . import rdkafka
 except ImportError:
@@ -179,7 +179,6 @@ class BalancedConsumer():
             hostname=socket.gethostname(),
             uuid=uuid4()
         )
-        self._partitions = set()
         self._setting_watches = True
 
         self._topic_path = '/consumers/{group}/owners/{topic}'.format(
@@ -210,7 +209,8 @@ class BalancedConsumer():
                 time.sleep(120)
                 if not self._running:
                     break
-                self._check_held_partitions()
+                if not self._check_held_partitions():
+                    self._rebalance()
             log.debug("Checker thread exiting")
         log.debug("Starting checker thread")
         return self._cluster.handler.spawn(checker)
@@ -218,6 +218,12 @@ class BalancedConsumer():
     @property
     def partitions(self):
         return self._consumer.partitions if self._consumer else None
+
+    @property
+    def _partitions(self):
+        """Convenient shorthand for set of partitions internally held"""
+        return set(
+            [] if self.partitions is None else itervalues(self.partitions))
 
     @property
     def held_offsets(self):
@@ -234,9 +240,9 @@ class BalancedConsumer():
                                   self._zookeeper_connection_timeout_ms)
         self._zookeeper.ensure_path(self._topic_path)
         self._add_self()
+        self._running = True
         self._set_watches()
         self._rebalance()
-        self._running = True
         self._setup_checker_worker()
 
     def stop(self):
@@ -244,10 +250,22 @@ class BalancedConsumer():
 
         This method should be called as part of a graceful shutdown process.
         """
-        if self._owns_zookeeper:
-            self._zookeeper.stop()
+        with self._rebalancing_lock:
+            # We acquire the lock in order to prevent a race condition where a
+            # rebalance that is already underway might re-register the zk
+            # nodes that we remove here
+            self._running = False
         self._consumer.stop()
-        self._running = False
+        if self._owns_zookeeper:
+            # NB this should always come last, so we do not hand over control
+            # of our partitions until consumption has really been halted
+            self._zookeeper.stop()
+        else:
+            self._remove_partitions(self._get_held_partitions())
+            self._zookeeper.delete(self._path_self)
+            # additionally we'd want to remove watches here, but there are no
+            # facilities for that in ChildrenWatch - as a workaround we check
+            # self._running in the watcher callbacks (see further down)
 
     def _setup_zookeeper(self, zookeeper_connect, timeout):
         """Open a connection to a ZooKeeper host.
@@ -261,13 +279,15 @@ class BalancedConsumer():
         self._zookeeper = KazooClient(zookeeper_connect, timeout=timeout / 1000)
         self._zookeeper.start()
 
-    def _setup_internal_consumer(self, start=True):
+    def _setup_internal_consumer(self, partitions=None, start=True):
         """Instantiate an internal SimpleConsumer.
 
         If there is already a SimpleConsumer instance held by this object,
         disable its workers and mark it for garbage collection before
         creating a new one.
         """
+        if partitions is None:
+            partitions = []
         reset_offset_on_start = self._reset_offset_on_start
         if self._consumer is not None:
             self._consumer.stop()
@@ -281,7 +301,7 @@ class BalancedConsumer():
             self._topic,
             self._cluster,
             consumer_group=self._consumer_group,
-            partitions=list(self._partitions),
+            partitions=partitions,
             auto_commit_enable=self._auto_commit_enable,
             auto_commit_interval_ms=self._auto_commit_interval_ms,
             fetch_message_max_bytes=self._fetch_message_max_bytes,
@@ -402,12 +422,16 @@ class BalancedConsumer():
         if len(self._topic.partitions) <= len(participants):
             raise KafkaException("Cannot add consumer: more consumers than partitions")
 
-        path = '{path}/{id_}'.format(
+        self._zookeeper.create(
+            self._path_self, self._topic.name, ephemeral=True, makepath=True)
+
+    @property
+    def _path_self(self):
+        """Path where this consumer should be registered in zookeeper"""
+        return '{path}/{id_}'.format(
             path=self._consumer_id_path,
             id_=self._consumer_id
         )
-        self._zookeeper.create(
-            path, self._topic.name, ephemeral=True, makepath=True)
 
     def _rebalance(self):
         """Claim partitions for this consumer.
@@ -417,6 +441,8 @@ class BalancedConsumer():
         if self._consumer is not None:
             self.commit_offsets()
         with self._rebalancing_lock:
+            if not self._running:
+                raise ConsumerStoppedException
             log.info('Rebalancing consumer %s for topic %s.' % (
                 self._consumer_id, self._topic.name)
             )
@@ -426,17 +452,27 @@ class BalancedConsumer():
                     # If retrying, be sure to make sure the
                     # partition allocation is correct.
                     participants = self._get_participants()
-                    partitions = self._decide_partitions(participants)
+                    if self._consumer_id not in participants:
+                        # situation that only occurs if our zk session expired
+                        self._add_self()
+                        participants.append(self._consumer_id)
 
-                    old_partitions = self._partitions - partitions
-                    self._remove_partitions(old_partitions)
+                    new_partitions = self._decide_partitions(participants)
 
-                    new_partitions = partitions - self._partitions
-                    self._add_partitions(new_partitions)
+                    # Update zk with any changes:
+                    # Note that we explicitly fetch our set of held partitions
+                    # from zk, rather than assuming it will be identical to
+                    # `self.partitions`.  This covers the (rare) situation
+                    # where due to an interrupted connection our zk session
+                    # has expired, in which case we'd hold zero partitions on
+                    # zk, but `self._partitions` may be outdated and non-empty
+                    current_zk_parts = self._get_held_partitions()
+                    self._remove_partitions(current_zk_parts - new_partitions)
+                    self._add_partitions(new_partitions - current_zk_parts)
 
                     # Only re-create internal consumer if something changed.
-                    if old_partitions or new_partitions:
-                        self._setup_internal_consumer()
+                    if new_partitions != self._partitions:
+                        self._setup_internal_consumer(list(new_partitions))
 
                     log.info('Rebalancing Complete.')
                     break
@@ -458,21 +494,15 @@ class BalancedConsumer():
     def _remove_partitions(self, partitions):
         """Remove partitions from the zookeeper registry for this consumer.
 
-        Also remove these partitions from the consumer's internal
-        partition registry.
-
         :param partitions: The partitions to remove.
         :type partitions: Iterable of :class:`pykafka.partition.Partition`
         """
         for p in partitions:
-            assert p in self._partitions
+            # TODO pass zk node version to make sure we still own this node
             self._zookeeper.delete(self._path_from_partition(p))
-        self._partitions -= partitions
 
     def _add_partitions(self, partitions):
         """Add partitions to the zookeeper registry for this consumer.
-
-        Also add these partitions to the consumer's internal partition registry.
 
         :param partitions: The partitions to add.
         :type partitions: Iterable of :class:`pykafka.partition.Partition`
@@ -484,51 +514,64 @@ class BalancedConsumer():
                     value=get_bytes(self._consumer_id),
                     ephemeral=True
                 )
-                self._partitions.add(p)
             except NodeExistsError:
                 raise PartitionOwnedError(p)
+
+    def _get_held_partitions(self):
+        """Build a set of partitions zookeeper says we own"""
+        zk_partition_ids = set()
+        all_partitions = self._zookeeper.get_children(self._topic_path)
+        for partition_slug in all_partitions:
+            try:
+                owner_id, stat = self._zookeeper.get(
+                    '{path}/{slug}'.format(
+                        path=self._topic_path, slug=partition_slug))
+                if owner_id == get_bytes(self._consumer_id):
+                    zk_partition_ids.add(int(partition_slug.split('-')[1]))
+            except NoNodeException:
+                pass  # disappeared between ``get_children`` and ``get``
+        return set(self._topic.partitions[_id] for _id in zk_partition_ids)
 
     def _check_held_partitions(self):
         """Double-check held partitions against zookeeper
 
-        Ensure that the partitions held by this consumer are the ones that
-        zookeeper thinks it's holding. If not, rebalance.
+        True if the partitions held by this consumer are the ones that
+        zookeeper thinks it's holding, else False.
         """
         log.info("Checking held partitions against ZooKeeper")
-        # build a set of partition ids zookeeper says we own
-        zk_partition_ids = set()
-        all_partitions = self._zookeeper.get_children(self._topic_path)
-        for partition_slug in all_partitions:
-            owner_id, stat = self._zookeeper.get(
-                '{path}/{slug}'.format(
-                    path=self._topic_path, slug=partition_slug))
-            if owner_id == self._consumer_id:
-                zk_partition_ids.add(int(partition_slug.split('-')[1]))
-        # build a set of partition ids we think we own
-        internal_partition_ids = set([p.id for p in self._partitions])
-        # compare the two sets, rebalance if necessary
-        if internal_partition_ids != zk_partition_ids:
+        zk_partitions = self._get_held_partitions()
+        if zk_partitions != self._partitions:
             log.warning("Internal partition registry doesn't match ZooKeeper!")
             log.debug("Internal partition ids: %s\nZooKeeper partition ids: %s",
-                      internal_partition_ids, zk_partition_ids)
-            self._rebalance()
+                      self._partitions, zk_partitions)
+            return False
+        return True
 
     def _brokers_changed(self, brokers):
+        if not self._running:
+            return False  # `False` tells ChildrenWatch to disable this watch
         if self._setting_watches:
             return
-        log.debug("Rebalance triggered by broker change")
+        log.debug("Rebalance triggered by broker change ({})".format(
+            self._consumer_id))
         self._rebalance()
 
     def _consumers_changed(self, consumers):
+        if not self._running:
+            return False  # `False` tells ChildrenWatch to disable this watch
         if self._setting_watches:
             return
-        log.debug("Rebalance triggered by consumer change")
+        log.debug("Rebalance triggered by consumer change ({})".format(
+            self._consumer_id))
         self._rebalance()
 
     def _topics_changed(self, topics):
+        if not self._running:
+            return False  # `False` tells ChildrenWatch to disable this watch
         if self._setting_watches:
             return
-        log.debug("Rebalance triggered by topic change")
+        log.debug("Rebalance triggered by topic change ({})".format(
+            self._consumer_id))
         self._rebalance()
 
     def reset_offsets(self, partition_offsets=None):
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -18,7 +18,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 __all__ = ["BalancedConsumer"]
-import functools
 import itertools
 import logging
 import socket
@@ -36,7 +35,7 @@ from .common import OffsetType
 from .exceptions import (KafkaException, PartitionOwnedError,
                          ConsumerStoppedException, NoPartitionsForConsumerException)
 from .simpleconsumer import SimpleConsumer
-from .utils.compat import range, get_bytes, itervalues
+from .utils.compat import range, get_bytes, itervalues, iteritems
 try:
     from . import rdkafka
 except ImportError:
@@ -89,6 +88,7 @@ class BalancedConsumer(object):
                  zookeeper=None,
                  auto_start=True,
                  reset_offset_on_start=False,
+                 post_rebalance_callback=None,
                  use_rdkafka=False):
         """Create a BalancedConsumer instance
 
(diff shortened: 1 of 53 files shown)


yungchin:
@kbourgoin @emmett9001 I think this is ready for harsh criticism!

I've not created an option in `Topic.get_simple_consumer` for now, so it needs to be created with something like
```
from pykafka.rdkafka import RdKafkaSimpleConsumer
cons = RdKafkaSimpleConsumer(topic, client.cluster, **kwargs)
```
and note that for optimal performance you'll have to pass in `queued_max_messages=10**5` or so, otherwise it uses the SimpleConsumer default.

As for the open tick boxes:

Coverage checking: I can obtain test coverage using gcov, and https://github.com/eddyxu/cpp-coveralls knows how to turn that into coveralls json. Then there's the question how to merge that with python coverage data. It seems that if we switch from python-coveralls to coveralls-python (confused I am) then the latter would know how to merge the cpp-coveralls json: https://github.com/coagulant/coveralls-python#multiple-languages-experimental

Logging out of librdkafka is not hooked up yet (ie ends up on stderr), but I'd like to split that off into a separate job. The plan would be to load the logging module from the C extension.

The remaining thing is whether we want/need a custom Message type instead of what we have now (the current thing is basically equivalent to a `namedtuple`), that avoids copying data out of `rd_kafka_message_t` into python `bytes` (in any case, I think that can be separate from this pullreq, as performance is fair without that already). There would be considerable extra code complexity. My current thinking is that it would have to support the buffer protocol, to maintain a "native python" sort of usability. But to have that, and still be able to write `print msg.value`, would actually require that `value` is the object exposing the buffer protocol, so we'd be creating some nested types, and I'm not sure how nice they would really be to use.

bench/simple_consumer_bench.py
```diff
@@ -0,0 +1,111 @@
...
+    "rdkafka": "cons = RdKafkaSimpleConsumer(topic, client.cluster, **kwargs)",
+    }
+
+
+# We seem to have an issue where consumers don't get garbage-collected, and so
+# their worker threads keep going even after timeit is done with them (and
+# even if we do slip in a stop() call, they don't release the memory for their
+# message queues - I tried that).  As a workaround, we must exit the
+# interpreter after every call to run_bench, and then aggregate the running
+# times in the analysis script instead.
```
yungchin: Just a quick note on this, so I don't forget - I looked at `SimpleConsumer` a bit, and guessed that the gc issue could be due to the worker threads holding a reference to `self` (passed in as a closure). So I had a shot at handing both the fetcher threads and the offset_commit thread a `weakref.proxy` instead, but no luck (even when I removed the `__del__` method (the worker threads, if handed a `weakref.proxy` die anyway, with a `ReferenceError`, so no need for `__del__`).

That is, maybe the above is part of it, but there's also something more that prevents gc.

bench/simple_consumer_bench.py
```diff
@@ -0,0 +1,111 @@
+import json
```
emmett9001: Before merging this file, it would be good for it to gain some docstrings and usage examples. It looks really useful, but I have to pick through the code to understand how it works.

bench/simple_consumer_bench.py
```diff
@@ -0,0 +1,111 @@
...
+
+# We seem to have an issue where consumers don't get garbage-collected, and so
+# their worker threads keep going even after timeit is done with them (and
+# even if we do slip in a stop() call, they don't release the memory for their
+# message queues - I tried that).  As a workaround, we must exit the
+# interpreter after every call to run_bench, and then aggregate the running
+# times in the analysis script instead.
+def run_bench(consumer_type,  # "pure-py" or "rdkafka"
+              num_partitions,  # this and the next must form ...
+              msg_size_bytes,  # ... a combo previously created by mk_topics
```
emmett9001: These comments are docstring material.

pykafka/rdkafka/_rd_kafkamodule.c
```diff
@@ -0,0 +1,448 @@
...
+
+/**
+ * Exception types
+ */
+
+static PyObject *ConsumerStoppedException;
+static PyObject *PyRdKafkaError;
+
+
+static void
```
emmett9001: Is separating the lines in this way the preferred C style? Seems like it would make more sense to insert linebreaks between arguments, but what do I know about C conventions?

pykafka/rdkafka/_rd_kafkamodule.c
```diff
@@ -0,0 +1,448 @@
...
+static PyObject *
+Consumer_stop(Consumer *self, PyObject *args) {
+    // Call stop on all partitions, then destroy all handles
+
+    PyObject *retval = Py_None;
+    if (self->rdk_topic_handle != NULL) {
+        Py_ssize_t i, len = PyList_Size(self->partition_ids);
+        for (i = 0; i != len; ++i) {
+            // Error handling here is a bit poor; we cannot bail out directly
+            // if we want to clean up as much as we can.  TODO logging
```
emmett9001: This and other logging/docstring TODOs should be cleared up.

pykafka/rdkafka/_rd_kafkamodule.c
```diff
@@ -0,0 +1,448 @@
...
+"twice, once with a `conf` list only, and again with `topic_conf` only.\n"
+"\n"
+"Repeated calls work incrementally; you can wipe configuration completely\n"
+"by calling Consumer_stop()\n";
+static PyObject *
+Consumer_configure(Consumer *self, PyObject *args, PyObject *kwds) {
+    char *keywords[] = {"conf", "topic_conf", NULL};
+    PyObject *conf = NULL;
+    PyObject *topic_conf = NULL;
+    if (! PyArg_ParseTupleAndKeywords(args,
```
emmett9001: Conditionals without curly braces around their blocks can be really dangerous. I'd suggest using `{}` around the `return NULL`.

bench/simple_consumer_bench.py
yungchin: Thanks, yes, will add documentation! These really got written in anger, so I was half thinking removing them would be cleaner, but on the other hand... we may want to benchmark again when making big or small changes that could affect performance. Keep?

pykafka/rdkafka/_rd_kafkamodule.c
```diff
@@ -0,0 +1,448 @@
...
+    char *keywords[] = {
+        "brokers",
+        "topic_name",
+        "partition_ids",
+        "start_offsets",  // same order as partition_ids
+        NULL};
+    const char *brokers = NULL;
+    const char *topic_name = NULL;
+    PyObject *partition_ids = NULL;
+    PyObject *start_offsets= NULL;
```
emmett9001: Watch out for little style snags like this. Are there any C style checkers for vim? I don't even know if C has any kind of PEP8 equivalent (I'm guessing it doesn't) but for pykafka I think it's fine to adhere to something close.

pykafka/rdkafka/_rd_kafkamodule.c
```diff
@@ -0,0 +1,448 @@
...
+    // is different than that on dealloc
+    if (self->partition_ids) {
+        set_PyRdKafkaError(RD_KAFKA_RESP_ERR__FAIL, "Already started.");
+        return NULL;
+    }
+    self->partition_ids = PySequence_List(partition_ids);
+    if (! self->partition_ids) return NULL;
+
+    // Configure and start a new RD_KAFKA_CONSUMER
+    char errstr[512];
```
emmett9001: Is 512 recommended by rdkafka?

pykafka/rdkafka/_rd_kafkamodule.c
yungchin: No, I agree, I actually never wrote in this style before. I sort of went with what I saw in a lot of https://github.com/python/cpython code, but I don't have any preference here.

pykafka/rdkafka/_rd_kafkamodule.c
```diff
@@ -0,0 +1,448 @@
...
+    self->rdk_handle = rd_kafka_new(
+            RD_KAFKA_CONSUMER, self->rdk_conf, errstr, sizeof(errstr));
+    self->rdk_conf = NULL;  // deallocated by rd_kafka_new()
+    if (! self->rdk_handle) {
+        set_PyRdKafkaError(RD_KAFKA_RESP_ERR__FAIL, errstr);
+        return NULL;
+    }
+    if (rd_kafka_brokers_add(self->rdk_handle, brokers) == 0) {
+        // XXX add brokers via conf setting instead?
```
emmett9001: The way you're adding brokers here seems fine. What would we gain by preconfiguring broker lists? This appears to match the method that's currently used in the pure python version - we get a connect string and parse it into a bunch of hosts and ports.

pykafka/rdkafka/_rd_kafkamodule.c
yungchin: Ah thanks. Even after you pointed it out I had to look three times to spot it :)

And also - great idea to look for a PEP style guide!! I just discovered https://www.python.org/dev/peps/pep-0007/ so I'll reformat everything to that.

pykafka/rdkafka/_rd_kafkamodule.c
yungchin: There isn't any doc that gives a safe size explicitly, but his `examples/rdkafka_example.c` uses 512.

pykafka/rdkafka/_rd_kafkamodule.c
```diff
@@ -0,0 +1,448 @@
...
+            set_PyRdKafkaError(rd_kafka_errno2err(errno), NULL);
+            goto fail;
+        }
+    }
+    Py_INCREF(Py_None);
+    return Py_None;
+
+fail:  ;
+    PyObject *err_type, *err_value, *err_traceback;
+    PyErr_Fetch(&err_type, &err_value, &err_traceback);
```
emmett9001: Does this line make the `goto`s above necessary? It seems like it would be cleaner to have this `fail` in its own function, but from what I can find about the `PyErr_Fetch` function, it relies on being called in the same stack frame as the error it's trying to handle. I guess not every language can be python.

pykafka/rdkafka/_rd_kafkamodule.c
```diff
@@ -0,0 +1,448 @@
...
+        PyStructSequence_SET_ITEM(retval, 2, PyLong_FromLong(
+                                  rkmessage->partition));
+        PyStructSequence_SET_ITEM(retval, 3, PyLong_FromLongLong(
+                                  rkmessage->offset));
+    } else if (rkmessage->err == RD_KAFKA_RESP_ERR__PARTITION_EOF) {
+        // Whenever we get to the head of a partition, we get this.  There may
+        // be messages available in other partitions, so if we want to match
+        // pykafka.SimpleConsumer behaviour, we ought to avoid breaking any
+        // iteration loops, and simply skip over this one altogether:
+        retval = Consumer_consume(self, args);
```
emmett9001: :+1:

pykafka/rdkafka/_rd_kafkamodule.c
yungchin: Ah yes - I couldn't make up my mind here. The only upside of adding them via the config list is that I could throw this if-block away, so less code, but on the other hand - this throws a more descriptive exception, so now that it already works anyway, I think you're right. I'll toss out the comment.

pykafka/rdkafka/_rd_kafkamodule.c
yungchin: Actually, I think it will ok in its own fail() function. The `goto` was here because in the first version of this I had some local variables that needed cleaning up (the `rd_kafka_conf_t` and so), but looking at it again now, I've actually pushed all of those onto the type object. So a separate `fail(self)` should work. I'll try that.

pykafka/rdkafka/_rd_kafkamodule.c
```diff
@@ -0,0 +1,448 @@
...
+    if (mod == NULL) return;
+
+    ConsumerStoppedException = PyErr_NewException(
+            "pykafka.rdkafka.ConsumerStoppedException", NULL, NULL);
+    if (! ConsumerStoppedException) return;
+    Py_INCREF(ConsumerStoppedException);
+    PyModule_AddObject(
+            mod, "ConsumerStoppedException", ConsumerStoppedException);
+    PyRdKafkaError = PyErr_NewException("pykafka.rdkafka.Error", NULL, NULL);
+    if (!PyRdKafkaError) return; // TODO goto error handler
```
emmett9001: Just marking this TODO, since it seems important

pykafka/rdkafka/simple_consumer.py
```diff
@@ -0,0 +1,152 @@
...
+       deviate.  In particular, here we may emit more than one message from
+       the same partition before visiting another
+    2. ignores num_consumer_fetchers: librdkafka will typically spawn at least
+       as many threads as there are kafka cluster nodes
+
+    For an overview of how configuration keys are mapped to librdkafka's, see
+    _mk_rdkafka_config_lists.
+    """
+
+    def _setup_fetch_workers(self):
```
emmett9001: It occurs to me that this function could use a more generic name in `SimpleConsumer` if it's going to be overridden this way. There aren't truly any "fetch workers" here to speak of.

pykafka/rdkafka/simple_consumer.py
```diff
@@ -0,0 +1,152 @@
...
+        rdk_consumer.start(brokers,
+                           self._topic.name,
+                           partition_ids,
+                           start_offsets)
+        return rdk_consumer
+
+    def consume(self, block=True):
+        timeout_ms = self._consumer_timeout_ms if block else 1
+        try:
+            msg = self._fetch_workers.consume(timeout_ms)
```
emmett9001: I understand why this works, but the `fetch_workers` terminology here makes this logic more confusing than it should be. I believe the naming in `SimpleConsumer` that gets overridden here should change slightly to accommodate it, if possible.

emmett9001:
This looks well-considered and I'm excited to see it in action. I left a bunch of comments, let me know if you have any questions.

pykafka/rdkafka/_rd_kafkamodule.c
yungchin: Most of that PEP7 stuff looks fine (although the requirement to write valid C89 is going to be tricky - gcc won't complain, and I won't always spot C99-isms).

The one thing I do have a bit of a beef with is also C89-related: I think putting all variable declarations at the top of the block really decreases code readability. So.... can I violate that one please :)

yungchin:
@emmett9001 huge thanks for the thorough review. I'll get it all fixed up asap.

I can't think of any good options to rename `fetch_workers` to, will have to think about that (I guess in a way, we're really using the rdkafka consumer as a "fetch worker" only, since we let pykafka handle all the metadata stuff, but yes, I see what you mean that it reads strange now).

Thanks!

bench/simple_consumer_bench.py
emmett9001: Having some standard benchmarking code around is helpful. I say clean up and keep.

pykafka/rdkafka/_rd_kafkamodule.c
emmett9001: Works for me

pykafka/rdkafka/_rd_kafkamodule.c
emmett9001: That seems fine. In that case, we can avoid checking this buffer for overflow until it's pointed out to be an issue, I suppose.

yungchin: Yes, I think that's right - all the librdkafka calls that take an errstr also require that you pass in sizeof(errstr), so the worst that could happen is that you get back a truncated error message.

pykafka/rdkafka/simple_consumer.py
yungchin: Whoever said that thing about variable naming being hard was right... I still haven't thought of a better name. In fact, given that we delegate absolutely everything except message-fetching to `SimpleConsumer`, you could even argue that `self._rdk_consumer` is really just a "fetch worker" for us :)

yungchin:
Quick status update: the only thing that I had still hoped to add on this branch, was sending rdkafka log messages into python `logging`. That turns out trickier than I'd hoped - I've been struggling with deadlocks for a whole day now (rdkafka logging calls come from non-python threads, and having these call into python (`Logger.log` et al) means you have to manage the GIL correctly, which I clearly don't - yet).

But if logging can wait for a pullreq of its own, this is otherwise ready for action.

edenhill:
@yungchin Would it help if it was possible to poll logs instead with `rd_kafka_poll()`?

yungchin:
@edenhill hey, thanks for checking this out!

If the callbacks are triggered through `rd_kafka_poll`, would that mean that they get called from that same thread, or at least, that they all return before `rd_kafka_poll` returns? That would mean we could assume, in the callback code, that the GIL was already held, so yes, in that case we could avoid all this lock-management.

AFAICT, 434e4df from yesterday also fixes it, but it does get a bit ugly with all the lock-release blocks. I haven't noticed a big performance hit, but I haven't yet benchmarked properly with it either. It would be great to be able to do away with that commit, if all rdkafka callbacks would trigger through `rd_kafka_poll` (well, except the partitioner_cb ones, but I suppose we will do partitioning in pure python before `rd_kafka_produce`).

PS - doesn't that mean you risk accumulating a lot of log messages, though?

edenhill:
It would call the log_cb in the same thread as `rd_kafka_poll()` is called, and only during poll()'s execution.
The log messages would be a bit outdated of course, but that's usually of lesser concern.

yungchin:
Ah! In that case, it would be like manna from heaven :)

edenhill:
For tracking: https://github.com/edenhill/librdkafka/issues/355

yungchin:
Ok, finally slapped "ready" on this one!

@emmett9001 @kbourgoin would be grateful if you could take it for a spin on your big boxen if time permits. It's not hooked-up yet to import anything from `pykafka.rdkafka` automatically, but it should be a drop-in replacement for `SimpleConsumer`, [like so][0]. I guess that's up for discussion: do we want to automatically have this stand in for `SimpleConsumer` whenever librdkafka is available?

[0]: https://github.com/Parsely/pykafka/blob/feature/rdkafka_extension/tests/pykafka/rdkafka/test_simple_consumer.py#L13

pykafka/rdkafka/simple_consumer.py
```diff
@@ -0,0 +1,181 @@
...
+        with self._stop_start_rdk_consumer():
+            return super(RdKafkaSimpleConsumer, self).fetch_offsets()
+
+    def reset_offsets(self, partition_offsets=None):
+        # Restart, because _rdk_consumer needs its internal offsets resynced
+        with self._stop_start_rdk_consumer():
+            super(RdKafkaSimpleConsumer, self).reset_offsets(partition_offsets)
+
+    @contextmanager
+    def _stop_start_rdk_consumer(self):
```
emmett9001: Helpful here would be a bit of explanation in the docstring of why exactly this function is necessary.

pykafka/rdkafka/simple_consumer.py
```diff
@@ -0,0 +1,181 @@
...
+            self._partitions_by_id[msg.partition_id].set_offset(msg.offset)
+        return msg
+
+    def stop(self):
+        if hasattr(self, "_rdk_consumer") and self._rdk_consumer is not None:
+            # Call _rd_kafka.Consumer.stop explicitly, so we may catch errors:
+            self._rdk_consumer.stop()
+            log.debug("Issued stop to _rdk_consumer.")
+            self._rdk_consumer = None
+        super(RdKafkaSimpleConsumer, self).stop()
```
emmett9001: These `super()` calls should probably all include `return` just for maximum futureproofing.

pykafka/rdkafka/simple_consumer.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def consume(self, block=True):
+        timeout_ms = self._consumer_timeout_ms if block else 1
+        try:
+            msg = self._rdk_consumer.consume(timeout_ms)
+        except _rd_kafka.ConsumerStoppedException:
+            raise ConsumerStoppedException
+        if msg is not None:
+            # set offset in OwnedPartition so the autocommit_worker can find it
+            self._partitions_by_id[msg.partition_id].set_offset(msg.offset)
```
emmett9001: I don't think this needs to be fixed right now, but I wish there was a clearer parallel to this line in `SimpleConsumer.consume`. This is doing essentially the same thing that happens there, but there it happens in a way that's split across multiple functions. Something for a future refactor.

pykafka/rdkafka/simple_consumer.py
yungchin: Indeed! I had to think about the answer for a bit :)

yungchin:
Have just updated this so it passes the new py3/pypy tests locally. Can't seem to figure out yet how to make tox on travis-ci find the non-default library/include paths too (what changed as part of the py3 push is that we're now on the newer "sudo-not-allowed" travis infrastructure - but that has meant that I needed to install librdkafka to a different location). Any input appreciated!!

rduplain:
From team discussion:

@yungchin says that all of the Python test cases apply and pass on this, but this needs some real-world testing.

> What is left in order to merge this?

* [ ] Use common exceptions. This could be an improvement after merge, but since exceptions are part of the API, let's do it before merge.
* [ ] Expand documentation. This one could be a separate pull request, even authored by another contributor.
* [ ] Decide: Should the import default to rdkafka if it's available, or should this instead be exposed as `use_rdkafka=True` keyword on the Kafka client? The default could change over time as things stabilize.

yungchin:
This failure on travis-ci just now deserves highlighting, because I've never seen it before but it is actually a very real failure: https://travis-ci.org/Parsely/pykafka/jobs/86804419#L637 - what I believe happens is that the `BalancedConsumer` is not finished with rebalancing, and the `consume()` call hits it at a moment where it was actually restarting the internal consumer. Presumably this shows up now because the recent merges to master have made rebalancing a slightly slower process.

What should happen in this scenario is actually the subject of #273 - that is, it's a problem that also exists for the pure-python consumer on master.

rduplain:
Current status (as stated in updated description): Depends on #238 and #273, then is ready for review.

yungchin:
Status update: with #273 merged and #238 mostly final, there are a few more things to do here.

- [ ] Make calls into the C module safe regardless the state of the objects; until now I got away with guarding against unsafe calls in the python wrapper classes, but the changes we made to BalancedConsumer (or perhaps just the extra tests we added, which trigger a lot of rebalancing and thus tear down the consumer repeatedly in quick succession) show that that's not good enough
- [ ] Following changes on #238 / #269, we need to return the `Message` object as `future.kafka_msg` here, too
- [ ] After merging master I should check that any newly added tests (particularly those from #269) also run against rdkafka. There's actually a bit of a design-smell here: every time we add tests to the suite, if we forgot to stick `use_rdkafka` flags in that test, it will only run against one or the other (it will run twice, but both times with the same backend - ugh).

amontalenti:
How likely that this can be merged by next Friday (1 week from now), @yungchin?

yungchin:
@amontalenti oooh, can I tell you tomorrow? I need to think a bit more about point 1 on the list I just posted. But let's say with luck, it's all out of the way before the end of day tomorrow, and so then it would come down to how many issues come up in reviewing. I think @emmett9001 has been over almost half of what's here, but that still leaves a lot to go.

yungchin:
@amontalenti @emmett9001 the work protecting the handle turned out quite big but it's done. I'm still working on the update to get this all conforming to the #269 changes, but expect to be able to slap "needs review" on it by tomorrow.

yungchin:
@emmett9001 can I tag you for this one? :)

In the update to conform to the #269 changes, I decided to give up on pypy support, as @amontalenti had already suggested might be pragmatic. The benefit of that is that we can avoid having to copy the payload out of the `pykafka.protocol.Message` (which on pypy is really necessary because it may opt to move the `Message` or its members around in memory at any time).

On Travis there continue to be crashes on pypy even in the consumer code, which bothers me a bit because I cannot reproduce them and I don't understand where they might come from - and thus, the worry is if they might in very rare cases also pop up on cpython (not that I've seen). Anyway, I've (reluctantly) also marked the consumer tests with `skipif` on pypy.

Some performance notes: the consumer is consistently faster than the pure-py one, with gains of ~7x with small messages against a two-node cluster on my laptop, and going up as you add nodes (I want to do more formal testing and repeat/improve on benchmarks we ran a while ago - at the time the gains with larger clusters were at 20x and up). The producer seems to be bottlenecked on the `produce()` method (which we share with the pure-py implementation), and I might have to reconsider if we don't want to do the message partitioning (which now happens in `produce()` via librdkafka's callback interface instead. That sort of change would come in a later pull request, which shouldn't break the current producer interface.

rduplain:
Up for review by @emmett9001.

yungchin:
Just a quick update on performance notes... the other I day I wrote

> The producer seems to be bottlenecked on the produce() method

It turns out that that is not true - I jumped to conclusions too quickly because the work that happens in callbacks on background threads isn't very visible when doing (too simplistic) profiling and benchmarking. What's really expensive, it turns out, is calling `Future.set_result()` (which happens in a callback on a background thread) and also `concurrent.futures.wait()`. I called `wait()` as part of the benchmark run, which turns out to more than double the run time (in retrospect that's maybe not surprising, because `Future` builds on `threading.Condition` which builds on `threading.RLock`, so waiting and setting results on a 100k long list of futures just means a lot of locking/unlocking).

If we immediately discard futures (that is, we just drop the return values from `Producer.produce()`) then `RdKafkaProducer` runs at about 2x the rate in messages/sec (for now only informally tested, with small message sizes) of `pykafka.Producer` (when we also discard futures in that benchmark).
And still, nearly 25% of that time is spent instantiating/discarding unused futures...

emmett9001:
A simple, kind of ugly way to get around the cost of instantiating unused futures is to add a kwarg that allows the instantiation of the `Future` to be turned off if the client doesn't plan on using it.

benchmark/simple_consumer_bench.py
```diff
@@ -0,0 +1,154 @@
...
+              msg_size_bytes,
+              num_consumer_fetchers,
+              queued_max_messages,
+              num_iterations,
+              filename_append="consumer_bench.json"):
+    """
+    Run a single timeit-benchmark for num_iterations of consume() calls
+
+    This is a somewhat awkward-looking runner, because we currently have an
+    issue where consumers don't get garbage-collected, and so their worker
```
emmett9001: Is this still valid? A lot of related issues were fixed recently.

pykafka/balancedconsumer.py
```diff
@@ -165,6 +172,7 @@ def __init__(self,
         self._zookeeper_connect = zookeeper_connect
         self._zookeeper_connection_timeout_ms = zookeeper_connection_timeout_ms
         self._reset_offset_on_start = reset_offset_on_start
+        self._use_rdkafka = rdkafka and use_rdkafka
```
emmett9001: To avoid confusion, I think this should raise an exception if `rdkafka == False and use_rdkafka == True`.

pykafka/rdkafka/producer.py
```diff
@@ -0,0 +1,163 @@
+from concurrent import futures
```
emmett9001: We should pull out the producer futures logic prior to merge, since that interface is still being debated and experimented with.

pykafka/rdkafka/producer.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def start(self):
+        if not self._running:
+            brokers = b','.join(b.host + b":" + get_bytes(str(b.port))
+                               for b in self._cluster.brokers.values())
+            conf, topic_conf = self._mk_rdkafka_config_lists()
+
+            self._rdk_producer = _rd_kafka.Producer()
+            self._rdk_producer.configure(conf=conf)
+            self._rdk_producer.configure(topic_conf=topic_conf)
```
emmett9001: An additional comment here about the order of these `configure` calls mattering would be helpful.

pykafka/rdkafka/_rd_kafkamodule.c
```diff
@@ -0,0 +1,1175 @@
...
+    return future;
+failed:
+    Py_XDECREF(wref);
+    Py_XDECREF(future);
+    return NULL;
+}
+
+
+static void
+Producer_delivery_report_callback(rd_kafka_t *rk,
```
emmett9001: I guess this method will stay around, but it will not manipulate futures anymore, instead maybe pushing failures onto a queue.

pykafka/rdkafka/simple_consumer.py
```diff
@@ -0,0 +1,239 @@
...
+    For an overview of how configuration keys are mapped to librdkafka's, see
+    _mk_rdkafka_config_lists.
+    """
+    def __init__(self,
+                 topic,
+                 cluster,
+                 consumer_group=None,
+                 partitions=None,
+                 fetch_message_max_bytes=1024 * 1024,
+                 num_consumer_fetchers=1,
```
emmett9001: This should probably just be removed if it's ignored.

pykafka/rdkafka/simple_consumer.py
```diff
@@ -0,0 +1,239 @@
...
+    1. rotating over partitions: while message ordering within partitions is
+       conserved (of course!), the order in which partitions are visited will
+       deviate.  In particular, here we may emit more than one message from
+       the same partition before visiting another
+    2. ignores num_consumer_fetchers: librdkafka will typically spawn at least
+       as many threads as there are kafka cluster nodes
+
+    For an overview of how configuration keys are mapped to librdkafka's, see
+    _mk_rdkafka_config_lists.
+    """
```
emmett9001: Docstrings for these kwargs would be great, especially if they explained the mapping to librdkafka configs.

setup.py
```diff
@@ -31,6 +33,7 @@ def get_version():
     'futures',
     'kazoo',
     'tabulate',
+    'futures',
```
emmett9001: Probably needs to be removed

pykafka/rdkafka/simple_consumer.py
emmett9001: Argh, can't remove it because it's passed in when instantiating the internal consumer in a `BalancedConsumer`. A possible option to make this signature more explicit is to remove `num_consumer_fetchers` and add an unused `**kwargs` to the end of the args list, but that seems far worse for comprehensibility. The only other alternatives I'm aware of involve dynamically building the kwargs list that's used on the internal consumer. Maybe that's actually better, but not a pressing issue. For now, the docstring handles it just fine.

emmett9001:
These changes look great. The main reason I'm hesitant to merge at the moment is that this branch still uses producer futures, which we've decided are not actually a good idea. Once we've replaced that mechanism on this branch and in master, we should be safe to merge this in.

benchmark/simple_consumer_bench.py
yungchin: No it's not. Actually the whole benchmark dir has gone very stale, and as we discussed over email back then, the methods could be better. Should I remove it all for now, and commit it when there's a fresh and more fair version?

pykafka/rdkafka/_rd_kafkamodule.c
yungchin: Yes, that's my plan. I'm leaving it untouched for now, it will just fail writing to the absent future, and the new version is coming up very soon.

pykafka/rdkafka/simple_consumer.py
yungchin: I had some dynamically built kwargs lists in the now-archived cffi branch, but @kbourgoin hated them because they were breaking command completion in things like ipython. If there's a way that makes it dynamic and that can also be grokked by fancy interpreters, that would be neat though.

pykafka/rdkafka/simple_consumer.py
yungchin: Still working on this one - coming to a pullreq near you very soon!

ottomata:
> this branch still uses producer futures, which we've decided are not actually a good idea
Just curious, because I've only recently started following github here.  Why is it a bad idea?  Link to discussion I missed? :)

yungchin:
Hi @ottomata I basically missed the fact that if you want to use this at high-throughput, you'd have tens of thousands of futures around. As noted [here][0] that gets a bit expensive - both in terms of time, and memory footprint actually (each future weighs in at well over 2kB if you take all the things it comes with (a `threading.Condition` mainly) into account.

I'm working on a revised version that exposes a thread-local queue to the user, which should hopefully enable delivery-reporting without this much overhead (futures slowed down the rdkafka producer by a little over 100%).

[0]: https://github.com/Parsely/pykafka/pull/176#issuecomment-155081460

ottomata:
Oh the answer was up above even on this thread, thank you! :p

yungchin:
@emmett9001 thanks for reviewing this pile of code once more!

I've removed the futures from the interface here, too, and merged latest master. Interestingly the merge from master revealed a bug that was previously invisible, where queues in the producer wouldn't always get flushed during finalisation (180e876).

pykafka/rdkafka/simple_consumer.py
yungchin: I've put further annotations and some improvements in the configuration mapping. It's not literally a docstring, but personally I'd feel that duplicating the mapping into a docstring (that's not code) would eventually leave it stale. If I can make readability improvements to the actual mapping in `_mk_rdkafka_config_lists` instead, so that that is more self-documenting, I think I'd prefer that option.

benchmark/simple_consumer_bench.py
emmett9001: I don't think it needs to be removed, but updating it soon sounds good.

pykafka/rdkafka/_rd_kafkamodule.c
```diff
@@ -0,0 +1,1182 @@
...
+    }
+    return 0;
+}
+
+
+/* Get shared lock and optionally check handle is running.  Returns non-zero
+ * if error has been set */
+static int
+RdkHandle_safe_lock(RdkHandle *self, int check_running)
+{
```
emmett9001: I think all (or at least many) of the functions in this file would benefit from docstrings indicating how they fit into the overall operation of the producer/consumer.

pykafka/rdkafka/simple_consumer.py
```diff
@@ -0,0 +1,240 @@
...
+            log.debug("Exiting RdKafkaSimpleConsumer poller thread.")
+
+        self._stop_poller_thread.clear()
+        self._poller_thread = self._cluster.handler.spawn(
+            poll, args=(self._rdk_consumer, self._stop_poller_thread))
+
+    def consume(self, block=True):
+        timeout_ms = self._consumer_timeout_ms if block else 1
+        try:
+            msg = self._rdk_consumer.consume(timeout_ms)
```
emmett9001: When I run an `RdKafkaSimpleConsumer` using `consume(block=True)`, Ctrl+C doesn't work to kill the process from my terminal. When I put a log directly before this line, that problem goes away. I think this is a case of the interpreter not getting a chance to stop executing C code and catch the `SIGINT`. If there's a way for us to discover how many messages are currently ready for consumption, we can do something similar to the use of `_messages_arrived` in `SimpleConsumer`. However, I'm not sure if that's possible. Either way, we do need to allow the process to be easily killed when `block=True`.

Investigating this has also shown me that this `consume()` method violates the semantics of its superclass in that it is not guaranteed to return a non-`None` message if `block=True`. Take a look at `SimpleConsumer.consume` - it's an important property of that method that allows `__iter__` to work the way it does.

yungchin: Argh, thanks, you're right, that's not very user friendly. If I put in a log line where you said it still blocks for me (unless I'm very quick). I think I'll just add another level of looping, so that we always call the C function with `timeout_ms=500`, or something like that, and the outer loop checks for the real timeout. Coming up.

Re the iteration: when I try it here, it behaves exactly as you say it should? If I leave the default `consumer_timeout_ms=-1` it blocks forever at the end of the topic, and if I put it to something like `5000` I indeed see it block 5 seconds at the end of the topic.

pykafka/rdkafka/_rd_kafkamodule.c
yungchin: Thanks, yes, I'll get on it.

pykafka/rdkafka/simple_consumer.py
emmett9001: Looks like `__iter__` works with the latest version of the code. As far as I'm concerned this branch is ok to merge now, since it doesn't break compatibility with anything I can see. Some documentation will be good, though.

yungchin:
@emmett9001 ok, I've finally gotten on top of the heisenbug that was #260 - as we had suspected, the problem was in the testing code (and in retrospect quite obvious). After some recent roll-backs on master, which changed timings during rebalancing, that bug basically triggered on 90% of test runs on this branch (mostly because an rdkafka consumer is a bit slower to restart than the pure pykafka consumer), so it was suddenly easy to debug.

I've added real docstrings to the C extension methods. Clearly codecov is not impressed with coverage on this branch, but to be fair the missed lines would have mostly been covered by the tests in the producer-futures work that we backed out. When that's back in, coverage should be pretty good.

emmett9001:
It looks to me like all of the pieces are in place for a `2.1.0` in the very near future. This branch, #324, and #344 are all ready to merge as far as I'm concerned. @yungchin let me know if there's anything you think I'm forgetting here, or extra bits you think are important to clean up before a new version. If not, I'll aim for tomorrow.

Issue #124, opened 2015-02-05
Python AsyncProducer Implementation


kbourgoin:
Moving this implementation to 0.8.2-02 because we at least have a sync producer implementation in Python.

emmett9001:
Is there a good reference point for the design of an async producer? Maybe another project that implements something similar or a kafka project design document. Given that, I can probably put one together.

kbourgoin:
That's the best idea I have.  The need for it was taken from the fact that there is a Scala one, so I've got no better source

yungchin:
The librdkafka docs may be of interest too: https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#producer-api

The approach there is that the underlying producer is always async; implementing a sync producer on top of it basically means producing through the async producer, and then waiting for the message delivery report(s) before returning:
https://github.com/edenhill/librdkafka/wiki/Sync-producer
(we did exactly that on the cffi branch: https://github.com/Parsely/pykafka/blob/feature/archived_rdkafka_cffi/pykafka/rdkafka/producer.py#L71 )

emmett9001:
Started work on this [here](https://github.com/Parsely/pykafka/tree/feature/async-producer). It's still very prototypal, but I figured out some basic things. It's nowhere near ready for review.

emmett9001:
Pull requested [here](https://github.com/Parsely/pykafka/pull/177). It's ready for initial review from @kbourgoin and/or @yungchin.

emmett9001:
Ready for final review and merge

emmett9001:
Merged!

PR #177, opened 2015-06-03
Asynchronous Producer

emmett9001:
This pull request updates the `Producer` to use a thread per broker instead of the main thread for all brokers.

[This code](https://gist.github.com/emmett9001/ef40bac8ce9a160377af) can be used to test this producer's throughput. Here are some [throughput test results with this script](https://gist.github.com/emmett9001/a3a7f0e9f52ac8471912) (how many times per second each producer can return from `produce()`). The old synchronous producer was doing about 4500 messages per second.

Connects to #124.

Todo:
*  [x] Unit test exception raised on worker thread
*  [x] Monitor thread death
*  [x] Abstract thread management logic into the `QueueInfo` class
*  [x] Make `Producer` as small as possible by having it be a special case of `AsyncProducer`
*  [x] Tests

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -65,7 +65,7 @@ class Producer(object):
                  ack_timeout_ms=10 * 1000,
                  max_queued_messages=100000,
                  min_queued_messages=70000,
-                 linger_ms=0,
+                 linger_ms=5 * 1000,
                  block_on_queue_full=True,
                  sync=False):
         """Instantiate a new AsyncProducer
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -135,7 +135,7 @@ class Producer():
                 for partition, presponse in partitions.iteritems():
                     if presponse.err == 0:
                         if q_info is not None:
-                            q_info.decrement_inflight()
+                            q_info.messages_inflight -= 1
                         continue  # All's well
                     if presponse.err == UnknownTopicOrPartition.ERROR_CODE:
                         log.warning('Unknown topic: %s or partition: %s. '
@@ -146,7 +146,7 @@ class Producer():
                         # Update cluster metadata to get new leader
                         self._cluster.update()
                         if q_info is not None:
-                            q_info.update_leader()
+                            q_info.producer._update_leaders()
                     elif presponse.err == RequestTimedOut.ERROR_CODE:
                         log.warning('Produce request to %s:%s timed out. '
                                     'Retrying.', broker.host, broker.port)
@@ -273,23 +273,6 @@ class QueueInfo(_QueueInfo):
         ins.messages_inflight = messages_inflight
         return ins
 
-    def decrement_inflight(self, acquire=True):
-        if acquire:
-            self.lock.acquire()
-        self.messages_inflight -= 1
-        if acquire:
-            self.lock.release()
-
-    def increment_inflight(self, acquire=True):
-        if acquire:
-            self.lock.acquire()
-        self.messages_inflight += 1
-        if acquire:
-            self.lock.release()
-
-    def update_leader(self):
-        self.producer._update_leaders()
-
 
 class AsyncProducer():
     """
@@ -375,7 +358,7 @@ class AsyncProducer():
     def stop(self):
         self._running = False
         last_log = time.time()
-        while any(q_info.messages_inflight
+        while any(q_info.messages_inflight > 0
                   for leader, q_info in self._workers_by_leader.iteritems()):
             if time.time() - last_log >= .5:
                 log.info("Waiting for all worker threads to exit")
@@ -439,7 +422,7 @@ class AsyncProducer():
             with q_info.lock:
                 to_extend = messages[:self._max_queued_messages - len(q_info.queue)]
                 for message in to_extend:
-                    q_info.increment_inflight(acquire=False)
+                    q_info.messages_inflight += 1
                 q_info.queue.extendleft(to_extend)
                 log.debug("Enqueued %d messages for broker %d",
                           len(messages), broker_id)
@@ -538,13 +521,20 @@ class AsyncProducer():
         self._produce(self._producer._partition_messages(messages))
 
     def _update_leaders(self):
+        """Ensure each message in each queue is in the queue owned by its
+            partition's leader
+
+        This function empties all broker queues, maps their messages to the
+        current leaders for their partitions, and enqueues the messages in
+        the appropriate queues.
+        """
         # empty queues and figure out updated partition leaders
         new_queue_contents = defaultdict(list)
         for broker, q_info in self._workers_by_leader.iteritems():
             q_info.lock.acquire()
             messages = [q_info.queue.pop() for _ in xrange(len(q_info.queue))]
             for (key, value), partition_id in messages:
-                q_info.decrement_inflight(acquire=False)
+                q_info.messages_inflight -= 1
                 new_leader = self._topic.partitions[partition_id].leader
                 new_queue_contents[new_leader.id].append(
                     ((key, value), partition_id))
@@ -556,6 +546,17 @@ class AsyncProducer():
             messages = new_queue_contents[broker.id]
             q_info.queue.extendleft(messages)
             for message in messages:
-                q_info.increment_inflight(acquire=False)
+                q_info.messages_inflight += 1
+
+            # manage condition variable state
+            if len(q_info.queue) >= self._max_queued_messages:
+                q_info.slot_available.clear()
+                q_info.flush_ready.set()
+            elif len(q_info.queue) == 0:
+                q_info.slot_available.set()
+                q_info.flush_ready.clear()
+            else:
+                q_info.slot_available.set()
+
             # when we get here, we're finally done with this broker
             q_info.lock.release()
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -480,11 +480,11 @@ class OwnedBroker(object):
     def resolve_event_state(self):
         """Invariants for the Event variables used for thread synchronization
         """
-        if len(self.queue) >= self.producer._max_queued_messages:
-            self.slot_available.clear()
-            self.flush_ready.set()
-        elif len(self.queue) == 0:
+        if len(self.queue) < self.producer._max_queued_messages:
             self.slot_available.set()
-            self.flush_ready.clear()
         else:
-            self.slot_available.set()
+            self.slot_available.clear()
+        if len(self.queue) >= self.producer._min_queued_messages:
+            self.flush_ready.set()
+        else:
+            self.flush_ready.clear()
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -252,12 +252,14 @@ class QueueInfo(_QueueInfo):
         queue
     :type lock: threading.Lock
     :ivar flush_ready: A condition variable that indicates that the queue is
-        ready to be flushed
+        ready to be flushed via requests to the broker
     :type flush_ready: threading.Condition
     :ivar slot_available: A condition variable that indicates that there is
         at least one position free in the queue for a new message
     :type slot_available: threading.Condition
-    :ivar queue: The message queue for this broker
+    :ivar queue: The message queue for this broker. Contains messages that have
+        been supplied as arguments to `produce()` waiting to be sent to the
+        broker.
     :type queue: collections.deque
     :ivar messages_inflight: A counter indicating how many messages have been
         enqueued for this broker and not yet sent in a request.
@@ -350,12 +352,14 @@ class AsyncProducer():
         self.start()
 
     def start(self):
+        """Connect to brokers and start worker threads"""
         if not self._running:
             self._setup_internal_producer()
             self._setup_workers()
             self._running = True
 
     def stop(self):
+        """Mark as stopped and wait for all inflight messages to send"""
         self._running = False
         last_log = time.time()
         while any(q_info.messages_inflight > 0
@@ -365,13 +369,16 @@ class AsyncProducer():
                 last_log = time.time()
 
     def __enter__(self):
+        """Context manager entry point - start the producer"""
         self.start()
         return self
 
-    def __exit__(self, type, value, traceback):
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Context manager exit point - stop the producer"""
         self.stop()
 
     def _setup_internal_producer(self):
+        """Instantiate the internal producer"""
         self._producer = Producer(
             self._cluster,
             self._topic,
@@ -387,7 +394,7 @@ class AsyncProducer():
         """Spawn workers for connected brokers
 
         Creates one :class:`queue.Queue` instance and one worker thread per
-        broker. The queue holds requests waiting to be sent, and the worker
+        broker. The queue holds messages waiting to be sent, and the worker
         thread reads the end of the queue and sends requests.
         """
         # _workers_by_leader maps each broker id to a QueueInfo instance
@@ -400,6 +407,11 @@ class AsyncProducer():
                 self._workers_by_leader[partition.leader.id] = q_info
 
     def _produce(self, message_partition_tups):
+        """Enqueue a set of messages for the relevant brokers
+
+        :param message_partition_tups: Messages with partitions assigned.
+        :type message_partition_tups: tuples of ((key, value), partition_id)
+        """
         # group messages by destination broker
         messages_by_leader = defaultdict(list)
         for ((key, value), partition_id) in message_partition_tups:
@@ -440,20 +452,10 @@ class AsyncProducer():
         :param broker: The broker for which to instantiate this queue and worker
         :type broker: :class:`pykafka.broker.Broker`
         """
-        # this lock is used to ensure only one thread is manipulating these
-        # Events or the queue at a given moment
         lock = threading.Lock()
-        # When flush_ready is set, the queue is ready to be flushed to the
-        # broker
         flush_ready = threading.Event()
-        # when slot_available is set, there is space in the queue for at least
-        # one message
         slot_available = threading.Event()
-        # this queue contains the messages that have been produced and are
-        # waiting to be sent to the broker
         message_queue = deque()
-        # counter that indicates the number of messages that have been
-        # enqueued and not yet sent successfully.
         messages_inflight = 0
 
         q_info = QueueInfo(lock=lock,
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -21,7 +21,6 @@ __all__ = ["Producer", "AsyncProducer"]
 from collections import defaultdict, deque
 import itertools
 import logging
-import threading
 import time
 
 from .common import CompressionType
@@ -425,13 +424,10 @@ class OwnedBroker(object):
 
     :ivar lock: The lock used to control access to shared resources for this
         queue
-    :type lock: threading.Lock
     :ivar flush_ready: A condition variable that indicates that the queue is
         ready to be flushed via requests to the broker
-    :type flush_ready: threading.Condition
     :ivar slot_available: A condition variable that indicates that there is
         at least one position free in the queue for a new message
-    :type slot_available: threading.Condition
     :ivar queue: The message queue for this broker. Contains messages that have
         been supplied as arguments to `produce()` waiting to be sent to the
         broker.
@@ -445,9 +441,9 @@ class OwnedBroker(object):
     def __init__(self, producer, broker):
         self.producer = producer
         self.broker = broker
-        self.lock = threading.Lock()
-        self.flush_ready = threading.Event()
-        self.slot_available = threading.Event()
+        self.lock = self.producer._cluster.handler.Lock()
+        self.flush_ready = self.producer._cluster.handler.Event()
+        self.slot_available = self.producer._cluster.handler.Event()
         self.queue = deque()
         self.messages_inflight = 0
 
@@ -456,7 +452,7 @@ class OwnedBroker(object):
                 batch = self.flush(self.producer._linger_ms)
                 if batch:
                     self.producer._prepare_request(batch, self)
-        log.info("Starting new produce worker thread for broker %s", broker.id)
+        log.info("Starting new produce worker for broker %s", broker.id)
         self.producer._cluster.handler.spawn(queue_reader)
 
     @property
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -326,7 +326,11 @@ class AsyncProducer(Producer):
                     request, attempt = request_queue.get_nowait()
                 except Empty:
                     continue
-                self._send_request(broker, request, attempt)
+                try:
+                    self._send_request(broker, request, attempt)
+                except ProduceFailureError as e:
+                    log.error("Producer error: %s", e)
+
         request_queue = self._cluster.handler.Queue()
         log.info("Starting new produce worker thread for broker %s", broker.id)
         self._cluster.handler.spawn(worker)
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -1,6 +1,6 @@
 from __future__ import division
 """
-Author: Keith Bourgoin
+Author: Emmett Butler, Keith Bourgoin
 """
 __license__ = """
 Copyright 2015 Parse.ly, Inc.
@@ -25,9 +25,15 @@ import time
 
 from .common import CompressionType
 from .exceptions import (
-    UnknownTopicOrPartition, NotLeaderForPartition, RequestTimedOut,
-    ProduceFailureError, SocketDisconnectedError, InvalidMessageError,
-    InvalidMessageSize, MessageSizeTooLarge, ProducerQueueFullError
+    InvalidMessageError,
+    InvalidMessageSize,
+    MessageSizeTooLarge,
+    NotLeaderForPartition,
+    ProduceFailureError,
+    ProducerQueueFullError,
+    RequestTimedOut,
+    SocketDisconnectedError,
+    UnknownTopicOrPartition
 )
 from .partitioners import random_partitioner
 from .protocol import Message, ProduceRequest
@@ -39,7 +45,7 @@ log = logging.getLogger(__name__)
 class Producer(object):
     """
     This class implements asynchronous producer logic similar to that found in
-    the JVM driver. In inherits from the synchronous implementation.
+    the JVM driver.
     """
     def __init__(self,
                  cluster,
@@ -69,18 +75,18 @@ class Producer(object):
         :param retry_backoff_ms: The amount of time (in milliseconds) to
             back off during produce request retries.
         :type retry_backoff_ms: int
-        :param required_acks: How many other brokers must have committed the
-            data to their log and acknowledged this to the leader before a
-            request is considered complete?
+        :param required_acks: The number of other brokers must have committed
+            the data to their log and acknowledged this to the leader before a
+            request is considered complete
         :type required_acks: int
-        :param ack_timeout_ms: Amount of time (in milliseconds) to wait for
+        :param ack_timeout_ms: The amount of time (in milliseconds) to wait for
             acknowledgment of a produce request.
         :type ack_timeout_ms: int
         :param max_queued_messages: The maximum number of messages the producer
             can have waiting to be sent to the broker. If messages are sent
             faster than they can be delivered to the broker, the producer will
             either block or throw an exception based on the preference specified
-            by block_on_queue_full.
+            with block_on_queue_full.
         :type max_queued_messages: int
         :param linger_ms: This setting gives the upper bound on the delay for
             batching: once the producer gets max_queued_messages worth of
@@ -94,6 +100,7 @@ class Producer(object):
             broker contains max_queued_messages, we must either stop accepting
             new messages (block) or throw an error. If True, this setting
             indicates we should block until space is available in the queue.
+            If False, we should throw an error immediately.
         :type block_on_queue_full: bool
         """
         self._cluster = cluster
@@ -108,7 +115,6 @@ class Producer(object):
         self._linger_ms = linger_ms
         self._block_on_queue_full = block_on_queue_full
         self._exception = None
-
         self._running = False
         self.start()
 
@@ -119,17 +125,22 @@ class Producer(object):
             id_=hex(id(self))
         )
 
-    @property
-    def messages_inflight(self):
-        """Indicates whether any messages are currently "in flight"
+    def __enter__(self):
+        """Context manager entry point - start the producer"""
+        self.start()
+        return self
 
-        "In flight" messages are those that have been used in calls to `produce`
-        and have not yet been dequeued and sent to the broker
-        """
-        return any(q.message_inflight for q in self._owned_brokers.itervalues())
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Context manager exit point - stop the producer"""
+        if self._exception is not None:
+            log.exception(traceback)
+            self.stop(wait=False)
+            raise self._exception
+        else:
+            self.stop()
 
     def start(self):
-        """Connect to brokers and start worker threads"""
+        """Set up data structures and start worker threads"""
         if not self._running:
             self._owned_brokers = {}
             for partition in self._topic.partitions.values():
@@ -137,6 +148,8 @@ class Producer(object):
                     self._owned_brokers[partition.leader.id] = OwnedBroker(
                         self, partition.leader)
             self._running = True
+        if self._exception is not None:
+            raise self._exception
 
     def stop(self, wait=True):
         """Mark as stopped and wait for all inflight messages to send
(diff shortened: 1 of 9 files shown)


pykafka/producer.py
```diff
@@ -232,13 +218,27 @@ def _produce(self, message_partition_tups, attempt):
...
+                self._request_built_callback(leader,
+                                             requests.pop(leader.id),
+                                             attempt)
 
         # Send any still not sent
         for leader, req in requests.iteritems():
-            self._send_request(leader, req, attempt)
+            self._request_built_callback(leader, req, attempt)
+
+    def _request_built_callback(self, leader, request, attempt):
```
emmett9001: Bad name, sounds like `request` is a verb.

pykafka/producer.py
```diff
@@ -250,3 +250,68 @@ def produce(self, messages):
...
+        Enqueues the request in the queue associated with `leader`
+
+        :param leader: The broker to which to send the request
+        :type leader: :class:`pykafka.broker.Broker`
+        :param request: The request to send
+        :type request: :class:`pykafka.protocol.ProduceRequest`
+        :param attempt: The attempt number for this send attempt
+        :type attempt: int
+        """
+        self._queues_by_leader[leader.id].put((request, attempt))
```
emmett9001: Should we block here if the queue is too full? That would fix the misleading throughput numbers issue.

yungchin: That sounds useful. I think the java producer has a `queue.buffering.max.messages` config parameter for that - librdkafka has the same.

pykafka/producer.py
```diff
@@ -250,3 +250,68 @@ def produce(self, messages):
...
+        :type broker: :class:`pykafka.broker.Broker`
+        """
+        def worker():
+            while True:
+                time.sleep(.0001)
+                try:
+                    request, attempt = request_queue.get_nowait()
+                except Empty:
+                    continue
+                self._send_request(broker, request, attempt)
```
yungchin: At some value of `attempt` this might raise a `ProduceFailureError` - would that kill the worker?

yungchin:
Hey, cool stuff!

About the error handling in the worker: I guess I've sort of half a feature request in mind while reading this, so there's a hidden agenda :)

The JVM async producer apparently has this option now to confirm delivery of messages, [including their offset in the partition][0], it would be really nice to have something similar. So when catching the `ProduceFailureError` we'd optionally call such an onCompletion callback with the exception, something like that?

The librdkafka producer also has such a delivery callback interface, so there would be no problem on that end.

[0]: http://kafka.apache.org/082/javadoc/org/apache/kafka/clients/producer/Callback.html

yungchin:
Another question: I noticed the ThreadingHandler runs threads in daemon mode - does that mean that if you async-produce some messages, then exit the main thread, there's a chance the worker dies before the messages are sent? (I'm now thinking of something like using `atexit` to check if the queues are empty, but that sounds a bit garish maybe.)

yungchin:
Ok, really cool. I would have thought it would take much more separate code to make an async producer!

So seeing how clean this all is, I just had a thought... I was following the code around where, for `attempt > 0` this is now calling `Producer._produce` from the worker thread (and I was suspicious if that was safe, but it looks like it is all great). So given that that works, actually would it be doable to always have `Producer._produce` run from the worker thread, ie also on the first attempt?

The benefit would be, that then you could do message batching (ie if I'm a user who calls `produce()` with one message at a time, you could still pack a whole bunch of them into one broker request).

Basically, the changes would be something like

 * AsyncProducer would override `_produce()` instead of  `_request_built_callback()`
 * this `_produce()` would stick its `messages` on partition Queues, without building requests for them
 * you'd have the worker threads getting multiple sets of messages off the queue, splicing them together, and calling `Producer._produce` on them for all the request building and sending stuff

Not sure if that makes sense - I need some sleep ;)

pykafka/producer.py
```diff
@@ -250,3 +250,68 @@ def produce(self, messages):
...
         # a random partition distribution changing that on the retry.
         self._produce(self._partition_messages(messages), 0)
+
+
+class AsyncProducer(Producer):
+    """
+    This class implements the asynchronous producer logic found in the
+    JVM driver. In inherits from the synchronous implementation.
+    """
+    def __init__(self, *args, **kwargs):
```
kbourgoin: -1 to not actually having args spelled out. makes it annoying to use if you've got an editor which is doing auto-complete like vim-jedi

pykafka/producer.py
```diff
@@ -250,3 +250,68 @@ def produce(self, messages):
...
+        given broker.
+
+        :param broker: The broker for which to instantiate this queue and worker
+        :type broker: :class:`pykafka.broker.Broker`
+        """
+        def worker():
+            while True:
+                time.sleep(.0001)
+                try:
+                    request, attempt = request_queue.get_nowait()
```
kbourgoin: I don't get why the `nowait` followed by sleep + continue. It's just going to create more busy-waiting like we had with the consumer. The queue is designed for long polling that won't chew through CPU just waiting around.

pykafka/producer.py
```diff
@@ -250,3 +250,68 @@ def produce(self, messages):
...
+                except Empty:
+                    continue
+                self._send_request(broker, request, attempt)
+        request_queue = Queue()
```
kbourgoin: This should use the Queue in `self._cluster.handler` so that we can later have it switch between the `threading` and `gevent` implementations.

pykafka/producer.py
emmett9001: So the dequeue operation here should use a timeout? That makes sense.

kbourgoin:
Took a quick look at the implementation and a few things come to mind:

1. You need to monitor thread death. If we're using `threading` (since that's the only handler supported right now) the thread will die silently and you'll just stop producing messages. This is usually solved by creating a registry of the threads running and having a monitor thread which will check their state every so often. Then it can either restart the dead thread or raise the error to the main thread (a tricky thing in its own right).

2. The subclass is very heavily dependent on the implementation of the main class. This isn't technically wrong, but it makes following the implementation of the async producer that much harder.  I'll take another look to puzzle it out, but it feels like a case where trying to avoid duplicated/similar code has made something much more confusing than it needs to be.

3. How does this respond to leaders changing?  If you've got a queue per broker and the broker is no longer correct, will it be shuffled to the right now?  This could be related to point 2, where I don't think I fully understand everything that's going on.

pykafka/producer.py
kbourgoin: No, it should block forever and the thread should stay asleep as long as there is no data coming in.  Since it's on a separate thread, is it a problem for it to sit there?

emmett9001: Ah I see - that's not a problem.

pykafka/producer.py
```diff
@@ -241,3 +242,122 @@ def produce(self, messages):
...
+
+        Creates a :class:`queue.Queue` instance and a worker thread for the
+        given broker.
+
+        :param broker: The broker for which to instantiate this queue and worker
+        :type broker: :class:`pykafka.broker.Broker`
+        """
+        def worker():
+            while True:
+                time.sleep(.0001)
```
kbourgoin: Why is there a sleep here? Needs a comment to explain why it's necessary, IMO.

pykafka/producer.py
```diff
@@ -241,3 +242,122 @@ def produce(self, messages):
...
+                try:
+                    request, attempt = request_queue.get_nowait()
+                    self._send_request(broker, request, attempt)
+                except Empty:
+                    continue
+                except ProduceFailureError as e:
+                    log.error("Producer error: %s", e)
```
kbourgoin: This doesn't seem right to me. Why would we log the error and then just continue on? Signaling failure from the thread is annoying, but I think this needs to be bubbled up to the user.

kbourgoin:
Connects to #124.

emmett9001:
Another note for the next time this gets worked on (hopefully tomorrow): @kbourgoin's main complaint about the design of this class so far is that batching across calls to `produce` is not supported. We need to be able to specify a batch size - the size of each message batch sent per request to kafka. This batch will build up in memory across multiple calls to `produce` and will send either when the batch size has been reached or a message in the batch has been waiting for a long enough time.
The current implementation does support batching, but only when all of the messages in the batch were supplied in the same `produce` call. This will be my primary focus when I work on this next, which I hope happens all day tomorrow.

emmett9001:
My latest changes here implement a non-busy waiting scheme similar to the one from @kbourgoin's deque example that uses condition variables to perform signaling between the queue reader thread and the writer thread, which in this case happens to be the main thread.
I'm currently trying to figure out how to handle the leader changing for a partition. Since the queue for that broker will probably be full of messages when this happens, we'll need to find all of the messages in that queue that are for that partition and move them to the queue for that partition's leader.

pykafka/producer.py
```diff
@@ -18,16 +18,17 @@
 limitations under the License.
 """
 __all__ = ["Producer", "AsyncProducer"]
+from collections import defaultdict, deque, namedtuple
 import itertools
 import logging
+import threading
```
kbourgoin: Don't use `threading`. The handler exists to satisfy all of these objects while allowing us to swap out the implementation when we work on gevent compatibility. (e.g. `handler.Lock`)

pykafka/producer.py
```diff
@@ -120,6 +105,11 @@ def _send_request(self, broker, req, attempt):
         :type req: :class:`pykafka.protocol.ProduceRequest`
         :param attempt: The current attempt count. Used for retry logic
         :type attempt: int
+        :param q_info: A QueueInfo instance containing information
+            about the queue from which the messages in `req` were removed.
+            Only used when called from the
+            :class:`pykafka.producer.AsyncProducer`
```
kbourgoin: If this is only relevant when used from `AsyncProducer` why is it in `Producer`?  This smells like something where Async should just have its own implementation instead of trying super hard to share code.

pykafka/producer.py
```diff
@@ -241,3 +237,326 @@ def produce(self, messages):
...
+                lock,
+                flush_ready,
+                slot_available,
+                queue,
+                producer,
+                messages_inflight):
+        ins = super(QueueInfo, cls).__new__(
+            cls, lock, flush_ready, slot_available, queue, producer)
+        # store messages_inflight separately to allow updating
+        ins.messages_inflight = messages_inflight
```
kbourgoin: If you have a mutable value, then an object is preferable to a namedtuple. Otherwise, it's not really a tuple anymore. If you're worried about the size of the object, you can use `__slots__`, but realistically there aren't going to be many of these hanging around, so it's probably not necessary.

(disclaimer: i realize protocol.py has this otherwise. It's something I've been working on and changed my mind on recently.)

pykafka/producer.py
```diff
@@ -241,3 +237,326 @@ def produce(self, messages):
...
         # a random partition distribution changing that on the retry.
         self._produce(self._partition_messages(messages), 0)
+
+
+_QueueInfo = namedtuple('QueueInfo', ['lock', 'flush_ready',
+                                      'slot_available', 'queue',
+                                      'producer'])
+
+
+class QueueInfo(_QueueInfo):
```
kbourgoin: It's not really "info" is it? It's really an async queue, since it contains the deque we're using as a queue.

pykafka/producer.py
```diff
@@ -241,3 +237,326 @@ def produce(self, messages):
...
+                messages_inflight):
+        ins = super(QueueInfo, cls).__new__(
+            cls, lock, flush_ready, slot_available, queue, producer)
+        # store messages_inflight separately to allow updating
+        ins.messages_inflight = messages_inflight
+        return ins
+
+
+class AsyncProducer():
+    """
```
kbourgoin: [nitpick] First line of the docstring should be a short description of the object, not a blank line.

pykafka/producer.py
```diff
@@ -241,3 +237,326 @@ def produce(self, messages):
...
+        # _workers_by_leader maps each broker id to a QueueInfo instance
+        # containing references to that broker's queue and its associated
+        # synchronization primitives
+        self._workers_by_leader = {}
+        for partition in self._topic.partitions.values():
+            if partition.leader.id not in self._workers_by_leader:
+                q_info = self._setup_queue(partition.leader)
+                self._workers_by_leader[partition.leader.id] = q_info
+
+    def _setup_queue(self, broker):
```
kbourgoin: The implementation no longer requires thread-per-broker because we already have a network thread for each broker. Adding more here just adds unnecessary complexity.

pykafka/producer.py
```diff
@@ -241,3 +237,326 @@ def produce(self, messages):
...
+        """Context manager entry point - start the producer"""
+        self.start()
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Context manager exit point - stop the producer"""
+        self.stop()
+
+    def _setup_internal_producer(self):
+        """Instantiate the internal producer"""
```
kbourgoin: Why an internal instance of the `Producer` instead of just subclassing?

pykafka/producer.py
emmett9001: Talked about this with @kbourgoin, and it's not actually clear that using the existing network thread reduces any complexity.

pykafka/producer.py
emmett9001: This poses an issue since we also need to use `Condition` variables here. I'm not sure how to integrate those into the handler.

pykafka/producer.py
```diff
@@ -83,17 +75,34 @@ def __init__(self,
         :param retry_backoff_ms: The amount of time (in milliseconds) to
             back off during produce request retries.
         :type retry_backoff_ms: int
-        :param required_acks: How many other brokers must have committed the
-            data to their log and acknowledged this to the leader before a
-            request is considered complete?
+        :param required_acks: The number of other brokers must have committed
```
yungchin: Probably missing a "that" before "must"

pykafka/producer.py
```diff
@@ -111,15 +125,114 @@ def __repr__(self):
...
+        If __exit__ is being called as a result of self._worker_exception
+        being raised, don't wait on the workers to finish - just re-raise
+        the exception. If __exit__ is being called because the with: block
+        has ended, do wait for workers to finish.
+        """
+        # If the thread crashed, don't wait for it
+        if self._worker_exception is not None:
+            log.exception(traceback)
+            self.stop(wait=False)
+            raise self._worker_exception
```
yungchin: I'm cheating and reading the context manager docs on the side, because I'd never written one, but those suggest this would be expected to `return False` instead of re-raising.

emmett9001: The reason I'm not doing that is that the exception that's raised when this condition is `True` is not the same exception held in `exc_value`.

yungchin: But wait - I'm probably confused. Is `self._worker_exception` always the same exception as the `exc_value` that caused the exit (that gets passed in args)?

yungchin: Ah, cross-post :)
Ok, got it!

emmett9001: `__exit__` is called when the `with` block ends or when an exception is raised. `self._worker_exception` being raised on a worker thread would not cause `__exit__` to be called. I just realized, though, that since all of the Producer's public methods raise `self._worker_exception` if it's set, `__exit__` doesn't actually need to worry about this exception at all.

pykafka/producer.py
```diff
@@ -167,7 +292,8 @@ def _get_partition_msgs(partition_id, req):
                     to_retry.extend(_get_partition_msgs(partition, req))
```
yungchin: I know these lines didn't change in this pullreq, but I think it's worth flagging up that these produce-errors (the ones that bear a `continue`) only cause a log warning, so they're sort of silently dropped.

pykafka/producer.py
```diff
@@ -111,15 +131,114 @@ def __repr__(self):
...
+        if wait:
+            self._wait_all()
+
+    def produce(self, messages):
+        """Produce a set of messages.
+
+        :param messages: The messages to produce
+        :type messages: Iterable of str or (str, str) tuples
+        """
+        def _partition_messages():
```
yungchin: Should this check `self._running`?

pykafka/producer.py
```diff
@@ -237,7 +428,167 @@ def produce(self, messages):
         :param messages: The messages to produce
         :type messages: Iterable of str or (str, str) tuples
         """
-        # Do partition distribution here. We need to be able to retry producing
-        # only *some* messages when a leader changes. Therefore, we don't want
-        # a random partition distribution changing that on the retry.
-        self._produce(self._partition_messages(messages), 0)
+        super(SynchronousProducer, self).produce(messages)
+        self._wait_all()
```
yungchin: If this will be the only difference between the two producers, couldn't we just have one producer, with a `synchronous=True` kwarg on the `produce` method?

emmett9001: Yeah, maybe that's better. It would eliminate the need to duplicate all of the kwarg documentation, and I can't think of any downsides.

pykafka/producer.py
```diff
@@ -237,7 +428,167 @@ def produce(self, messages):
...
+            if acquire:
+                self.lock.acquire()
+            if len(self.queue) >= self.producer._max_queued_messages:
+                self.slot_available.clear()
+            if acquire:
+                self.lock.release()
+            if self.producer._block_on_queue_full:
+                self.slot_available.wait()
+            else:
+                raise ProducerQueueFullError("Queue full for broker %d",
```
yungchin: When I call `produce` with a bunch of messages and this happens, is there an easy way to find out which of my messages got queued to an OwnedBroker and which didn't?

pykafka/producer.py
```diff
@@ -237,7 +428,167 @@ def produce(self, messages):
...
+        :type acquire: bool
+        """
+        assert acquire or self.lock.locked()
+        self._wait_for_slot_available(acquire=acquire)
+        if acquire:
+            self.lock.acquire()
+        self.queue.extendleft(messages)
+        self.messages_inflight += len(messages)
+        if len(self.queue) >= self.producer._max_queued_messages:
+            if not self.flush_ready.is_set():
```
yungchin: In rdkafka there's one more parameter besides `_max_queued_messages` (which is called `min.queued.messages`, which is a bit of a confusing name perhaps), which I suspect might help performance: instead of waiting until you hit max_queued, at which point the `produce` call may block while waiting for the worker thread to flush, you allow the worker thread to flush as soon as min_queued is exceeded. I wouldn't know a good name for that parameter though :)

pykafka/producer.py
emmett9001: Here's a potential solution. https://github.com/Parsely/pykafka/commit/597641040464cf5c7e327925b0c0ea6ab1eb509f To support infinite sequences, `produce` has to turn into a generator.

pykafka/producer.py
emmett9001: Sounds like you're essentially suggesting that the message count thresholds for emptying the queue and declaring that it's full. Seems reasonable to me.

pykafka/producer.py
```diff
@@ -17,17 +17,25 @@
 See the License for the specific language governing permissions and
 limitations under the License.
 """
-__all__ = ["Producer", "AsyncProducer"]
+__all__ = ["Producer", "SynchronousProducer"]
```
kbourgoin: Was going to comment about this being a breaking change, but it looks like this is how the JVM client is named too. However, I'd recommend `SyncProducer`. If Java has the short name, let's not out-length it.

pykafka/producer.py
```diff
@@ -36,26 +44,16 @@
...
-        raise NotImplementedError("AsyncProducer is unimplemented")
-
-
-class Producer():
+class Producer(object):
     """
-    This class implements the synchronous producer logic found in the
-    JVM driver.
+    This class implements asynchronous producer logic similar to that found in
+    the JVM driver.
```
kbourgoin: [nitpick] Docstring first lines should be short, one-line, and describe the class. It make code folding and doc generation cleaner. It should also be on the same line as the triple quote.

```
"""Asynchronous Producer Implementation.

Blah blah blah.
"""
```

pykafka/producer.py
```diff
@@ -102,7 +124,32 @@ def __init__(self,
...
+        self._linger_ms = linger_ms
+        self._block_on_queue_full = block_on_queue_full
+        self._worker_exception = None
+        self._worker_trace_logged = False
+        self._running = False
+        self.start()
+
+    def raise_worker_exceptions(fn):
+        """Decorator that raises exceptions encountered on worker threads
+        """
```
kbourgoin: [nitpick] If the docstring fits on one line, the closing triple-quote should be on that line too.

pykafka/producer.py
```diff
@@ -111,15 +158,112 @@ def __repr__(self):
             id_=hex(id(self))
         )
 
-    def _send_request(self, broker, req, attempt):
+    def __enter__(self):
+        """Context manager entry point - start the producer"""
+        self.start()
+        return self
```
kbourgoin: You're calling `self.start()` twice. In `__init__` and here. I'd leave it in `__init__` and remove this one.

pykafka/producer.py
```diff
@@ -111,15 +158,112 @@ def __repr__(self):
...
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Context manager exit point - stop the producer
+
+        If __exit__ is being called as a result of self._worker_exception
+        being raised, don't wait on the workers to finish. If __exit__ is being
+        called because the with: block has ended, do wait for workers to finish.
+        """
+        # If the thread crashed, don't wait for it
+        self.stop(wait=self._worker_exception is None)
```
kbourgoin: If there's an exception here, `wait` doesn't matter because it'll get raised before that matters.

pykafka/producer.py
```diff
@@ -111,15 +158,112 @@ def __repr__(self):
...
+
+    @raise_worker_exceptions
```
kbourgoin: How is it possible there have been exceptions before `start` is called? It's called at the end of `__init__` before the workers have started.

pykafka/producer.py
```diff
@@ -111,15 +158,112 @@ def __repr__(self):
...
+        """Set up data structures and start worker threads"""
+        if not self._running:
+            self._owned_brokers = {}
+            for partition in self._topic.partitions.values():
+                if partition.leader.id not in self._owned_brokers:
+                    self._owned_brokers[partition.leader.id] = OwnedBroker(
+                        self, partition.leader)
+            self._running = True
+
+    @raise_worker_exceptions
```
kbourgoin: I'm not sure this should raise exceptions, actually. If it raises because one worker is dead, how will the others be notified they should stop?

pykafka/producer.py
```diff
@@ -111,15 +158,112 @@ def __repr__(self):
...
+                    key = None
+                    value = message
+                else:
+                    key, value = message
+                yield (key, str(value)), self._partitioner(partitions, message).id
+        if not self._running:
+            raise ProducerStoppedException()
+        pairs = itertools.izip(self._produce(_partition_messages()), messages)
+        for message_enqueued in pairs:
+            yield message_enqueued
```
kbourgoin: Why is produce giving you the messages right back? This seems unnecessary.

kbourgoin: And potentially confusing as to what a user should do with this.

pykafka/producer.py
```diff
@@ -111,15 +158,112 @@ def __repr__(self):
...
+            request.add_message(
+                Message(value, partition_key=key),
+                self._topic.name,
+                partition_id
+            )
+        log.debug("Sending %d messages to broker %d",
+                  len(message_batch), owned_broker.broker.id)
+        self._send_request(request, 0, owned_broker)
+
+    @raise_worker_exceptions
```
kbourgoin: Doesn't this run on a background thread?

pykafka/producer.py
```diff
@@ -102,7 +124,32 @@ def __init__(self,
...
+        self._max_queued_messages = max_queued_messages
+        self._min_queued_messages = min_queued_messages
+        self._linger_ms = linger_ms
+        self._block_on_queue_full = block_on_queue_full
+        self._worker_exception = None
+        self._worker_trace_logged = False
+        self._running = False
+        self.start()
+
+    def raise_worker_exceptions(fn):
```
kbourgoin: So, I'm actually not a fan of a decorator here.  This is a check which needs to happen, but it also looks like something which was sprinkled around without a ton of consideration as to when we're raising exceptions and how. It's called in places where it's detrimental (`stop`) and called places where it'll be ineffective (`_prepare_request` and `_send_request`). For the former case, it should be called after `self._running = False`, but you can't do that because it's a decorator.  In that latter case, it'll crash a background thread by raising an exception which won't be raised to the user and might kill an otherwise healthy thread.

This is why I don't like to decorator approach. It encourages adding it all over the place "just to be safe" because decorators look safe. Maybe it's just how I perceive it, but if this were a function call at the top of user-facing functions (i.e. a Guard), it'd be less likely to be added as it has been.

Maybe that's just me, though. Keep the decorator if you think it's useful, but right now it's used in way too many places. It should be used to guard user-facing functions, and in cases where it's not going to hurt things which otherwise should be happening (like shutting down the producer).

pykafka/producer.py
```diff
@@ -180,56 +331,128 @@ def _get_partition_msgs(partition_id, req):
             attempt += 1
             if attempt < self._max_retries:
                 time.sleep(self._retry_backoff_ms / 1000)
-                self._produce(to_retry, attempt)
+                # we have to call _produce here since the broker/partition
+                # target for each message may have changed
+                for _ in self._produce(to_retry, attempt):
+                    pass
```
kbourgoin: This is definitely a point against having `_produce` return an iterator.

pykafka/producer.py
kbourgoin: With everything being iterables here, what happens if someone calls `produce` and never iterates over the result?

pykafka/producer.py
```diff
@@ -180,56 +331,128 @@ def _get_partition_msgs(partition_id, req):
...
-            else:
-                key, value = message
-            value = str(value)
-            yield (key, value), self._partitioner(partitions, message).id
+        # empty queues and figure out updated partition leaders
+        new_queue_contents = defaultdict(list)
+        for owned_broker in self._owned_brokers.itervalues():
+            owned_broker.lock.acquire()
+            current_queue_contents = owned_broker.flush(0, acquire=False,
+                                                        release_inflight=True)
```
kbourgoin: Use a re-entrant lock here. Having to tell the queue whether or not you want it to acquire the lock is poor separation of concerns. If you use the re-entrant lock, it'll be able to still acquire the lock since it's on the same thread.

pykafka/producer.py
```diff
@@ -180,56 +331,128 @@ def _get_partition_msgs(partition_id, req):
...
-                key = None
-                value = message
-            else:
-                key, value = message
-            value = str(value)
-            yield (key, value), self._partitioner(partitions, message).id
+        # empty queues and figure out updated partition leaders
+        new_queue_contents = defaultdict(list)
+        for owned_broker in self._owned_brokers.itervalues():
+            owned_broker.lock.acquire()
```
kbourgoin: A better world is one where this function doesn't even know that the broker has locks. Is that possible here? I'm not actually sure.

pykafka/producer.py
```diff
@@ -180,56 +331,128 @@ def _get_partition_msgs(partition_id, req):
...
+                                 self._block_on_queue_full,
+                                 acquire=False)
+            owned_broker.resolve_event_state()
+            owned_broker.lock.release()
 
-    def _produce(self, message_partition_tups, attempt):
-        """Publish a set of messages to relevant brokers.
+    @raise_worker_exceptions
+    def _wait_all(self):
+        """Block until all messages in flight are sent
```
kbourgoin: Just a thought, but "in flight" is an awkward term to use for these messages. When I see "in flight," I think "on the network on their way to the broker."  To me, this is more like "pending" messages since they're waiting to be sent. I can see how it cuts either way, so just a thought.

pykafka/producer.py
kbourgoin: IIRC it's how the JVM producer is implemented. I have no strong feelings either way, and don't see any downsides to combining them.

pykafka/producer.py
emmett9001: I've actually considered all of this, and have waffled back and forth between having it be a decorator and a simple two lines of code at the end of every function it's needed in. I liked the decorator approach better than calling a member function at the end of every function. I also considered only putting it on user-facing functions, but I chose to put it everywhere because of the possibility that users will call private functions. I couldn't find a case where having these checks on functions called on worker threads actually causes a problem - the thread will not have a chance to raise the exception a second time from the decorator, since it will already have exited after the original exception (from the `break` in its main loop error handler). I've also not been able to replicate a case where having `exit()` decorated causes a problem, but that's the one I'm least confident about.

Since a lot of what you've said here aligns with my gut instinct about this, I believe I'll keep the decorator on public methods only, unless I've inadvertently convinced you otherwise here.

pykafka/producer.py
emmett9001: "pending" is much better, I don't think it's close.

pykafka/producer.py
emmett9001: Thanks, this sounds like the "correct" way to do what I'm trying here.

pykafka/producer.py
kbourgoin: The decorator's fine, but definitely only on the public functions. If someone is using private functions, that's their own problem.

As for background threads, this is using `self._worker_exception`, so as soon as that's set from any thread, doesn't that cause every other background thread to raise the exception once they call one a decorated function?

pykafka/producer.py
emmett9001: Yeah, this is not great. Could use some wrapper function to make this code cleaner, and to automatically iterate the generator for users who want to ignore the yielded values.

pykafka/producer.py
emmett9001: Yes, and I've come to believe that my reasoning for doing that was flawed. I'll just keep this decorator on public methods.

pykafka/producer.py
emmett9001: After discussion, we agree that it makes more sense to stop supporting sequences of messages in `produce`. That will make the whole `block_on_queue_full` issue go away. It will also require that we use a major version bump.

pykafka/producer.py
emmett9001: Agree, this will be removed.

pykafka/producer.py
emmett9001: These exceptions would have happened during `start`, not before it, since `raise_worker_exceptions` does its work after running the decorated function.

pykafka/producer.py
emmett9001: What do you think about completely removing the `SyncProducer` class and having synchronous behavior switched on and off by a kwarg on `Producer`?

kbourgoin: Yeah, I think that's a great idea. If we're in the real of making breaking changes, that'd make things overall a lot cleaner.

pykafka/producer.py
yungchin: My guess would be that's not possible here, because you need to hold locks across all brokers, to prevent new messages getting slipped-in between the old ones while they are being redistributed.

pykafka/producer.py
emmett9001: I don't follow this logic. Where is the exception guaranteed to be raised before the context manager exits? The idea here is to not hang infinitely waiting on the messages to send when the worker thread has crashed.

kbourgoin: You know, this is what happens when I don't read the decorator closely.  :)

I'm thought it did stuff before the function call, not after, so that wasn't necessarily true. Of course, the decorator is gone now, so the point is moot.

emmett9001: I got the impression you thought that, but wasn't certain. I'm not actually sure that there's a case where this setting will matter, but it's a safeguard in case there is one.

pykafka/producer.py
emmett9001: Yeah I think the nature of the problem means that this function has to lock them all at once.

pykafka/producer.py
```diff
@@ -111,15 +157,97 @@ def __repr__(self):
...
+        :param message: The message to produce
+        :type message: str
+        :param partition_key: The key to use when deciding which partition to send this
+            message to
+        :type partition_key: str
+        """
+        if not self._running:
+            raise ProducerStoppedException()
+        partitions = self._topic.partitions.values()
+        partition_id = self._partitioner(partitions, (message, partition_key)).id
```
yungchin: This bug reported in #232 affects this line too.

pykafka/producer.py
```diff
@@ -111,15 +157,97 @@ def __repr__(self):
...
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Context manager exit point - stop the producer
+
+        If __exit__ is being called as a result of self._worker_exception
+        being raised, don't wait on the workers to finish. If __exit__ is being
+        called because the with: block has ended, do wait for workers to finish.
+        """
+        # If the thread crashed, don't wait for it
```
yungchin: This comment and part of the docstring are leftovers from a previous iteration.

yungchin: Something else: do we also need a `__del__` that calls `stop` too?

PR #183, opened 2015-06-12
[RDY] tests: cherry-picked changes from feature/rdkafka-extension

yungchin:
These are the accumulated changes to test_simpleconsumer on the rdkafka
branch at de4b5453910fe903ec51a7d29c13a65445f8ce7f.  There are basically
two changes:

 * The instantiation of a consumer is spun out into a separate method,
   so that this may be overridden (that is, to insert an rdkafka consumer
   instead)
 * The assertions that required that offsets obtained were always on a
   clean cluster, and thus zero-based, have been rewritten (but
   preserving the aim of each test).  This was needed because the tests
   would erratically pass then fail then pass, as the offsets obtained
   through the Offset API would sometimes be stale offsets from a
   previous run or a previous TestCase (I observed both)

### DIFF
--- a/tests/pykafka/test_simpleconsumer.py
+++ b/tests/pykafka/test_simpleconsumer.py
@@ -49,9 +49,7 @@ class TestSimpleConsumer(unittest2.TestCase):
 
             offsets_fetched = {r[0]: r[1].offset
                                for r in consumer.fetch_offsets()}
-            offset_diff = sum(offsets_fetched[i] - offsets_committed[i]
-                              for i in offsets_committed)
-            self.assertEquals(offset_diff, 0)
+            self.assertEquals(offsets_fetched, offsets_committed)
 
     def test_offset_resume(self):
         """Check resumed internal state matches committed offsets"""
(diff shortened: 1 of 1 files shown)


yungchin:
@kbourgoin @emmett9001 hoping these can go in to avoid diverging from master too much.

kbourgoin:
@yungchin, I'm fine merging this in, but what's the success rate been on Travis? Does this help or hurt the rate at which tests are freezing, causing the build to error out?

Also, as far as keeping up to date with master goes, I'd recommend `git merge master` frequently. It'll help you stay up to date so that when we're ready to merge the rdkafka changes in, it'll be as seamless as possible.

yungchin:
Thanks @kbourgoin. I haven't seen any difference in test freeze rates, and
fwiw the two times that Travis ran against this pullreq both passed. I
suspect the higher frequency of test freezes may be particular to my
somewhat resource-poor local setup (still I should probably put some of the
unhandled ConnectionLoss exceptions on the issue tracker when I catch them
again).

I'll make sure to keep merging from master often. The reason I thought I'd
ask to "fast track" these particular bits is that they touch so many lines
in that file that resolving conflicts later is bound to be ugly.

PR #192, opened 2015-06-26
Feature/kafka tools

kbourgoin:
This PR adds the `kafka-tools` executable as part of the PyKafka install. The tool provides some basic tools for working with a kafka cluster including the ability to reset consumer offsets and inspect various information about topics, partitions, and consumers.

### DIFF
--- a/pykafka/cli/kafka_tools.py
+++ b/pykafka/cli/kafka_tools.py
@@ -1,0 +1,218 @@
+import argparse
+import calendar
+import datetime as dt
+
+import tabulate
+
+import pykafka
+from pykafka.common import OffsetType
+
+
+def fetch_offsets(client, topic, offset):
+    """Fetch raw offset data from a topic.
+
+    :param client: KafkaClient connected to the cluster.
+    :type client:  :class:`pykafka.KafkaClient`
+    :param topic:  Name of the topic.
+    :type topic:  :class:`pykafka.topic.Topic`
+    :param offset: Offset to reset to. Can be earliest, latest or a datetime.
+        Using a datetime will reset the offset to the latest message published
+        *before* the datetime.
+    :type offset: :class:`pykafka.common.OffsetType` or
+        :class:`datetime.datetime`
+    :returns: dict of {partition_id: :class:`pykafka.protocol.OffsetPartitionResponse`}
+    """
+    if offset.lower() == 'earliest':
+        return topic.earliest_available_offsets()
+    elif offset.lower() == 'latest':
+        return topic.latest_available_offsets()
+    else:
+        offset = dt.datetime.strptime(offset, "%Y-%m-%dT%H:%M:%S")
+        offset = int(calendar.timegm(offset.utctimetuple())*1000)
+        return topic.fetch_offset_limits(offset)
+
+
+def fetch_consumer_lag(client, topic, consumer_group):
+    """Get raw lag data for a topic/consumer group.
+
+    :param client: KafkaClient connected to the cluster.
+    :type client:  :class:`pykafka.KafkaClient`
+    :param topic:  Name of the topic.
+    :type topic:  :class:`pykafka.topic.Topic`
+    :param consumer_group: Name of the consumer group to reset offsets for.
+    :type consumer_groups: :class:`str`
+    :returns: dict of {partition_id: (latest_offset, consumer_offset)}
+    """
+    latest_offsets = fetch_offsets(client, topic, 'latest')
+    consumer = topic.get_simple_consumer(consumer_group=consumer_group,
+                                         auto_start=False)
+    current_offsets = consumer.fetch_offsets()
+    return {p_id: (latest_offsets[p_id].offset[0], res.offset)
+            for p_id, res in current_offsets}
+
+
+def print_consumer_lag(client, topic, consumer_group):
+    """Print lag for a topic/consumer group.
+
+    :param client: KafkaClient connected to the cluster.
+    :type client:  :class:`pykafka.KafkaClient`
+    :param topic:  Name of the topic.
+    :type topic:  :class:`str`
+    :param consumer_group: Name of the consumer group to reset offsets for.
+    :type consumer_groups: :class:`str`
+    """
+    # Don't auto-create topics.
+    if topic not in client.topics:
+        raise ValueError('Topic {} does not exist.'.format(topic))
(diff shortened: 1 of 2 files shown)


pykafka/cli/kafka_tools.py
```diff
@@ -0,0 +1,285 @@
...
+    :returns: dict of {partition_id: :class:`pykafka.protocol.OffsetPartitionResponse`}
+    """
+    if offset.lower() == 'earliest':
+        return topic.earliest_available_offsets()
+    elif offset.lower() == 'latest':
+        return topic.latest_available_offsets()
+    else:
+        offset = dt.datetime.strptime(offset, "%Y-%m-%dT%H:%M:%S")
+        offset = int(calendar.timegm(offset.utctimetuple())*1000)
+        return topic.fetch_offset_limits(offset)
```
emmett9001: Does this actually work? I wasn't aware you could give this function a time.

pykafka/cli/kafka_tools.py
```diff
@@ -0,0 +1,285 @@
...
+                             '[default: localhost:9092]')
+
+    subparsers = output.add_subparsers(help='Commands', dest='command')
+
+    # Desc Topic
+    parser = subparsers.add_parser(
+        'desc_topic',
+        help='Print detailed info for a topic.'
+    )
+    parser.add_argument('topic',
```
dan-blanchard: If you're going to add the same argument for multiple subparsers, it's a bit cleaner to make functions like:

```python
def add_topic(subparsers):
    parser.add_argument('topic', help='Topic name.')
```

and then do:
```python
add_topic(parser)
```

pykafka/cli/kafka_tools.py
```diff
@@ -0,0 +1,285 @@
...
+    current_offsets = consumer.fetch_offsets()
+    return {p_id: (latest_offsets[p_id].offset[0], res.offset)
+            for p_id, res in current_offsets}
+
+
+#
+# Commands
+#
+
+def desc_topic(client, topic):
```
emmett9001: This feels like something that should be included in `Topic` itself.

pykafka/cli/kafka_tools.py
```diff
@@ -0,0 +1,285 @@
...
+    # Print Consumer Lag
+    parser = subparsers.add_parser(
+        'print_consumer_lag',
+        help='Get consumer lag for a topic.'
+    )
+    parser.add_argument('topic',
+                        metavar='TOPIC',
+                        help='Topic name.')
+    parser.add_argument('consumer_group',
+                        metavar='CONUSMER_GROUP',
```
dan-blanchard: should be `CONSUMER_GROUP`

pykafka/cli/kafka_tools.py
kbourgoin: Yeah, it's crazy. It's based on segment boundaries in the data files.  The very end of [this post](http://cfchou.github.io/blog/2015/04/23/a-closer-look-at-kafka-offsetrequest/) sort of explains it.

pykafka/cli/kafka_tools.py
kbourgoin: It's really an output function, though. It's just formatting data that's already there for human consumption, hence it being in this tool.

pykafka/cli/kafka_tools.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+def main():
+    parser = _get_arg_parser()
+    args = parser.parse_args()
+
+    # Connect to Kafka
+    client = pykafka.KafkaClient(hosts=args.host)
+
+    if args.command == 'reset_offsets':
```
dan-blanchard: Instead of doing this to make the correct function get called, you can just do:

```python
parser.set_defaults(func=reset_offsets)
```

when you're creating the subparser, and make each of these function take `args` instead of individual arguments.

Then this whole if block can just become:

```python
args.func(args)
```

pykafka/cli/kafka_tools.py
```diff
@@ -0,0 +1,285 @@
...
+    # Don't auto-create topics.
+    if topic not in client.topics:
+        raise ValueError('Topic {} does not exist.'.format(topic))
+    topic = client.topics[topic]
+
+    offsets = fetch_offsets(client, topic, offset)
+    consumer = topic.get_simple_consumer(consumer_group=consumer_group,
+                                         auto_start=False)
+    for owned_partition, partition in consumer._partitions.iteritems():
+        owned_partition.last_offset_consumed = int(offsets[partition.id].offset[0])
```
emmett9001: If this works, awesome - it was not really my intention to allow this, though. `_partitions` is supposed to be private, and it's a coincidence that modifying `last_offset_consumed` like this happens to do what you want it to. We should probably make this operation rely on public, documented features.

pykafka/cli/kafka_tools.py
kbourgoin: @dan-blanchard what do you think? cast the deciding vote, mon ami.

pykafka/cli/kafka_tools.py
kbourgoin: Ugh. While I realize you're right about this, it's also just a pain in the ass to do all the work around that. :)

emmett9001:
Looks great overall. Adding to this file over time will be really helpful. A function for measuring throughput based on periodic sampling of the offset sum could be useful here as well, though it's not super important.

pykafka/cli/kafka_tools.py
dan-blanchard: If this is something you expect people to use programmatically, put it in `Topic`.  If not, I think having it in the script is fine.

dan-blanchard:
:+1:

emmett9001:
:tada:

pykafka/cli/kafka_tools.py
emmett9001: @kbourgoin I realized you should at least be using `set_offset` here instead of directly modifying `last_offset_consumed`. That way, you're at least safer if the underlying `OwnedPartition` implementation changes.

kbourgoin: Not even using them anymore. I can create the offset commit requests and push those to the broker easily enough. I was ultimately only using the simple consumer because I wanted to use it's retry logic.  For a tool like this though, it's not truly necessary to have all that.

Issue #187, opened 2015-06-21
Manually Set Offsets in Consumers


zhynxn:
Hello, 
I have question about set offset for consumer-group on partition. 
I know there is auto_offset_reset parameter to reset the offset to LATEST or EARLIEST when OffsetOutOfRangeError happens. It is really helpful. 

My question is can I manually set specific offset number for consumer-group, partition? So I can do more error handling in my own program. 

There is such function in kafka-python package (http://kafka-python.readthedocs.org/en/latest/apidoc/kafka.consumer.html) 

Please let me know. 
Thanks,
Yanxin 

set_topic_partitions(*topics)
Set the topic/partitions to consume Optionally specify offsets to start from

Accepts types:

str (utf-8): topic name (will consume all available partitions)
tuple: (topic, partition)
dict:
{ topic: partition }
{ topic: [partition list] }
{ topic: (partition tuple,) }
Optionally, offsets can be specified directly:

tuple: (topic, partition, offset)
dict: { (topic, partition): offset, ... }

yungchin:
Hi Yanxin,

I'm currently using something like `consumer._partitions_by_id[partition_id].set_offset(offset)` to do this, but I suppose that feels like using what is a semi-public interface at best. It would probably be good to get @emmett9001's advice on this too.

emmett9001:
The `SimpleConsumer` currently doesn't directly support starting consumption from an arbitrary offset. You can only start from `OFFSET_EARLIEST` or `OFFSET_LATEST`. @yungchin is correct that it's possible to adjust the current offset of the consumer with `consumer._partitions_by_id[partition_id].set_offset(offset)`, but this method has two problems. First, you have to make this call after instantiating the consumer or your offset changes will be overwritten. Second, it's a private method that isn't really meant to be used externally.

This is a feature that's worth adding, so I'll add this to the next release milestone.

yungchin:
Can I add two related feature requests here for discussion?

I suppose a big use case for this feature is that people store their offsets outside of kafka. For that use case, in addition to manually setting, it would be nice to have a simple interface for getting offsets out of the consumer (this is different from `fetch_offsets` which gets them from kafka). I think this code in tests might be good enough for that? https://github.com/Parsely/pykafka/blob/24c364527012fe0768e2c6addc19be9990f3c1f6/tests/pykafka/test_simpleconsumer.py#L67

[...]

PR #194, opened 2015-06-26
public interface for offset resetting

emmett9001:
This pull request adds a public interface to the consumers that allows client code to set the current partition offsets to whatever they choose. Previously, this had to be done via a semi-public undocumented interface as detailed in [this issue](https://github.com/Parsely/pykafka/issues/187).

Given these changes, the new workflow for specifying offsets from which to consume looks like this:
```python
consumer = topic.get_simple_consumer(consumer_group="testgroup")
partition_offset_pairs = [(p, get_offset_for_partition(p)) for p in consumer.partitions.itervalues()]
consumer.reset_offsets(partition_offsets=partition_offset_pairs)
```
This pull request also ties `reset_offsets` to the `fetch` locking mechanism so that a partition can never be fetching and resetting its offsets at the same time. This makes the unlocking logic at the end of `fetch()` notably more complicated, since one of its error conditions involves calling `reset_offsets`.

**The following has been resolved in this pull request**
~~The only issue I see with this method is that the consumer has a chance to start fetching messages between `get_simple_consumer` and `reset_offsets`. To get around this,  we could accept the `partition_offset_pairs` in `get_simple_consumer`, but that would require some way to obtain a list of partitions before actually instantiating a consumer. This sounds like a road I'd rather not go down, so the question is: how bad is it that it's impossible to start the consumer at an arbitrary offset that's not the head or tail of the partition? My gut tells me it doesn't matter and this solution is ok.~~
Let me know what you think, @kbourgoin and @yungchin.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -30,7 +30,7 @@ from .utils.compat import Semaphore
 from .exceptions import (OffsetOutOfRangeError, UnknownTopicOrPartition,
                          OffsetMetadataTooLarge, OffsetsLoadInProgress,
                          NotCoordinatorForConsumer, SocketDisconnectedError,
-                         ConsumerStoppedException, ERROR_CODES)
+                         ConsumerStoppedException, KafkaError, ERROR_CODES)
 from .protocol import (PartitionFetchRequest, PartitionOffsetCommitRequest,
                        PartitionOffsetFetchRequest, PartitionOffsetRequest)
 from .utils.error_handlers import (handle_partition_responses, raise_error,
@@ -482,12 +482,11 @@ class SimpleConsumer():
             owned_partition_offsets = dict((self._partitions[p], offset)
                                            for p, offset in partition_offsets)
         except KeyError as e:
-            raise("Unknown partition supplied to reset_offsets\n%s", e)
+            raise KafkaError("Unknown partition supplied to reset_offsets\n%s", e)
 
         log.info("Resetting offsets for %s partitions", len(list(owned_partition_offsets)))
 
         for i in xrange(self._offsets_reset_max_retries):
-            log.debug("Retrying offset reset")
             # group partitions by leader
             by_leader = defaultdict(list)
             for partition, offset in owned_partition_offsets.iteritems():
@@ -534,6 +533,7 @@ class SimpleConsumer():
 
             if len(parts_by_error) == 1 and 0 in parts_by_error:
                 break
+            log.debug("Retrying offset reset")
 
         if self._consumer_group is not None:
             self.commit_offsets()
@@ -587,8 +587,8 @@ class SimpleConsumer():
                 # release the lock in these cases, since resolving the error
                 # requires an offset reset and not releasing the lock would
                 # lead to a deadlock in reset_offsets. For successful requests
-                # or requests with # different errors, we still assume # that
-                # it's ok to # retain the lock since no offset_reset can happen
+                # or requests with different errors, we still assume that
+                # it's ok to retain the lock since no offset_reset can happen
                 # before this function returns
                 out_of_range = parts_by_error.get(OffsetOutOfRangeError.ERROR_CODE, [])
                 for owned_partition, res in out_of_range:
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -520,10 +520,10 @@ class SimpleConsumer():
 
                 if 0 in parts_by_error:
                     parts_by_error.pop(0)
-                partitions = []
-                partitions.extend(
-                    [part for errcode, parts in parts_by_error.iteritems()
-                     for part in parts])
+                owned_partition_offsets = dict(
+                    (part, owned_partition_offsets[part])
+                    for errcode, parts in parts_by_error.iteritems()
+                    for part, _ in parts)
 
             # release all locks to allow fetching
             for errcode, owned_partitions in parts_by_error.iteritems():
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -194,8 +194,6 @@ class SimpleConsumer():
         # Figure out which offset wer're starting on
         if self._reset_offset_on_start:
             self.reset_offsets()
-            # make sure the reset is saved in kafka before it can rebalance
-            self.commit_offsets()
         elif self._consumer_group is not None:
             self.fetch_offsets()
 
@@ -484,6 +482,8 @@ class SimpleConsumer():
 
             if len(parts_by_error) == 1 and 0 in parts_by_error:
                 break
+
+        self.commit_offsets()
 
     def fetch(self):
         """Fetch new messages for all partitions
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -30,7 +30,7 @@ from .utils.compat import Semaphore
 from .exceptions import (OffsetOutOfRangeError, UnknownTopicOrPartition,
                          OffsetMetadataTooLarge, OffsetsLoadInProgress,
                          NotCoordinatorForConsumer, SocketDisconnectedError,
-                         ConsumerStoppedException, ERROR_CODES)
+                         ConsumerStoppedException, KafkaException, ERROR_CODES)
 from .protocol import (PartitionFetchRequest, PartitionOffsetCommitRequest,
                        PartitionOffsetFetchRequest, PartitionOffsetRequest)
 from .utils.error_handlers import handle_partition_responses, raise_error
@@ -429,8 +429,10 @@ class SimpleConsumer():
                     # so account for this here by passing offset - 1
                     owned_partition.set_offset(new_offset - 1)
                 else:
-                    log.warning("Offset reset for partition {} failed.".format(
-                                owned_partition.partition.id))
+                    msg = "Offset reset for partition {} failed.".format(
+                        owned_partition.partition.id)
+                    log.warning(msg)
+                    raise KafkaException(msg)
                 # release locks on succeeded partitions to allow fetching
                 # to resume
                 owned_partition.fetch_lock.release()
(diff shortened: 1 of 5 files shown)


yungchin:
Cool cool, this is a feature I've good use for.

> The only issue I see with this method is that the consumer has a chance to start fetching messages between get_simple_consumer and reset_offsets.

Ah, good point - I hadn't realised that when I posted on #187. So iiuc, that means if you start the consumer, then change the offsets, you might get a message from `consume()` at a different offset, because that got on the queue inbetween? That could be confusing. Would it be possible to flush the queues on all partitions passed to `reset_offsets()`?

> we could accept the partition_offset_pairs in get_simple_consumer, but that would require some way to obtain a list of partitions before actually instantiating a consumer

That wouldn't be completely terrible either. Presumably, someone using this feature has some other place where they store offsets (I've an app where the offsets sit in postgres), and so they would usually know which partitions they want to begin with.

Although... it might get messy when the number of partitions is changed (by a cluster operator). If you then have to provide a list of partitions+offsets on init, you then can't make it grab the new partitions. So I guess for that scenario your current solution is better: you can grab whatever partitions there are (ie don't specify a list on init), and then for the ones you already knew about you can reset the offsets.

Note to self: I think I have to handle `reset_offsets()` in #176, if it's going to be a public method. Otherwise, the rdkafka consumer would completely ignore changed offsets after `start()`.

emmett9001:
I'm not totally convinced that there's a good reason to allow setting offsets on `__init__`, but I do think it's a smart idea to have `reset_offsets` always flush the internal queues when called by a user. When called internally in response to an error, it shouldn't, but this can be handled with a kwarg.

yungchin:
> I'm not totally convinced that there's a good reason to allow setting offsets on `__init__`

Yes, no indeed - after re-reading my earlier ramble, I agree :)

emmett9001:
I've added the option to lock and flush the partition queues during an offset reset. This means that users no longer need to manually pause consumption before calling `reset_offsets`. In implementing this, I also discovered a bug: if `consume` is called when there are some messages in the queue and after `reset_offsets` (`_reset_offsets` in current master) has been called, the partition's internal offset counter (`last_offset_consumed`) will be reset to match the message just returned from `consume`. That means that the offset set by `reset_offsets` will be effectively overwritten and the offset will appear not to have been reset. It's quite possible that this is the bug we've been seeing in production in which not all offsets are reset when deploying with `reset_offsets_on_start`.
To fix this, we now lock all fetcher threads and clear the per-partition message queues before resetting offsets so that any subsequent calls to `consume`  will start from the offset intended by `reset_offsets`, not from the offset of some other message hanging around in the queue.

emmett9001:
This is ready for review @yungchin @kbourgoin. I'll be testing it against real streams as soon as possible, too.

pykafka/simpleconsumer.py
```diff
@@ -408,37 +410,50 @@ def _handle_success(parts):
...
-        returned offset in the OwnedPartition per self._auto_offset_reset
+        returned offset in the OwnedPartition
 
-        :param partitions: the partitions for which to reset offsets
+        :param partitions: (`partition`, `offset`) pairs to reset
+            where `partition` is the partition for which to reset the offset
+            and `offset` is the new offset the partition should have
         :type partitions: Iterable of
-            :class:`pykafka.simpleconsumer.OwnedPartition`
+            (:class:`pykafka.simpleconsumer.OwnedPartition`, int)
```
yungchin: Could this become "Iterable of (int, int)"  - so that is, (partition_id, offset) - instead? Otherwise a user would have to go through `self._partitions_by_id` to obtain instances of `OwnedPartition` first before they'd call this, I think.

pykafka/simpleconsumer.py
```diff
@@ -462,6 +477,15 @@ def _handle_success(parts):
                     [part for errcode, parts in parts_by_error.iteritems()
                      for part in parts])
 
+            # release all locks to allow fetching
+            for errcode, owned_partitions in parts_by_error.iteritems():
```
yungchin: So if we get here, and we're at our `self._offsets_reset_max_retries`th retry, I think this may leave the errored partitions at some other offset than specified - could we raise an exception at that point?

pykafka/simpleconsumer.py
```diff
@@ -194,7 +194,7 @@ def start(self):
 
         # Figure out which offset wer're starting on
         if self._reset_offset_on_start:
-            self._reset_offsets()
+            self.reset_offsets()
             # make sure the reset is saved in kafka before it can rebalance
```
yungchin: Here in `start()`, this is followed immediately by `commit_offsets`. Would it make sense to append that at the end of `reset_offsets` (so something like `if self._consumer_group is not None: commit_offsets()`) instead?

One scenario where that makes a difference is if the user calls `reset_offsets` and only then `start`s the consumer: in that case, if we didn't commit the offsets yet, `start` would overwrite them with the output of `fetch_offsets`, I think?

pykafka/simpleconsumer.py
emmett9001: They can actually use the `partitions` property that's currently in master. I think it's cleaner and more consistent with the rest of the API to supply `Partition` instances over their ids.

pykafka/simpleconsumer.py
emmett9001: Not sure, you mean the errored partitions will still be at the offset they were at before the call to `reset_offsets`? That's also true of current master. The thing about raising an exception is that it means the client has to handle it. Since `reset_offsets` is actually called in response to another error condition, this could get messy.

pykafka/simpleconsumer.py
emmett9001: From your first sentence, it sounds more like you mean something like this:
```python
if self._reset_offset_on_start:
    self.reset_offsets()
def reset_offsets(self, ...):
    ...
    self.commit_offsets()
```
This makes a lot of sense to me, but it doesn't match the code you provided. I believe it solves the error case you mention. Am I missing something?

yungchin: Ah yes, that's indeed what I meant - but I think that last line needs to be wrapped in `if self._consumer_group is not None`, because otherwise `commit_offsets` might raise an exception?

pykafka/simpleconsumer.py
yungchin: I see what you mean, when `reset_offsets` is called internally by the error handler that's a bit messy. I was mostly thinking of use of `reset_offsets` by the user application - if we don't raise an exception, does that mean the user must always check offsets on messages to see if any errors occured?

emmett9001: That is a good point. I think this is a big enough issue to warrant its own ticket.

pykafka/simpleconsumer.py
yungchin: I see what you mean that that feels inconsistent, but on the other hand as a lazy user of the interface I'd say that the nice thing about ids is that you can save them to disk, and so when you get them from disk, you could pass them straight in here.

Maybe I'm missing a trick though, and there's a shortcut somehow. Am I right in thinking you'd have to get the ids, then map them to `partition.Partition`s (which is what you get from the `BalancedConsumer.partitions` property) and then again map to `simpleconsumer.OwnedPartition`s to call `reset_offsets`?

emmett9001: Oh, sure - if you're saving partition IDs to disk then you would have to do a bit of work to construct the input to `reset_offsets`. I think a cleaner solution is to stop this function from accepting `OwnedPartition` and have it take `Partition` instead, since `OwnedPartition` is supposed to be private. I'll hold off on merging until I've figured that out, then.

yungchin:
Ok, so basically I've only lazy complaints about the function signature :)
The lack of exceptions I agree is not a regression from current master, so it makes sense to save them for a separate ticket, and as for how it looks under the hood otherwise: no complaints at all.

emmett9001:
Awesome, thanks for the feedback. Your comments did make me realize it's no good to expose `OwnedPartition` to the client, so I need to figure that out.

emmett9001:
It turns out that kafka-python allows this by [promising to fetch the given offset next](https://github.com/mumrah/kafka-python/blob/master/kafka/consumer/kafka.py#L206), not by actually committing the given offset to kafka. This is a problem because it creates a mismatch between the offset stored in kafka and the offset stored in the consumer; this is contrary to the design we've tried to maintain in pykafka so far.
Given that we'd like to commit offsets as soon as possible to kafka, we could try simply sending the user-supplied offsets to kafka during `reset_offsets` in an `OffsetRequest`. Unfortunately, the `OffsetRequest` only accepts timestamps, not actual offset values. When supplied with a timestamp it doesn't like, the request returns with an error code of `0` and an empty list for the offset. This seems to be a bug or at least a poor design choice in kafka.
It occurs to me that we could allow the user to supply either timestamps or actual offset values to `reset_offsets`. If the request succeeds and returns an offset (the dual error condition that we apparently have to check), we're all good. If it returns the empty offset list, we can simply set the consumer's internal offset to the supplied value and continue on. This will mirror the behavior of kafka-python and provide a bit more versatility.

emmett9001:
This is now ready for review from @yungchin and @kbourgoin. @yungchin, this hasn't changed a ton since you looked at it, but I fixed some deadlocks and improved the interface for user-supplied offsets.

pykafka/simpleconsumer.py
```diff
@@ -461,6 +525,19 @@ def _handle_success(parts):
                     [part for errcode, parts in parts_by_error.iteritems()
```
yungchin: Given that `partitions` has been dropped from the code above this, I think L523-525 may need rewriting to write into `owned_partition_offsets` instead?

yungchin:
Cool, very happy with the addition of `held_offsets`. The deadlock fix looks perfectly good too (it did take me quite some time to get my head around that commit - it's become quite a challenge to reason about where we may grab and release `fetch_lock`!)

The new way of handling both timestamps and actual offsets seems a neat solution to me. I should add though that I don't feel particularly qualified to comment on that bit, given that I've only a poor understanding of Kafka's OffsetRequest semantics so far. Probably best if @kbourgoin can give that a read too.

I've added one new note (see above), which seems important for repeating failed requests. Other than that, all happy!

Issue #189, opened 2015-06-22
Long-lived pykafka consumer occasionally freezes.


rduplain:
Reporting this bug as I've heard it through @amontalenti. I don't have any information, as I'm new to the project, but want to post this issue to get other reports from others.

msukmanowsky:
@rduplain are you using the BalancedConsumer or SimpleConsumer? Also, any logs from around the time the consumer freezes?

rduplain:
I heard this issue through @amontalenti. The questions you are asking @msukmanowsky are exactly what I'd like to see too to assist in addressing the problem. I opened the issue, because it didn't look like one already existed. This is now an open call for those with details to post to the thread.

msukmanowsky:
Gotcha! We'll try to contribute what we've seen in and around when our BalancedConsumer does this. Think @kbourgoin or @emmett9001 may have some historical logs (though I don't think they tell much). Pretty sure this issue is isolated to the BalancedConsumer and likely some threading issues, don't think the SimpleConsumer suffers from any problems here.

yungchin:
As noted in comments elsewhere, I mentioned to @kbourgoin seeing lots of freezes in the test suite - always on tests involving a BalancedConsumer. I need to figure out how to get nose to give me more info, as --nologcapture doesn't seem to do the trick.

yungchin:
Ok, just caught log output from a stalled test. The test in question is [this one](https://github.com/Parsely/pykafka/blob/0431280d6e2a68736e04fd91614f2dc028f087a6/tests/pykafka/test_producer.py#L21), and the log output is [in a gist here](https://gist.github.com/yungchin/543e3a8d640407c57ebb). The test was run like so: `nosetests --stop --nologcapture --nocapture --logging-config tests/debug-logging.conf tests.pykafka.test_producer:ProducerIntegrationTests > /tmp/nosetests.logging`. Entries from 21:55 (L138) onwards were prompted by me hitting ctrl-c. When it doesn't stall, this test finishes in 15 seconds or so.

emmett9001:
Is this a rebalancing loop? Looks like it rebalanced at least ten times in quick succession.

yungchin:
Could that be because the topic/partitions have only just been created (in test setup) and so the brokers are still settling?

I've updated the gist with sections of the server logs, starting from 21:52 (so about 20 seconds before the test starts).

rduplain:
Now that I have the test runner going, I am seeing similar behavior as @yungchin:

[...]

PR #195, opened 2015-06-27
Avoid test_producer stalling randomly

yungchin:
This works around an issue mentioned in, but probably not the subject of, #189. Details in commit messages.

### DIFF
--- a/tests/pykafka/test_producer.py
+++ b/tests/pykafka/test_producer.py
@@ -25,7 +25,8 @@ class ProducerIntegrationTests(unittest2.TestCase):
             # produced in a previous test
             payload = uuid4().bytes
 
-            prod = self.client.topics[self.topic_name].get_producer(batch_size=5)
+            prod = self.client.topics[self.topic_name].get_producer(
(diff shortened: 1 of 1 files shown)


tests/pykafka/test_producer.py
```diff
@@ -20,12 +21,19 @@ def tearDownClass(cls):
...
+            # produced in a previous test
+            payload = uuid4().bytes
+
+            prod = self.client.topics[self.topic_name].get_producer(
+                batch_size=5)
+            prod.produce([payload])
+
+            # set a timeout so we don't wait forever if we break producer code
+            consumer = self.client.topics[self.topic_name].get_simple_consumer(
+                consumer_timeout_ms=1000)
```
emmett9001: Good thing I just fixed the bug that makes this work the way it's supposed to! https://github.com/Parsely/pykafka/commit/3ffa301926825a86081a5603b4945a6bbc90f58e

emmett9001:
:+1: I approve

PR #203, opened 2015-07-14
consistent reset_offsets

emmett9001:
This pull request ensures that the offsets supplied during a call to `reset_offsets` will be saved to the consumer's internal counters even in cases where the maximum number of retries was reached.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -421,7 +421,7 @@ class SimpleConsumer():
         """Reset offsets for the specified partitions
 
         Issue an OffsetRequest for each partition and set the appropriate
-        returned offset in the OwnedPartition
+        returned offset in the consumer's internal offset counter.
 
         :param partition_offsets: (`partition`, `timestamp_or_offset`) pairs to
             reset where `partition` is the partition for which to reset the offset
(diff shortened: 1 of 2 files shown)


emmett9001:
@yungchin let me know if this looks sane to you.

pykafka/simpleconsumer.py
```diff
@@ -525,10 +523,18 @@ def _handle_success(parts):
...
             for errcode, owned_partitions in parts_by_error.iteritems():
                 if errcode != 0:
                     for owned_partition in owned_partitions:
+                        # set internal counter appropriately to avoid possibly
+                        # leaving it in a state inconsistent with the given
+                        # offset
+                        given_offset = owned_partition_offsets[owned_partition]
+                        # don't reset counter to a magic value, only to a real
+                        # offset
+                        if given_offset != self._auto_offset_reset:
```
yungchin: Not entirely sure if I got that bit, but could `given_offset` also be a timestamp at this point?

emmett9001: It could. If it is, it will result in an error on the next fetch attempt and will be reset to `auto_offset_reset`.

PR #205, opened 2015-07-16
Bugfix/zookeeper failure

fortime:
Stop the consumer if any of its worker threads exits.


emmett9001:
I'd love some second opinions on this - @yungchin @kbourgoin @rduplain

Issue #204, opened 2015-07-15
BalancedConsumer incorrectly handles loss of its ephemeral zookeeper nodes


fortime:
Once the connection from one consumer to zookeeper losts, the consumer_id on zookeeper will be deleted. After reconnecting to zookeeper, rebalancing of this consumer will fail. The following line at func `_decide_partitions`:
```python
participants.index(self._consumer_id)`
```
will raise an exception.
Finally, all threads calling `_rebalance` will exit. But the worker thread still fetch messages from broker.

I think that it should check if the consumer is still on zookeeper when `_consumers_changed` is called. If not, calling `_add_self` registers this consumer again. In the meantime, if register failed, we should exit all other threads. BTW, can you provide a function to check if the consumer is running?

fortime:
I think if the internal consumer of BalancedConsumer should stop if the connection to zookeeper lost. Other consumers may consume its partitions when it tries to reconnect zookeeper.

emmett9001:
Hi again, @fortime! I've dug a bit into this issue and what you're saying makes sense. I'm still not totally clear on when exactly this situation arises - under what circumstances is the connection to zookeeper lost? Can you provide some example code that shows a case where the zookeeper connection can die in this way? I haven't been able to trigger the error you mention at `participants.index()`.

emmett9001:
I've started some fixes on [this branch](https://github.com/Parsely/pykafka/tree/bugfix/zookeeper_failure), but I haven't yet fully thought through the consequences of the change, or if it's actually solving the right thing.

fortime:
I think that it should take care if all watcher is alive. A simple solution is to stop the consumer and let the user restart a new one once a watcher is not alive.

fortime:
Hi @emmett9001 , I have created a PR: #205 . A simple way to handle any failure which happens in worker thread.

yungchin:
Here's an updated summary, also taking into account some details implied in pullreq #215:

After a zookeeper reconnect, `BalancedConsumer` should handle the potential loss of its ephemeral nodes. That means that

  1. it cannot assume that `self._consumer_id` is still a listed participant on zookeeper (and we might be able to salvage some logic from #206 to fix that)
[...]

PR #206, opened 2015-07-16
zookeeper failure

emmett9001:
This pull request allows the `BalancedConsumer` to recover from some ZooKeeper failure scenarios. Related to https://github.com/Parsely/pykafka/issues/204

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -26,12 +26,13 @@ import time
 from uuid import uuid4
 
 from kazoo.client import KazooClient
-from kazoo.exceptions import NoNodeException, NodeExistsError
+from kazoo.exceptions import (NoNodeException, NodeExistsError, ConnectionLoss,
+                              ConnectionClosedError)
 from kazoo.recipe.watchers import ChildrenWatch
 
 from .common import OffsetType
 from .exceptions import (KafkaException, PartitionOwnedError,
-                         ConsumerStoppedException)
+                         ConsumerStoppedException, ZookeeperConnectionLost)
 from .simpleconsumer import SimpleConsumer
 
 
@@ -238,13 +239,14 @@ class BalancedConsumer():
 
         This method should be called as part of a graceful shutdown process.
         """
-        self._zookeeper.stop()
         # If internal consumer is not running and this consumer is running,
         # consume() will re-setup the internal consumer.
         # To avoid a race condition, set this consumer to not running before
         # stopping internal consumer.
         self._running = False
-        self._consumer.stop()
+        if self._consumer is not None:
+            self._consumer.stop()
+        self._zookeeper.stop()
 
     @property
     def running(self):
@@ -270,6 +272,7 @@ class BalancedConsumer():
         disable its workers and mark it for garbage collection before
         creating a new one.
         """
+        log.info("Resetting internal consumer")
         reset_offset_on_start = self._reset_offset_on_start
         if self._consumer is not None:
             self._consumer.stop()
@@ -401,7 +404,7 @@ class BalancedConsumer():
         participants = self._get_participants()
         if self._consumer_id in participants:
             return
-        if len(self._topic.partitions) <= len(participants):
+        if len(self._topic.partitions) < len(participants):
             self.stop()
             raise KafkaException("Cannot add consumer: more consumers than partitions")
 
@@ -449,6 +452,9 @@ class BalancedConsumer():
                                     ex.partition, i)
                         raise
                     log.info('Unable to acquire partition %s. Retrying', ex.partition)
+                    time.sleep(i * (self._rebalance_backoff_ms / 1000))
+                except (ConnectionLoss, ConnectionClosedError):
+                    log.error("Zookeeper connection lost. Retrying.")
                     time.sleep(i * (self._rebalance_backoff_ms / 1000))
 
     def _path_from_partition(self, p):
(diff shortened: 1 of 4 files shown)


emmett9001:
@kbourgoin @rduplain @yungchin Let me know if anyone has comments on this. I think the `try...finally` around the worker thread logic is just fine, as is the automatic resetting of the internal consumer. My main concern is whether it makes sense to call `_add_self()` on consumer change, since I think it can result in an infinite rebalancing loop.

fortime:
hi @emmett9001 , how is it going?

pykafka/balancedconsumer.py
```diff
@@ -539,13 +556,20 @@ def consume(self, block=True):
...
             if self._consumer_timeout_ms == -1:
                 return False
             disp = (time.time() - self._last_message_time) * 1000.0
             return disp > self._consumer_timeout_ms
+
+        # auto-restart the internal consumer if it has stalled
+        if not self._consumer.running:
+            if not self._running:
+                return None
+            self.commit_offsets()
```
yungchin: I think this `commit_offsets` should go in `SimpleConsumer.stop()` instead, because it's basically compensating for the fact that with this pullreq, `stop` may be called unexpectedly from any worker thread. If we only do it here in `BalancedConsumer`, then running a bare instance of `SimpleConsumer` might lead to using stale offsets when an exception occurs.

pykafka/simpleconsumer.py
```diff
@@ -295,6 +308,9 @@ def consume(self, block=True):
         :param block: Whether to block while waiting for a message
         :type block: bool
         """
+        if not self._running:
+            return None
```
yungchin: This check is already handled below in L331, where it raises an exception instead, so these two lines should probably go?

pykafka/balancedconsumer.py
yungchin: PS also, it probably needs to be wrapped in `if self._consumer_group is not None`

emmett9001: Oof, good catch. I agree it makes more sense in the `SimpleConsumer`.

pykafka/balancedconsumer.py
```diff
@@ -509,6 +525,7 @@ def _consumers_changed(self, consumers):
         if self._setting_watches:
             return
         log.debug("Rebalance triggered by consumer change")
+        self._add_self()
```
yungchin: Could this get a short comment with it? I don't fully grasp what situation it's needed in.

yungchin:
@emmett9001 apologies for the delay, I missed this one. I think the `_add_self` is ok - iiuc it's a no-op when the instance is already in the list of participants, so that should be sufficient to prevent infinite loops I think?

pykafka/balancedconsumer.py
emmett9001: If the zookeeper connection dies, the consumer's ID is removed from ZK's registry. When the zookeeper connection is regained, the consumer needs to actively add itself to that registry. Since we can't count on the zookeeper connection failing in a particular piece of code, this seems like the best place to handle it. If we have an active ZK connection, this function will be called by its watch and `_add_self` will happen.

emmett9001:
@yungchin Yup, thanks for the comments. You've helped me clarify the most dangerous-seeming bits of this change, so I think it's good to merge now.

pykafka/balancedconsumer.py
yungchin: Wow, ok, got it. Thanks!!

pykafka/balancedconsumer.py
```diff
@@ -235,8 +239,17 @@ def stop(self):
         This method should be called as part of a graceful shutdown process.
         """
         self._zookeeper.stop()
-        self._consumer.stop()
+        # If internal consumer is not running and this consumer is running,
```
yungchin: Final question (I hope) and not actually new in this pullreq:
shouldn't `_zookeeper.stop` actually come last here? As long as `_consumer` still runs it might commit offsets, so it should keep exclusive hold of its partitions, right?

emmett9001: Great point.

yungchin:
Ok, actually, can I be a chicken? The auto-restart bit in `BalancedConsumer.consume` scares me. I think the fact that we now `stop` on any exceptions is a really good improvement, so it won't all stall silently, but wouldn't that be good enough by itself?

So then, if you call `consume`, you'd just get a `ConsumerStoppedException` rather than it restarting things under the hood. It's not a super big deal - if you prefer the auto-restarting, this is ready to merge, for me.

Only Travis isn't happy but that's just a missing mock attribute I think?

emmett9001:
@yungchin I think I agree with you. The consequences of automatically restarting while `consume` is in progress are hard to reason about, but I think it's possible for the consumer to, at least for a short period, be consuming the same partitions as another.

emmett9001:
@fortime, please try your test case on this branch and let me know if your original error is properly handled.

fortime:
@emmett9001 ok, i will have some tests on it.

pykafka/balancedconsumer.py
```diff
@@ -386,7 +407,10 @@ def _add_self(self):
         number of partitions.
         """
         participants = self._get_participants()
-        if len(self._topic.partitions) <= len(participants):
+        if self._consumer_id in participants:
+            return
+        if len(self._topic.partitions) < len(participants):
```
fortime: why do you remove the equal condition? If the number of participants and the number of partitions are equal, the consumer should not register itself.

fortime:
@emmett9001 I have tested on the branch and found some problems. And I have created two PRs to fix them. #214 #215. 
There are two problems too.
One is from kazoo: when we call zookeeper.stop(), it may freeze. After some debugs, it seems its _STOP signal is missing in the queue.
The other is a concurrent problem of _add_self. For example, if there are 3 or more consumers which register themselves on a topic with 2 partitions. There is a certain probability that all of them will register successfully. But only two of them will call _setup_internal_consumer(). Finally, when we call consume(), it will raise an exception "'NoneType' object has no attribute 'consume'"

yungchin:
@fortime I'm afraid we've let this branch go stale. There was always some difficulty with it because it addressed more than one issue, making it hard to review it.

I've spun out as many separate tickets as I could identify in the commits here, so that we can basically start over on this. The main two things here though were 1) failure handling for the consumer workers, which @emmett9001 is creating a more comprehensive solution for under #235, and 2) handling loss of our ephemeral nodes on zookeeper, which as you point out in #215 is not completely correctly handled in this pullreq, so at this point we should probably give that a fresh start, too. I'll update #204 with the relevant details.

PR #208, opened 2015-07-22
Reset offsets for all brokers in simpleconsumer.py

crispin-linden:
Breaking here prevents the rest of the by_leader dict from being processed. Since we've successfully read the offsets of the current broker, we should continue the iteration instead.


wer-linden:
+1  without this fix, only the first partition/group  is having it's current offset updated.   This  causes a balanced_consumer to consume from only one partition (ignoring the rest if they exist), when reset_offset_on_start=True is used.

emmett9001:
Thanks, this change in combination with [this one](https://github.com/Parsely/pykafka/commit/4cbe0a16407e2edcc7336e404c4f102269a9dd09) should fix a few of the retry bugs.

PR #211, opened 2015-07-23
Add some check when stopping balancedconsumer

fortime:
Add some check when stopping balancedconsumer


emmett9001:
@fortime the branch you're pull requesting against has changed since this was created, but I've applied the change in that branch [here](https://github.com/Parsely/pykafka/commit/48c7225196b89ff46a14f3f4d397902ddc81194f).

fortime:
Thanks

Issue #210, opened 2015-07-22
Stopping BalancedConsumer ends in infinite loop


tiriplicamihai:
I noticed today a problem with BalancedConsumer. The steps to reproduce this behavior are:

1. Call consume method (blocking or nonblocking)
2. On another thread get the instance and call `stop` method (For my scenario it was in a signal handler)
3. The consume will never end

I think this happens because of this code:
https://github.com/Parsely/pykafka/blob/1.0.3/pykafka/simpleconsumer.py#L297-L310

`SimpleConsumer` always raises `ConsumerStoppedException` and `BalancedConsumer` always ignores it (`continue` skips the blocking part).

Is this the expected behavior? If not, maybe before calling `continue`, self._running should be check. If you thing it's ok I'd be happy to create a PR for this fix (or think about another one).

Thank you!

emmett9001:
Thanks @tiriplicamihai. My original intent with the `continue` statement was to allow `consume()` to continue without returning during a call to `_rebalance()`. That can stay, but I agree that a check on `self._running` before the `continue` in `BalancedConsumer` should fix this issue. If you're willing to open a pull request for this, please do!

tiriplicamihai:
@emmett9001 Thank you for your response! I will open a PR tomorrow.

PR #212, opened 2015-07-23
210 fix stop balanced consumer

tiriplicamihai:
Fixes https://github.com/Parsely/pykafka/issues/210

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -553,6 +553,8 @@ class BalancedConsumer():
             try:
                 message = self._consumer.consume(block=block)
             except ConsumerStoppedException:
+                if hasattr(self, '_running') and not self._running:
+                    return
                 continue
             if message:
                 self._last_message_time = time.time()
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -163,6 +163,7 @@ class BalancedConsumer():
         self._zookeeper_connect = zookeeper_connect
         self._zookeeper_connection_timeout_ms = zookeeper_connection_timeout_ms
         self._reset_offset_on_start = reset_offset_on_start
+        self._running = False
 
         self._rebalancing_lock = cluster.handler.Lock()
         self._consumer = None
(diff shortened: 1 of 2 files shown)


tiriplicamihai:
@emmett9001 can you please review the changes? As a feedback, I think you should add a `requirements.txt` file in the repo. I couldn't find the right versions of the packages and I could not run the integration tests for BalancedConsumer. I can help with this too if you can tell me the versions.

emmett9001:
@tiriplicamihai Did you install pykafka with `python setup.py develop`? This should install all of the necessary packages.

tiriplicamihai:
Sorry, I didn't. Now seems to work ok. Thank you!

pykafka/balancedconsumer.py
```diff
@@ -553,6 +553,8 @@ def consumer_timed_out():
             try:
                 message = self._consumer.consume(block=block)
             except ConsumerStoppedException:
+                if hasattr(self, '_running') and not self._running:
```
emmett9001: Why do you need to check `hasattr` here? It seems unnecessary to me.

tiriplicamihai: `_running` is set on `self` when start is called. If, for some reason, somebody calls consume before start I thought it will not be ok to see an AttributeError. The other solution is to set `self._running` to false in `__init__`. Do you think it's ok?

emmett9001: Yes, it's an oversight on my part that `_running` isn't set to `False` on `__init__`. I think that's preferable to using `hasattr` in this case.

tiriplicamihai: Done :)

emmett9001:
Looks great. Thanks for the test, too!

tiriplicamihai:
Happy to help! If there are any other issues I can help with, you can ping me (I'm looking for an OS project I can contribute to).

emmett9001:
Wonderful, I'll keep that in mind. Please feel free to contribute issues or pull request fixes and improvements if you ever feel inclined to do so. I'd also recommend joining the [mailing list](https://groups.google.com/forum/#!forum/pykafka-user). Thanks!

PR #213, opened 2015-07-26
Tests for SimpleConsumer.reset_offsets

yungchin:
This adds tests for reset_offsets (which I'm also looking to use to ensure the implementation of reset_offsets in #176 will be conformant), and fixes some issues with that method which I encountered in the course of writing tests.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -521,17 +521,16 @@ class SimpleConsumer():
 
                 time.sleep(i * (self._offsets_channel_backoff_ms / 1000))
 
+                for errcode, owned_partitions in parts_by_error.iteritems():
+                    if errcode != 0:
+                        for owned_partition in owned_partitions:
+                            owned_partition.fetch_lock.release()
 
-            for errcode, owned_partitions in parts_by_error.iteritems():
-                if errcode != 0:
-                    for owned_partition in owned_partitions:
-                        owned_partition.fetch_lock.release()
-
-            if len(parts_by_error) == 1 and 0 in parts_by_error:
+            if not owned_partition_offsets:
                 break
             log.debug("Retrying offset reset")
 
-        if any([a != 0 for a in parts_by_error]):
+        if owned_partition_offsets:
             raise OffsetRequestFailedError("reset_offsets failed after %d "
                                            "retries",
                                            self._offsets_reset_max_retries)
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -497,9 +497,6 @@ class SimpleConsumer():
                     partition.flush()
                     by_leader[partition.partition.leader].append((partition, offset))
 
-            # reset this dict to prepare it for next retry
-            owned_partition_offsets = {}
-
             # get valid offset ranges for each partition
             for broker, offsets in by_leader.iteritems():
                 reqs = [owned_partition.build_offset_request(offset)
@@ -522,12 +519,10 @@ class SimpleConsumer():
 
                 if 0 in parts_by_error:
                     parts_by_error.pop(0)
-                errored_partitions = dict(
+                owned_partition_offsets = dict(
                     (part, owned_partition_offsets[part])
                     for errcode, parts in parts_by_error.iteritems()
                     for part, _ in parts)
-                owned_partition_offsets = dict(errored_partitions.items() +
-                                               owned_partition_offsets.items())
 
             for errcode, owned_partitions in parts_by_error.iteritems():
                 if errcode != 0:
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -508,7 +508,11 @@ class SimpleConsumer():
                     success_handler=_handle_success,
                     partitions_by_id=self._partitions_by_id)
 
-                if len(parts_by_error) == 1 and 0 in parts_by_error:
+                if 0 in parts_by_error:
+                    # drop successfully reset partitions for next retry
+                    successful = [part for part, _ in parts_by_error.pop(0)]
+                    map(owned_partition_offsets.pop, successful)
+                if not parts_by_error:
                     continue
                 log.error("Error resetting offsets for topic %s (errors: %s)",
                           self._topic.name,
@@ -517,12 +521,6 @@ class SimpleConsumer():
 
                 time.sleep(i * (self._offsets_channel_backoff_ms / 1000))
 
-                if 0 in parts_by_error:
-                    parts_by_error.pop(0)
-                owned_partition_offsets = dict(
-                    (part, owned_partition_offsets[part])
-                    for errcode, parts in parts_by_error.iteritems()
-                    for part, _ in parts)
 
             for errcode, owned_partitions in parts_by_error.iteritems():
                 if errcode != 0:
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -636,7 +636,18 @@ class OwnedPartition(object):
         return self._messages.qsize()
 
     def flush(self):
-        self._messages = Queue()
+        """Flush internal queue"""
+        # This is a slow way to empty the queue, but the safest solution that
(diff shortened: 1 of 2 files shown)


yungchin:
@emmett9001 I ran into an issue that came in with 4cbe0a1 (see commit message of 50561ee for details), and have put in an alternative fix here with 3cea51c, and some more changes in 3e76c15. I'd be grateful if you can check on those two commits in particular, because `reset_offsets` logic is just so complex, so there's a real chance I'm missing the point.

tests/pykafka/test_simpleconsumer.py
```diff
@@ -55,18 +65,46 @@ def test_offset_resume(self):
...
+            # We can't control which partition(s) our setUpClass produce calls
+            # write to, so we don't know on which partitions the following will
+            # result in a valid offset. But at least one will be valid, which
+            # is good enough.
+            custom_offs = [(consumer.partitions[pid], latest - 5)
+                           for pid, latest in latest_offs.items()]
+            consumer.reset_offsets(partition_offsets=custom_offs)
+            msg = consumer.consume()
+            # FIXME this 4 should be a 5 or I'm getting something else wrong:
+            self.assertEqual(msg.offset, latest_offs[msg.partition_id] - 4)
```
yungchin: This detects an off-by-one error, currently (I've committed -4 here to make the tests pass, but it should be -5 as per L102 above). I've no idea yet whether that's an error in the test code or elsewhere: any insight much appreciated.

Here's the test log:
```
======================================================================
FAIL: Test resetting to user-provided offsets
----------------------------------------------------------------------
  File "/srv/pykafka/tests/pykafka/test_simpleconsumer.py", line 107, in test_reset_offsets
    self.assertEqual(msg.offset, latest_offs[msg.partition_id] - 5)
AssertionError: 295 != 294
    '295 != 294' = '%s != %s' % _common_shorten_repr(295, 294)
    '295 != 294' = self._formatMessage('295 != 294', '295 != 294')
>>  raise self.failureException('295 != 294')
    
-------------------- >> begin captured logging << --------------------
pykafka.simpleconsumer: INFO: Resetting offsets for 3 partitions
pykafka.simpleconsumer: INFO: Flushed queue for partition 1
pykafka.simpleconsumer: INFO: Flushed queue for partition 0
pykafka.simpleconsumer: INFO: Flushed queue for partition 2
pykafka.simpleconsumer: DEBUG: Fetcher thread exiting
pykafka.simpleconsumer: INFO: Starting 1 fetcher threads
pykafka.simpleconsumer: INFO: Resetting offsets for 3 partitions
pykafka.simpleconsumer: INFO: Flushed queue for partition 0
pykafka.simpleconsumer: INFO: Flushed queue for partition 2
pykafka.simpleconsumer: INFO: Flushed queue for partition 1
pykafka.simpleconsumer: WARNING: Offset reset for partition 1 to timestamp -6 failed. Setting partition 1's internal counter to -6
pykafka.simpleconsumer: INFO: Resetting offsets in response to OffsetOutOfRangeError
pykafka.simpleconsumer: INFO: Resetting offsets for 1 partitions
pykafka.simpleconsumer: INFO: Flushed queue for partition 1
pykafka.simpleconsumer: WARNING: Offset reset for partition 0 to timestamp 594 failed. Setting partition 0's internal counter to 594
pykafka.simpleconsumer: WARNING: Offset reset for partition 2 to timestamp 294 failed. Setting partition 2's internal counter to 294
pykafka.simpleconsumer: DEBUG: Fetched 5 messages for partition 0
pykafka.simpleconsumer: DEBUG: Partition 0 queue holds 5 messages
pykafka.simpleconsumer: DEBUG: Fetched 5 messages for partition 2
pykafka.simpleconsumer: DEBUG: Partition 2 queue holds 5 messages
--------------------- >> end captured logging << ---------------------
```

yungchin: Ah - the problem is in the test assertion of course: we're comparing a `last_offset_consumed` to the next `msg.offset`, so these _should_ be off by one. Sorry for the noise.

pykafka/simpleconsumer.py
```diff
@@ -520,25 +521,16 @@ def _handle_success(parts):
...
-                    for errcode, parts in parts_by_error.iteritems()
-                    for part, _ in parts)
-                owned_partition_offsets = dict(errored_partitions.items() +
-                                               owned_partition_offsets.items())
-
-            for errcode, owned_partitions in parts_by_error.iteritems():
-                if errcode != 0:
-                    for owned_partition in owned_partitions:
-                        owned_partition.fetch_lock.release()
+                for errcode, owned_partitions in parts_by_error.iteritems():
```
emmett9001: This is a good catch, though I'm surprised it wasn't already fixed. I could've sworn I made this exact change somewhere else.

pykafka/simpleconsumer.py
```diff
@@ -511,7 +508,11 @@ def _handle_success(parts):
                     success_handler=_handle_success,
                     partitions_by_id=self._partitions_by_id)
 
-                if len(parts_by_error) == 1 and 0 in parts_by_error:
+                if 0 in parts_by_error:
+                    # drop successfully reset partitions for next retry
```
emmett9001: This is a much better way to do what I was trying to do.

tests/pykafka/test_simpleconsumer.py
```diff
@@ -55,18 +66,61 @@ def test_offset_resume(self):
...
+        difference = sum(latest_minus_one[i] - earliest_minus_one[i]
+                         for i in latest_minus_one.keys())
+        self.assertEqual(difference, self.total_msgs)
+
+    def test_reset_offsets(self):
+        """Test resetting to user-provided offsets"""
+        with self._get_simple_consumer(
+                auto_offset_reset=OffsetType.LATEST,
+                reset_offset_on_start=True) as consumer:
+            part_id, latest_offset = next(  # get a non-empty partition
```
emmett9001: The name `latest_offset` is maybe confusing, since it's really the result of `held_offsets`.

emmett9001:
The tests look great, and as far as I can tell the important invariants of `reset_offsets` are maintained. The code is also cleaner, which is good. It would be great if you could explain a bit more what issue commit https://github.com/Parsely/pykafka/commit/50561eeb793f0a72edfe41c71f80e457f30020ab is meant to solve, since I don't fully understand it. I'll also let @kbourgoin weigh in if he's interested.

yungchin:
Thanks for checking this! So 50561ee is the result of a `git revert 4cbe0a1` - which was needed because the latter's `owned_partition_offsets = {}` caused an error, when this line tried to access it after it got emptied: https://github.com/Parsely/pykafka/blob/feature/test_reset_offsets/pykafka/simpleconsumer.py#L464

I then tried to do the same thing with 3cea51c, but wanted to check with you that that really does all the things it should do.

yungchin:
I'm afraid I've one or two more commits coming up, because the last test is now so strict that there's no good way for `RdKafkaSimpleConsumer` to satisfy it :)

(The difference in behaviour is, that when you hand SimpleConsumer an invalid offset, it almost immediately detects that, through the fetcher thread, and so a split second afterwards you've already got an auto-resetted valid offset. Whereas in RdKafkaSimpleConsumer, the held offsets only get updated once you've consumed from that partition, so it takes until after a consume() before the valid offset appears.)

emmett9001:
Thanks for the explanation, the error case makes sense now.

tests/pykafka/test_simpleconsumer.py
yungchin: Ah yes, thanks. I was being lazy here, not wanting to go through `topic.latest_available_offsets()` but I agree that makes it very unclear what the intent is. I've changed it now, so it reads from latest_available_offsets as it should.

yungchin:
Ok, apologies for slapping on more commits, I missed some problems jumping back and forth between this branch and the rdkafka branch. Should now be really actually totally ready.

The additional changes affect `OwnedPartition.flush` and `test_reset_offsets` (in particular, the part where it tries invalid offsets).

PR #214, opened 2015-07-26
check if the internal consumer is running.

fortime:
The rebalance may have been finished before we check if we are rebalancing. So we check if the internal consumer is running before we raise an exception.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -594,6 +594,9 @@ class BalancedConsumer():
                 # don't raise the exception if we're rebalancing
                 if self._rebalancing_lock.locked():
                     continue
+                # rebalancing has been finished
+                if self._consumer.running:
+                    continue
                 raise
             if message:
                 self._last_message_time = time.time()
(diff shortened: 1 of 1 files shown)


yungchin:
:+1:

pykafka/balancedconsumer.py
```diff
@@ -594,6 +594,9 @@ def consumer_timed_out():
                 # don't raise the exception if we're rebalancing
                 if self._rebalancing_lock.locked():
                     continue
+                # rebalancing has been finished
+                if self._consumer.running:
```
emmett9001: Seems like this could be combined with the above check like this:

```python
if self._rebalancing_lock.locked() or self._consumer.running:
    continue
```

fortime: Oh, yes. I will change it.

emmett9001:
It's not clear why this change is needed. Can you further explain the failure case?

yungchin:
IIUC, there's a small chance that a rebalance is going on, which finishes just between receiving the `ConsumerStoppedException` and checking the rebalancing lock on the next line - so at that point you expect the new `self._consumer` to have started. So then this change would avoid propagating the exception when it's not really necessary.

emmett9001:
Makes sense to me

PR #217, opened 2015-07-27
Fixing a typo in topic.py, partiton instead of partitions

AgustinCB:


kbourgoin:
Thanks for the catch!

Issue #219, opened 2015-07-28
Client.py - KafkaClient, source ip address and port argument


madhavan020985:
Can KafkaClient in client.py be enhanced to take a source ip address and port argument. https://github.com/Parsely/pykafka/blob/master/pykafka/client.py#L38

If i am not wrong, KafkaClient constructor creates a tcp socket internally using socket.create_connection which supports source ip address/port argument (since python 2.7)

This will be greatly helpful when running consumers on public servers where we don't prefer to keep many ports open.

emmett9001:
This seems reasonable to me. It doesn't really affect any users who don't want to set it, and it's helpful for those who do.

emmett9001:
Pull requested here https://github.com/Parsely/pykafka/pull/221

PR #221, opened 2015-07-28
Allow KafkaClient to configure source_address

emmett9001:
This pull request adds a kwarg `source_address` to `KafkaClient`. This is used in `socket.create_connection` to set the source of the connection.

Related to #219

### DIFF
--- a/pykafka/broker.py
+++ b/pykafka/broker.py
@@ -46,7 +46,9 @@ class Broker():
                  handler,
                  socket_timeout_ms,
                  offsets_channel_socket_timeout_ms,
-                 buffer_size=1024 * 1024):
+                 buffer_size=1024 * 1024,
+                 source_host='',
+                 source_port=0):
         """Create a Broker instance.
 
         :param id_: The id number of this broker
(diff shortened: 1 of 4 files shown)


pykafka/broker.py
```diff
@@ -46,7 +46,9 @@ def __init__(self,
                  handler,
                  socket_timeout_ms,
                  offsets_channel_socket_timeout_ms,
-                 buffer_size=1024 * 1024):
+                 buffer_size=1024 * 1024,
+                 source_host='',
+                 source_port=0):
```
kbourgoin: Such weird defaults....  (yes, i know they're the socket defaults. just weird.)

pykafka/connection.py
```diff
@@ -63,7 +76,8 @@ def connect(self, timeout):
         """Connect to the broker."""
         self._socket = socket.create_connection(
             (self.host, self.port),
-            timeout=timeout / 1000
+            timeout / 1000,
+            self.source_host, self.source_port
```
kbourgoin: If these are kwargs, that would be preferable to using over positional.

kbourgoin:
Travis failed, so there's something up with the socket creation. Once that's fixed, this looks good to me. Might want to put a link to the python documentation for that setting in a docstring (maybe the client?) so that it's obvious what it is. It's not a setting I've seen before, so it might help other people too.

pykafka/connection.py
emmett9001: [They don't appear to be kwargs](https://docs.python.org/2/library/socket.html#socket.create_connection)

kbourgoin: I was afraid that was the case. The stdlib docs are super confusing about that sometimes and functions will sometimes accept the kwarg when they shouldn't.

PR #231, opened 2015-08-20
Port to python3.

sontek:
This makes all but one of the tests pass so far.  I think its the closest to a workable version of py3 (based on looking at the other PRs)  can someone like @emmett9001 @mwhooker or @kbourgoin review it?

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -33,7 +33,7 @@ from .common import OffsetType
 from .exceptions import (KafkaException, PartitionOwnedError,
                          ConsumerStoppedException)
 from .simpleconsumer import SimpleConsumer
-from .utils.compat import range
+from .utils.compat import range, get_bytes
 
 
 log = logging.getLogger(__name__)
@@ -302,7 +302,9 @@ class BalancedConsumer():
         :type participants: Iterable of str
         """
         # Freeze and sort partitions so we always have the same results
-        p_to_str = lambda p: '-'.join([p.topic.name, str(p.leader.id), str(p.id)])
+        p_to_str = lambda p: b'-'.join([
+            get_bytes(p.topic.name), bytes(p.leader.id), bytes(p.id)]
+        )
         all_parts = self._topic.partitions.values()
         all_parts = sorted(all_parts, key=p_to_str)
 
@@ -321,6 +323,7 @@ class BalancedConsumer():
         log.info('Balancing %i participants for %i partitions.\nOwning %i partitions.',
                  len(participants), len(all_parts), len(new_partitions))
         log.debug('My partitions: %s', [p_to_str(p) for p in new_partitions])
+
         return new_partitions
 
     def _get_participants(self):
@@ -416,7 +419,6 @@ class BalancedConsumer():
                     # partition allocation is correct.
                     participants = self._get_participants()
                     partitions = self._decide_partitions(participants)
-
                     old_partitions = self._partitions - partitions
                     self._remove_partitions(old_partitions)
 
@@ -470,7 +472,7 @@ class BalancedConsumer():
             try:
                 self._zookeeper.create(
                     self._path_from_partition(p),
-                    value=self._consumer_id,
+                    value=get_bytes(self._consumer_id),
                     ephemeral=True
                 )
                 self._partitions.add(p)
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -303,10 +303,10 @@ class BalancedConsumer():
         # Freeze and sort partitions so we always have the same results
         p_to_str = lambda p: '-'.join([p.topic.name, str(p.leader.id), str(p.id)])
         all_parts = self._topic.partitions.values()
-        all_parts.sort(key=p_to_str)
+        all_parts = sorted(all_parts, key=p_to_str)
 
         # get start point, # of partitions, and remainder
-        participants.sort()  # just make sure it's sorted.
+        participants = sorted(participants)  # just make sure it's sorted.
         idx = participants.index(self._consumer_id)
         parts_per_consumer = math.floor(len(all_parts) / len(participants))
         remainder_ppc = len(all_parts) % len(participants)
@@ -343,7 +343,7 @@ class BalancedConsumer():
                     participants.append(id_)
             except NoNodeException:
                 pass  # disappeared between ``get_children`` and ``get``
-        participants.sort()
+        participants = sorted(participants)
         return participants
 
     def _set_watches(self):
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -314,8 +314,8 @@ class BalancedConsumer():
         parts_per_consumer = math.floor(len(all_parts) / len(participants))
         remainder_ppc = len(all_parts) % len(participants)
 
-        start = parts_per_consumer * idx + min(idx, remainder_ppc)
-        num_parts = parts_per_consumer + (0 if (idx + 1 > remainder_ppc) else 1)
+        start = int(parts_per_consumer * idx + min(idx, remainder_ppc))
+        num_parts = int(parts_per_consumer + (0 if (idx + 1 > remainder_ppc) else 1))
 
         # assign partitions from i*N to (i+1)*N - 1 to consumer Ci
         new_partitions = itertools.islice(all_parts, start, start + num_parts)
--- a/pykafka/protocol.py
+++ b/pykafka/protocol.py
@@ -63,6 +63,7 @@ from zlib import crc32
 from .common import CompressionType, Message
 from .exceptions import ERROR_CODES, NoMessagesConsumedError
 from .utils import Serializable, compression, struct_helpers
+from .utils.compat import iteritems, itervalues, buffer, get_bytes
 
 
 log = logging.getLogger(__name__)
@@ -71,7 +72,7 @@ log = logging.getLogger(__name__)
 class Request(Serializable):
     """Base class for all Requests. Handles writing header information"""
     HEADER_LEN = 21  # constant for all messages
-    CLIENT_ID = 'pykafka'
+    CLIENT_ID = b'pykafka'
 
     def _write_header(self, buff, api_version=0, correlation_id=0):
         """Write the header for an outgoing message.
@@ -191,12 +192,15 @@ class Message(Message, Serializable):
         if self.partition_key is None:
             fmt = '!BBii%ds' % len(self.value)
             args = (self.MAGIC, self.compression_type, -1,
-                    len(self.value), self.value)
+                    len(self.value), get_bytes(self.value))
         else:
             fmt = '!BBi%dsi%ds' % (len(self.partition_key), len(self.value))
             args = (self.MAGIC, self.compression_type,
-                    len(self.partition_key), self.partition_key,
-                    len(self.value), self.value)
+                    len(self.partition_key),
+                    get_bytes(self.partition_key),
+                    len(self.value),
+                    get_bytes(self.value))
+
         struct.pack_into(fmt, buff, offset + 4, *args)
         fmt_size = struct.calcsize(fmt)
         crc = crc32(buffer(buff[(offset + 4):(offset + 4 + fmt_size)]))
@@ -334,7 +338,16 @@ class MetadataRequest(Request):
 
         :param topics: Topics to query. Leave empty for all available topics.
         """
-        self.topics = topics or []
+        self._topics = topics or []
+
+    @property
+    def topics(self):
+        if self._topics:
+            return [
+                t.encode('utf-8') for t in self._topics if hasattr(t, 'encode')
+            ]
+
+        return self._topics
 
     def __len__(self):
         """Length of the serialized message, in bytes"""
@@ -454,11 +467,11 @@ class ProduceRequest(Request):
     def __len__(self):
         """Length of the serialized message, in bytes"""
         size = self.HEADER_LEN + 2 + 4 + 4  # acks + timeout + len(topics)
-        for topic, parts in self.msets.iteritems():
+        for topic, parts in iteritems(self.msets):
             # topic name
             size += 2 + len(topic) + 4  # topic name + len(parts)
             # partition + mset size + len(mset)
-            size += sum(4 + 4 + len(mset) for mset in parts.itervalues())
+            size += sum(4 + 4 + len(mset) for mset in itervalues(parts))
         return size
 
     @property
@@ -471,8 +484,8 @@ class ProduceRequest(Request):
         """Iterable of all messages in the Request"""
         return itertools.chain.from_iterable(
             mset.messages
-            for topic, partitions in self.msets.iteritems()
-            for partition_id, mset in partitions.iteritems()
+            for topic, partitions in iteritems(self.msets)
+            for partition_id, mset in iteritems(partitions)
         )
 
     def add_message(self, message, topic_name, partition_id):
@@ -497,11 +510,13 @@ class ProduceRequest(Request):
         struct.pack_into('!hii', output, offset,
                          self.required_acks, self.timeout, len(self.msets))
         offset += 10
-        for topic_name, partitions in self.msets.iteritems():
+        for topic_name, partitions in iteritems(self.msets):
+            topic_name = get_bytes(topic_name)
             fmt = '!h%dsi' % len(topic_name)
-            struct.pack_into(fmt, output, offset, len(topic_name), topic_name, len(partitions))
+            struct.pack_into(fmt, output, offset, len(topic_name),
+                             topic_name, len(partitions))
             offset += struct.calcsize(fmt)
-            for partition_id, message_set in partitions.iteritems():
+            for partition_id, message_set in iteritems(partitions):
                 mset_len = len(message_set)
                 struct.pack_into('!ii', output, offset, partition_id, mset_len)
                 offset += 8
@@ -617,7 +632,7 @@ class FetchRequest(Request):
         """Length of the serialized message, in bytes"""
         # replica + max wait + min bytes + len(topics)
         size = self.HEADER_LEN + 4 + 4 + 4 + 4
-        for topic, parts in self._reqs.iteritems():
+        for topic, parts in iteritems(self._reqs):
             # topic name + len(parts)
             size += 2 + len(topic) + 4
             # partition + fetch offset + max bytes => for each partition
@@ -641,11 +656,15 @@ class FetchRequest(Request):
         struct.pack_into('!iiii', output, offset,
                          -1, self.timeout, self.min_bytes, len(self._reqs))
         offset += 16
-        for topic_name, partitions in self._reqs.iteritems():
+        for topic_name, partitions in iteritems(self._reqs):
+            topic_name = get_bytes(topic_name)
             fmt = '!h%dsi' % len(topic_name)
-            struct.pack_into(fmt, output, offset, len(topic_name), topic_name, len(partitions))
+            struct.pack_into(
+                fmt, output, offset, len(topic_name), topic_name,
+                len(partitions)
+            )
             offset += struct.calcsize(fmt)
-            for partition_id, (fetch_offset, max_bytes) in partitions.iteritems():
+            for partition_id, (fetch_offset, max_bytes) in iteritems(partitions):
                 struct.pack_into('!iqi', output, offset,
                                  partition_id, fetch_offset, max_bytes)
                 offset += 16
@@ -749,7 +768,7 @@ class OffsetRequest(Request):
         """Length of the serialized message, in bytes"""
         # Header + replicaId + len(topics)
         size = self.HEADER_LEN + 4 + 4
-        for topic, parts in self._reqs.iteritems():
+        for topic, parts in iteritems(self._reqs):
             # topic name + len(parts)
             size += 2 + len(topic) + 4
             # partition + fetch offset + max bytes => for each partition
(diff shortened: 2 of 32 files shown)


sontek:
This requires this PR to be merged: https://github.com/Parsely/testinstances/pull/4  since testinstances wasn't python3 compatible either

sontek:
I also added support for py26 support to testinstances because `pykafka` had py26 in its `tox` but the code currently isn't compatible with `py26` (it uses dict comprehensions).  I'm not really interested in 2.6 support but I left it in there in case someone just didn't realize it was broken.

sontek:
You can see the successful py27 run:

https://travis-ci.org/Parsely/pykafka/jobs/76518663

once the `testinstances` stuff is merged and released we can re-run the build for py34

emmett9001:
Hi @sontek, thanks for the pull request. I'll be able to take a deeper look at this tomorrow. We've already decided not to support 2.6, so the fact that it's still in `tox` is an error. We're very close to merging a complete [overhaul of the `Producer` class](https://github.com/Parsely/pykafka/pull/177), after which supporting python 3 will be my primary focus. I think it's important to get that PR merged first so we can update this branch to make it py3 compatible. We also need to figure out a reasonable workflow for running tests under 2.7 and 3 to make sure future code remains compatible with both, so if you have any thoughts on that, they're quite welcome.

emmett9001:
Now that #277 has been merged, this branch can be updated to make `Producer` compatible with python 3. @sontek, are you interested in helping out with that?

sontek:
@emmett9001 Yeah, I'll fix it up

vortec:
Thanks @sontek for your work!

sontek:
@vortec @emmett9001 Just pushed the fix that fixes everything from that latest big merge.   Its back down to just the same test failure on gzip as it was having before.  So its pretty close to being ready

sontek:
```
tests/pykafka/test_protocol.py::TestFetchAPI::test_gzip_decompression FAILED
```

Thats the only failure I'm seeing

yungchin:
Wow - that's a massive amount of work, kudos!!
Only had a brief look at the failing test just yet - I think it might be that `msg` needs to be `bytes`, not `str`?

sontek:
@yungchin Yeah but I built everything else to support not passing bytes and having it just figure it out, thats the one test that fails with those checks in place.  Switching it to bytes fixes it but I feel like we shouldn't have to =)

emmett9001:
@yungchin @sontek how are you running the tests under multiple python interpreters? `tox` fails for me once it tries to install `python-snappy` under 3.4. Ever run into this?

sontek:
@emmett9001 I'm just running `tox -e py34`  whats the error you are getting with snappy?


```
$ mktmpenv -p /usr/bin/python3.4 
Running virtualenv with interpreter /usr/bin/python3.4
Using base prefix '/usr'
New python executable in tmp-5da056596403ca4a/bin/python3.4
Also creating executable in tmp-5da056596403ca4a/bin/python
Installing setuptools, pip...done.
This is a temporary environment. It will be deleted when you run 'deactivate'.
(tmp-5da056596403ca4a)~/venvs/tmp-5da056596403ca4a

$ pip install python-snappy
You are using pip version 6.0.8, however version 7.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Collecting python-snappy
  Using cached python-snappy-0.5.tar.gz
Installing collected packages: python-snappy
  Running setup.py install for python-snappy
    building '_snappy' extension
    x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -g -fstack-protector-strong -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -fPIC -I/usr/include/python3.4m -I/home/sontek/venvs/tmp-5da056596403ca4a/include/python3.4m -c snappymodule.cc -o build/temp.linux-x86_64-3.4/snappymodule.o
    cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++
    x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -g -fstack-protector-strong -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -fPIC -I/usr/include/python3.4m -I/home/sontek/venvs/tmp-5da056596403ca4a/include/python3.4m -c crc32c.c -o build/temp.linux-x86_64-3.4/crc32c.o
    x86_64-linux-gnu-g++ -pthread -shared -Wl,-O1 -Wl,-Bsymbolic-functions -Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,-Bsymbolic-functions -Wl,-z,relro -g -fstack-protector-strong -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 build/temp.linux-x86_64-3.4/snappymodule.o build/temp.linux-x86_64-3.4/crc32c.o -lsnappy -o build/lib.linux-x86_64-3.4/_snappy.cpython-34m.so
Successfully installed python-snappy-0.5
```

sontek:
@emmett9001 This PR is also required for any of this to work:

https://github.com/Parsely/testinstances/pull/4

sontek:
@emmett9001 That error is because you don't have `python3.4-dev` installed:

Ubuntu:

```
sudo apt-get install python3.4-dev
```

yungchin:
On 28 August 2015 at 00:28, Emmett J. Butler <notifications@github.com>
wrote:

> @yungchin <https://github.com/yungchin> @sontek
> <https://github.com/sontek> how are you running the tests under multiple
> python interpreters? tox fails for me once it tries to install
> python-snappy under 3.4. Ever run into this?
>
I hadn't actually tested yet, was just sniping from the sidelines tbh :)
But that error looks like you're missing a python-dev package for py3?

yungchin:
On 28 August 2015 at 00:25, John Anderson <notifications@github.com> wrote:

> @yungchin <https://github.com/yungchin> Yeah but I built everything else
> to support not passing bytes and having it just figure it out, thats the
> one test that fails with those checks in place. Switching it to bytes fixes
> it but I feel like we shouldn't have to =)
>
Ah yeah, got it :)

I do agree that would be more elegant. On the other hand... quite a few of
those protocol methods aren't supposed to ever be handed any unicode in
their regular use, so making them all unicode-friendly would in a sense be
an ongoing waste of cpu cycles. I mean, for this particular case, can't we
safely assume that `FetchResponse.__init__` and the
`struct_helpers.unpack_from` will always get binary strings straight out of
kafka?

emmett9001:
Recommend a way to get tox to use the testinstances branch? I've tried adding 

    -e git+ssh://git@github.com/sontek/testinstances.git@fix_tests#egg=testinstances

to `test-requirements.txt` but tox appears not to like that https://gist.github.com/emmett9001/6c1a22f33a20f6233e18

sontek:
@yungchin Yeah, the problem is backwards compatibility.  All these functions are exposed and probably used in clients already and they already allow strings... so we should probably keep the compatibility and announce it will be going away

sontek:
@emmett9001 I pushed up a change that uses the fix_tests branch of testinstances

pykafka/protocol.py
```diff
@@ -191,16 +192,20 @@ def pack_into(self, buff, offset):
...
+                    get_bytes(self.partition_key),
+                    len(self.value),
+                    get_bytes(self.value))
+
         struct.pack_into(fmt, buff, offset + 4, *args)
         fmt_size = struct.calcsize(fmt)
-        crc = crc32(buffer(buff[(offset + 4):(offset + 4 + fmt_size)]))
-        struct.pack_into('!i', buff, offset, crc)
+        data = buffer(buff[(offset + 4):(offset + 4 + fmt_size)])
+        crc = crc32(data) & 0xffffffff
```
emmett9001: Is this a type conversion, or something else? Is it necessary for py34 compatibility?

pykafka/test/kafka_instance.py
```diff
@@ -128,12 +130,14 @@ class KafkaInstance(ManagedInstance):
     def __init__(self,
                  num_instances=1,
                  kafka_version='0.8.2.1',
+                 scala_version='2.10',
```
emmett9001: :+1:

pykafka/utils/compression.py
```diff
@@ -79,15 +79,22 @@ def encode_snappy(buff, xerial_compatible=False, xerial_blocksize=32 * 1024):
     Adapted from kafka-python
     https://github.com/mumrah/kafka-python/pull/127/files
     """
+    buff = get_bytes(buff)
+    #snappy segfaults if it gets a read-only buffer on PyPy
+    if IS_PYPY:
```
emmett9001: This is kind of ugly - I wonder if there's a way to keep this logic under the covers in `compat`.

pykafka/protocol.py
sontek: This is actually required for cross runtime compatibility. Python3, 2.7, 2.6 and puppy will have different results without it check the "note" section in the crc32 docs for a better description of the issue

tests/pykafka/test_balancedconsumer.py
```diff
@@ -128,10 +134,14 @@ def test_consume(self):
                 consumer_a._partitions | consumer_b._partitions,
                 set(self.client.topics[self.topic_name].partitions.values())
             )
+        except Exception as e:
+            print(e)
+            raise
+            print(hi)
```
emmett9001: what's this?

pykafka/utils/compression.py
sontek: I thought about providing this inside the buffer compat but it's only a bug for snappy so it seemed most relevant here

tests/pykafka/test_balancedconsumer.py
sontek: Leftover debug cruft, I'll fix :)

emmett9001:
@sontek This looks great to me. I've left a few comments, but generally I think this work is really sorely needed and you've done a nice thorough job implementing it. Thanks!

I've managed to get all of the tests passing under 2.7, 3.4 and pypy on my machine. I think I'll need to add a note in the readme or in `CONTRIBUTING` about the preferred way to run tests - merging this will mean that all pull requests in the future will need to be tested against all three interpreters. The same goes for `testinstances`, which I see you've also taken into account.

@kbourgoin take a look at this if you'd like. I think this is at worst neutral for users who don't care about versions other than 2.7, and it actively helps out other users. Importantly, it doesn't make anyone's user experience worse.

pykafka/protocol.py
emmett9001: Wow, that's a strange interface. I suppose there must be a good reason for it, though.

pykafka/utils/compression.py
emmett9001: Yes, I think that makes sense.

yungchin:
>
> @yungchin Yeah, the problem is backwards compatibility. All these
> functions are exposed and probably used in clients already and they already
> allow strings... so we should probably keep the compatibility and announce
> it will be going away
>
I don't think I share the compatibility concern: we previously didn't
support py3, so these methods were only ever promised to handle py2 `str`s,
ie `bytes`, and we wouldn't be breaking any promise if we explicitly
changed that to mean `bytes`. A user who's porting to py3 will expect to
have to do some work when they were previously passing in `str`, so they
won't be surprised to have to swap to `bytes` (and that is presuming anyone
was calling protocol methods directly at all).

In my opinion, pykafka shouldn't be concerned with what kind of payloads it
is moving around. The kafka protocol treats messages as opaques. Pretty
much the only API point where it would make sense to me to optionally
accept unicode is `Producer.produce()`.

If we convert protocol methods and the like to optionally accept unicode
(which if they're used for their intended purposes, they will never get to
see any of), I'm afraid I don't see much upside to that, and a lot of extra
`if hasattr(value, 'encode')` checks going on inside potentially tight
loops. More importantly though, sticking unicode handling under the covers
just makes me worry that we'd need to do a lot more testing, to make sure
we still handle binary payloads correctly.

I do very much apologise for showing up nagging about stuff at such a late
stage in this pullreq cycle, I should have stayed in the loop on this and
started whining sooner. I can see and very much appreciate how much careful
work already went into converting all those methods to swing both ways.

pykafka/balancedconsumer.py
```diff
@@ -301,12 +302,14 @@ def _decide_partitions(self, participants):
         :type participants: Iterable of str
         """
         # Freeze and sort partitions so we always have the same results
-        p_to_str = lambda p: '-'.join([p.topic.name, str(p.leader.id), str(p.id)])
+        p_to_str = lambda p: b'-'.join([
+            get_bytes(p.topic.name), bytes(p.leader.id), bytes(p.id)]
```
yungchin: This might be more readable if the lambda just returns a plain tuple, without any bytes or str coercions.

pykafka/utils/compat.py
```diff
@@ -1,16 +1,71 @@
...
+        return iter(d.items(**kw))
+
+    def itervalues(d, **kw):
+        return iter(d.values(**kw))
+
+    def iterkeys(d, **kw):
+        return iter(d.keys(**kw))
+
+    def buffer(val):
+        return bytes(memoryview(val))
```
yungchin: I believe this combination will always force a copy operation. Do we have lots of places where a plain `memoryview()` would fail?

pykafka/utils/compat.py
```diff
@@ -1,16 +1,71 @@
...
 else:
+    range = xrange
     from threading import Condition, Lock
     # could use monotonic.monotonic() backport as well here...
     from time import time as _time
+    from Queue import Queue, Empty
+    from StringIO import StringIO
+
+    def iteritems(d, **kw):
+        return iter(d.iteritems(**kw))
```
yungchin: Not super-important at all, but since we call this in lots of places: I think the wrapping `iter()` should be superfluous.

pykafka/utils/compression.py
```diff
@@ -113,7 +120,7 @@ def decode_snappy(buff):
         raise ImportError("Please install python-snappy")
     if _detect_xerial_stream(buff):
         out = StringIO()
-        body = buffer(buff[16:])
+        body = bytes(buffer(buff[16:]))
```
yungchin: Re unicode/bytestring acceptance, I think this is a case in point - I'd be in favour of fixing the test instead, so that it sends a bytearray, or bytes, or both.

pykafka/utils/struct_helpers.py
```diff
@@ -68,6 +70,7 @@ def _unpack(fmt, buff, offset, count=1):
     """
     items = []
     array_fmt = None
+    buff = get_bytes(buff)
```
yungchin: Same thing here, and in L49 too: if the tests expect this to accept unicode, it's the tests that need fixing.

tests/pykafka/test_balancedconsumer.py
```diff
@@ -56,18 +55,18 @@ def test_consume_graceful_stop(self):
...
         """Test partition assignment for a number of partitions/consumers."""
         # 100 test iterations
-        for i in xrange(100):
+        for i in range(100):
             # Set up partitions, cluster, etc
             num_participants = i + 1
             num_partitions = 100 - i
-            participants = ['test-debian:{p}'.format(p=p)
-                            for p in xrange(num_participants)]
+            participants = sorted(['test-debian:{p}'.format(p=p)
```
yungchin: I'm curious why this test didn't need this to be sorted before? I don't really understand the implications right now.

tests/pykafka/test_producer.py
```diff
@@ -36,7 +35,7 @@ def test_produce(self):
 
         # set a timeout so we don't wait forever if we break producer code
         message = self.consumer.consume()
-        self.assertTrue(message.value == payload)
+        assert message.value == payload
```
yungchin: Is this also a py3 thing?

tests/pykafka/test_producer.py
```diff
@@ -96,11 +95,11 @@ def test_async_produce_thread_exception(self):
     def test_async_produce_unicode(self):
         """Ensure that the producer can handle unicode strings"""
         topic = self.client.topics[self.topic_name]
-        payload = u"tester"
+        payload = b"tester"
```
yungchin: This changes the nature of the test (which wants to test with unicode specifically).

(Also, it should probably send some non-ascii chars to really test that.)

pykafka/test/utils.py
```diff
@@ -2,6 +2,16 @@
 
 from pykafka.test.kafka_instance import KafkaInstance, KafkaConnection
 
+try:
+    import unittest2 as unittest
```
yungchin: This may not even be needed, given that we only support python 2.7 and up. Would save us a pip install if we just used the standard lib.

yungchin:
Ok, have finally made it to the end of that massive diff. Major kudos for all the work.

Running through it in more detail, I realised I was very ignorant when I said only `produce()` should accept unicode: it would make a lot of sense of course to also let users specify topic names and consumer groups as unicode, and not slam them for forgetting a `b` all the time. But still only in the consumer and producer interfaces, imho. If those classes make sure the names are valid for kafka and would convert them to `bytes` before saving them on `self`, we could avoid having all the other coercions while packing message sets, I'd hope.

Thanks!!

tests/pykafka/test_balancedconsumer.py
```diff
@@ -78,28 +77,31 @@ def test_decide_partitions(self):
...
 
-class BalancedConsumerIntegrationTests(unittest2.TestCase):
+class BalancedConsumerIntegrationTests(unittest.TestCase):
     maxDiff = None
 
     @classmethod
     def setUpClass(cls):
         cls.kafka = get_cluster()
-        cls.topic_name = 'test-data'
+        cls.topic_name = b'test-data'
```
yungchin: Right, so re "topic names as unicode": these tests should probably be required to work without this `b` here.

pykafka/utils/compat.py
yungchin: (Same thing probably for the next two methods.)

pykafka/test/utils.py
sontek: Yeah, I realized that afterwards.  Originally I was going to try to make everything in py26 work since it was in tox but decided that the support wasn't important to me :P

tests/pykafka/test_producer.py
sontek: This was me cleaning up the tests so that they make sense in failures.  The error message looked like this before:

```
False is not true
```

Instead of:

```
b'test-data' != 'test-data'
```

sontek:
@yungchin I 100% agree with you, I didn't want to support it everywhere but though backwards compatibility within all the APIs would be important but I definitely prefer only doing it at the producer/consumer level so I'll go through and change that up.

sontek:
There is actually one backwards incompatibility I introduced in this PR which is the CRC32 check, the original code wasn't doing the bitshifting which is necessary to work across all the different runtimes:

https://github.com/Parsely/pykafka/pull/231/files#diff-d239c7fe52131e016d3cf2f1ef348427R207

This also makes it so we are getting a long back and not an int so the struct unpacking is assuming long now.

tests/pykafka/test_producer.py
yungchin: Ah thanks, got it. Yeah that should probably have been `self.assertEqual` from the start, but `assert` is definitely an improvement already.

pykafka/protocol.py
```diff
@@ -191,16 +192,20 @@ def pack_into(self, buff, offset):
...
+        struct.pack_into('!I', buff, offset, crc)
```
yungchin: Ah thanks for the pointer back here. I missed the change from `!i` to `!I` (ie signed vs unsigned?) when I was going over this the first time. I really don't understand the finer points here, but I suppose if it still works it means kafka is happy to accept the messages so it's probably all good.

Just to be sure though, let's ping the author :)
@kbourgoin

sontek: I think this would only not work during the upgrade.  i.e if you have data produced from the old client I don't believe the new client would be able to process it properly because the data types are different

yungchin: Looking at [the protocol spec][0], it defines this field as a signed int32. So I get the feeling that reverting this to `!i` would be safer?

[0]: https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol

(Btw, not really in relation to the project in this pullreq, but when I `git grep` for "crc32" I only find this line here - does that mean we never check the crc when unpacking messages for `consume()`?)

sontek: I changed it because the tests were larger than a signed int, you can see it doing something like:


```
>>> from binascii import crc32
>>> crc32(b'hello') & 0xffffffff
907060870
>>> 2147483647 > crc32(b'hello bar') & 0xffffffff
False
```

but even we were able to change it back we should have the backwards compatibility issue because the previous code wasn't doing the bit shifting so we'd have different results between python2 and python3:

Python2:

```
>>> crc32(b'hello bar')
-1737004441
>>> crc32(b'hello bar') & 0xffffffff
2557962855
```

Python3:

```
>>> from binascii import crc32
>>> crc32(b'hello bar')
2557962855
>>> crc32(b'hello bar') & 0xffffffff
2557962855
```

yungchin: Ah, now I get how it works. Thanks for explaining all this so patiently.

Yeah, so with those examples I understand that packing it into the struct as if it were an unsigned field is all fine - it's still the same bit pattern. Thanks!

emmett9001:
@sontek Looks like there are some more changes you'd like to make here? I'm interested in getting this merged as soon as possible, since it enables bugfixes on master to benefit from the testing setup and py3 support. The `i`/`I` backwards incompatibility is ok in my opinion, since folks who are upgrading can easily start consuming from the latest offset and skip all of the old incompatible data.

sontek:
@emmett9001 I'd like to move it to note doing as many `get_bytes` calls and force bytes to be passed in to everything outside of producer/consumer but I probably wont have time to do it until Thursday (or this weekend).   I'm ok with it being merged as-is or if you want to wait I'll get to it eventually

emmett9001:
Closing this in favor of #246. With that pull request, @yungchin and I will be able to make progress on some of the last changes we've talked about here. @sontek, if you have more changes you'd like to make, you can open a new pull request against the [sontek_port_to_pyhon34](https://github.com/Parsely/pykafka/tree/sontek_port_to_python34) branch.

tests/pykafka/test_balancedconsumer.py
emmett9001: Since the `_decide_partitions` function used `.sort()` instead of `sorted`, it was modifying the test function's copy of the `participants` array. This made the `idx` assignment always match its value in `_decide_partitions`, which could potentially have masked some errors. This way is actually better.

pykafka/test/utils.py
emmett9001: Fixed in #246

pykafka/utils/compat.py
emmett9001: Fixed in #246

PR #232, opened 2015-08-21
Fix _partition_messages in producer

ryananguiano:
_partition_messages(self, messages) was running the partitioner on the entire message instead of the key. Partitioners were not returning consistent partitions with the same key.

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -198,7 +198,7 @@ class Producer():
             else:
                 key, value = message
             value = str(value)
-            yield (key, value), self._partitioner(partitions, message).id
+            yield (key, value), self._partitioner(partitions, key).id
 
     def _produce(self, message_partition_tups, attempt):
         """Publish a set of messages to relevant brokers.


yungchin:
Ouch, thanks, that's a great catch. I think I've read that line more than a few times and never noticed.

This also needs fixing in #177, which currently also passes `(message, partition_key)` rather than just `partition_key`. Since it looks like #177 will make for a 2.0 release, this should also separately be merged in a 1.x point release.

Would you be able to add a test case too?

emmett9001:
I'm not sure this fix works - the partition key [can't ever be `None`](https://github.com/Parsely/pykafka/blob/master/pykafka/partitioners.py#L74) if we want to stay compatible with the `HashingPartitioner`. Something like passing `key or value` would probably do the trick.

yungchin:
I would be ok with `HashingPartitioner` throwing exceptions in that situation: if you're going to send messages without a key, you shouldn't be configuring a non-default partitioner.

ryananguiano:
Sure. I might have time for the test case this weekend. And since the hashing_partitioner must be set explicitly, I also think it should raise an exception if you don't pass a key.

yungchin:
Awesome, thanks!

Issue #224, opened 2015-07-29
SocketDisconnectedError on commit_offsets()


youngpm:
Related to #209, the code snippet below gives me a `SocketDisconnectedError` using pykafka 1.1.0 (the version in pypi) and Kafka 0.8.2.0 (I just spun up Kafka locally as described in their quickstart).

I start the consumer, put a message on the queue, and wait 30 minutes before committing the offset.

```python
#!/usr/bin/env python
import time

from pykafka import KafkaClient
from pykafka.common import OffsetType

if __name__ == "__main__":

    time_to_sleep = 0.5*60*60

    # Get a consumer.
    client = KafkaClient(hosts='localhost:9092')
    input_topic = client.topics['test']
    consumer = input_topic.get_simple_consumer(consumer_group='testgroup',
                                               auto_offset_reset=OffsetType.EARLIEST,
                                               auto_commit_enable=False)

    # List messages on topic.
    for message in consumer:
        print('{} {}'.format(message.value, message.offset))
        time.sleep(time_to_sleep)
        print('committing...')
        consumer.commit_offsets()
        print('committed!')
```

The traceback looks like:

```
In [2]: run consumer_bug.py
message16 15
committing...
---------------------------------------------------------------------------
SocketDisconnectedError                   Traceback (most recent call last)
/mnt/panasas/pyoung/pykafka_bug/consumer_bug.py in <module>()
     22         time.sleep(time_to_sleep)
     23         print('committing...')
---> 24         consumer.commit_offsets()
     25         print('committed!')
     26

/mnt/panasas/pyoung/pykafka_bug/venv/lib/python2.7/site-packages/pykafka/simpleconsumer.pyc in commit_offsets(self)
    347
    348             response = self._offset_manager.commit_consumer_group_offsets(
--> 349                 self._consumer_group, 1, 'pykafka', reqs)
    350             parts_by_error = handle_partition_responses(
    351                 self._default_error_handlers,

/mnt/panasas/pyoung/pykafka_bug/venv/lib/python2.7/site-packages/pykafka/broker.pyc in commit_consumer_group_offsets(self, consumer_group, consumer_group_generation_id, consumer_id, preqs)
    306                                   consumer_id,
    307                                   partition_requests=preqs)
--> 308         return self._offsets_channel_req_handler.request(req).get(OffsetCommitResponse)
    309
    310     def fetch_consumer_group_offsets(self, consumer_group, preqs):
[...]

Issue #223, opened 2015-07-29
Consumer hanging after a broker gets down


AgustinCB:
Trying the library, I set the following situation:

Five brokers in the kafka cluster, one producer and two consumers over one topic. 

If I stop one of the brokers, the producer get some warning messages and after some seconds, restarts its normal behavior with no problem. The consumer, on the other side, fails with a Socket Disconnect exception on a different thread, leaving the process alive, but not doing anything.

Using what I saw on the producer code, I modified the consumer to use a similar approach:

https://github.com/appsoma/pykafka/pull/2

However, I'm not sure if there were a reason to the behavior of the consumer, if my code breaks something else or if it doesn't fit the code philosophy that you're following.

Excuse my poor English, please.

fortime:
I have encoutered this problem too. I think it is a bug of broker connection. When a Socket Disconnect exception happens, it will disconnect the broker connection, and it will never re-connect. I think it should check the connection to broker is still alive. If not, re-connect it.

emmett9001:
Current master now immediately raises an exception if a broker goes down, which at least allows the client code to catch the exception and reinstantiate the consumer. A more complete fix will involve smartly retrying fetch requests after updating broker metadata. This effort will be tracked in this ticket.

PR #235, opened 2015-08-21
[WIP] Raise worker exceptions

emmett9001:
This pull request uses the same method in use in `Producer` to surface exceptions encountered on worker threads to the main thread. This can potentially help with the issue seen in #224, #223, and #189.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -20,8 +20,10 @@ limitations under the License.
 __all__ = ["SimpleConsumer"]
 import itertools
 import logging
+import sys
 import time
 import threading
+import traceback
 from collections import defaultdict
 
 from .common import OffsetType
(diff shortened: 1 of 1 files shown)


ltucker:
@emmett9001 when you say it can help with the listed consumer issues, is the idea that an application would catch SocketDisconnected (and whatever else) and establish an new Consumer instance when there is a failure on any individual broker connection?

emmett9001:
The basic idea behind this pull request is that client code could catch these exceptions and instantiate a new consumer. However, this pull request is still very experimental. I haven't given much thought to what it actually looks like from the client perspective. I certainly wouldn't want clients to have to care about the internal connection objects.

rduplain:
Current status: need to replicate error and confirm that this fixes it.

rduplain:
This work is on-going (I had an incorrect comment there briefly, which is now deleted).

PR #236, opened 2015-08-26
Ports are ints.

aeroevan:
On a client I was having issues with getaddrinfo() argument 2 must be an integer or string when using unicode string for the connection information (yay python 2...).

This seems to fix that case.


emmett9001:
Thanks! This looks fine to me - just an oversight on our part.

Issue #218, opened 2015-07-27
Starting > 1 consumer hangs python process


Fluxx:
Using `pykafka==1.1.0`, running against a single node Kafka `0.8.2.1` server, the following script hang and never exits:

```python
from pykafka import KafkaClient


client = KafkaClient('kafka:9092')

consumer1 = client.topics['test1'].get_simple_consumer()
consumer2 = client.topics['test2'].get_simple_consumer()

print 'Consumers made'
```

Running I see this:

```bash
$ python application consumertest.py
Consumers made
```

And the process seems to hang forever.  I have waited as long as 5 minutes and it never exits.

For some more evidence, if I change the consumer constructor to `get_simple_consumer(auto_start=False)` the script exits, however, if I then add this code below...

```python
consumer1.start()
consumer2.start()
```

...the process hangs again, leaving me to think somewhere in the `consumer.start()` process is where the issue is.  Finally, when hanging, if I `SIGTERM` Kafka the Python process exits without any exception being thrown.

I'm about to start digging into the `start()` code to see if I can track down the issue myself, but I wanted to post this here in case this is a known issue (I couldn't find any) or if you guys had a better idea what might be going on here?

Fluxx:
Hmmm....more info.  If I change the [sleep in the fetch worker](https://github.com/Parsely/pykafka/blob/1.1.0/pykafka/simpleconsumer.py#L278) to `0.1` Python exits just fine.  Also if comment out [the `fetch_lock.release()`](https://github.com/Parsely/pykafka/blob/1.1.0/pykafka/simpleconsumer.py#L607) as part of fetching new messages, Python also exits too.  Not 100% sure what this means, but I'll keep digging.

emmett9001:
A simple workaround for this is calling `consumer1.stop(); consumer2.stop()` when you'd like your program to exit. This ensures clean shutdown of the worker threads used in message fetching.

Fluxx:
@emmett9001 thanks for the quick reply and suggestion, that certainly works.  I'm stumped as to why this is happening in the first place though as the threads started all all daemon threads so they shouldn't hang like this...so weird.

emmett9001:
My thoughts exactly. This will require some digging.

Fluxx:
Sounds good.  Happy to help debug if needed.

emmett9001:
Any further information you can provide on possible causes is really appreciated.

emmett9001:
This bug still exists in `2.0.0`

[...]

PR #237, opened 2015-08-28
properly lock partitions

emmett9001:
Lock partitions rather than `OwnedPartition`s. Related to #218.

### DIFF
--- a/pykafka/partition.py
+++ b/pykafka/partition.py
@@ -51,6 +51,7 @@ class Partition():
         self._replicas = replicas
         self._isr = isr
         self._topic = topic
+        self.fetch_lock = self.topic._cluster.handler.Lock()
 
     def __repr__(self):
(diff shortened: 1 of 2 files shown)


emmett9001:
@yungchin @kbourgoin let me know what you think of this fix. I can't think of anything it breaks or makes worse for the average user, but I feel like I may be missing something. My thinking is that if we care to lock each partition in a single-consumer context, we should probably also be locking each partition for N consumers.

yungchin:
I don't really understand why this fix works. As I understood it previously, the fetch lock was to protect the message queue and the stored offsets, both of which are in `OwnedPartition`. Should there be additional state in `Partition` that needs guarding?

I'm half worried that this fix might in fact disguise a bug somewhere else, in the same way that that longer `sleep` from #218 hides it (I suppose that works because if the fetcher thread sleeps most of the time, that decreases the likelihood of concurrent requests?).

PR #245, opened 2015-09-01
Make auto_offset_reset now work as expected for new consumer groups

### LINKED ISSUE #241: auto_offset_reset=-1 not respected for new consumer groups
I noticed that new consumer groups don't start consuming from the end of a topic even if ```auto_offset_reset=-1``` (the default), unless I also set ```reset_offset_on_start=True```.  However, if I set ```reset_offset_on_start=True```, existent consumer groups will always consume from the end of the topic, even if they have committed offsets.

I tested 4 different combinations of consumer groups and ```reset_offset_on_start```.  For all cases below, ```auto_offset_reset=-1``` (the default).  Here's what I found:

- New consumer group, ```reset_offset_on_start=False```: starts consuming at earliest offset.
- New consumer group, ```reset_offset_on_start=True```: starts consuming at latest offset.
- Existent consumer group, ```reset_offset_on_start=False```: starts consuming at committed offset.
- Existent consumer group, ```reset_offset_on_start=True```: starts consuming at latest offset.

I'm running Kafka 0.8.2.1 and pykafka 2.0.0.

This is a pretty annoying bug, as it means there is no simple way to have a consumer that starts consuming from latest offset UNLESS there is a committed offset, which I believe is the intended behavior of ```auto_offset_reset=-1```.

What should happen with ```auto_offset_reset=-1``` and  ```reset_offset_on_start=False```:

- New consumer group, ```reset_offset_on_start=False```: starts consuming at **latest** offset.
- Existent consumer group, ```reset_offset_on_start=False```: starts consuming at committed offset.

Thanks!

ottomata:
Fixes #241


rduplain:
@ottomata Thanks for this!
@emmett9001 The build is hanging on:

```
test_consume (tests.pykafka.test_balancedconsumer.BalancedConsumerIntegrationTests) ...
```

Travis CI is killing it after hanging for 10 minutes. Any ideas?

ottomata:
Oh, I bet you that since ```auto_offset_reset=-1``` works now, the test is now waiting for a new message to be produced, since it is trying to consume from the end of a topic.

yungchin:
It occasionally hangs on that test on my machine too - it's a quirk presumably inside BalancedConsumer that we've not figured out so far. I've restarted the build.

yungchin:
Hmmm, seems stuck again. Maybe it's not that quirk, and @ottomata's explanation is right - did you run the tests locally?

ottomata:
I was able to get tests to run by setting ```auto_offset_reset=-2``` for the BalancedConsumers created by BalancedConsumerIntegrationTests.  I will submit a patch to do this, and I will try to add a new test that also tests ```auto_offset_reset=-1```

emmett9001:
@ottomata mind merging current master into this branch before continuing to test? I just updated it to use `tox`, which allows us to make sure bugfixes like this remain compatible with py34 and pypy.

ottomata:
@emmett9001 and I discussed this in IRC and agreed on a better solution.  I'll submit that in a new pull request.

PR #246, opened 2015-09-01
[sontek] Port to python 3.4

emmett9001:
This is an `origin` version of #231 that allows us to make some further changes before merging.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ dist/
 .*.swp
 .*.un~
 *.bak
+.coverage
+coverage.xml
+*#*#*
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ notifications:
     - emmett@parsely.com
 
 before_install:
-    - sudo apt-get install -qq libev-dev libsnappy-dev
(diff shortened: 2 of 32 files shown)


yungchin:
@sontek 
For now, I've reverted most uses of `get_bytes` on this branch, I hope that's ok. As we discussed, the protocol methods were not going to have unicode handling, but on second thought it also seemed potentially messy to handle unicode on `produce()` et al: given that you cannot annotate the encoding of your payloads in the message format (and so upon `consume()` users would have to know whether to decode), it might be good to make that explicit to users: "what you're using here is just a binary transport".

That's not to say that unicode handling will not be in the cards ever. But sticking to py2 `str`s keeps this pullreq closer to a minimal/sufficient version of py3 support, and will leave us some time and optionality to discuss unicode handling in a new pullreq where we could think about the interface a bit more, if we still want it.

Thanks again for the massive job on this, and for working out all the moving parts of setup/testing. I almost fried my CPU with tox runs tonight, as a small taster of how much effort this branch has been :)

sontek:
@yungchin Nice, that is looking pretty nice!

PR #247, opened 2015-09-02
Make auto_offset_reset=-1 consume from latest with new consumer group

### LINKED ISSUE #241: auto_offset_reset=-1 not respected for new consumer groups

ottomata:
Fixes #241 so that auto_offset_reset=-1 consumes from latest available offset when the consumer group has no committed offsets.

This changes the default auto_offset_reset to -2 (earliest), so as not to break backwards
compatibility.  Hopefully this will be changed back to -1 in some future major version.

### DIFF
--- a/tests/pykafka/test_balancedconsumer.py
+++ b/tests/pykafka/test_balancedconsumer.py
@@ -153,7 +153,7 @@ class BalancedConsumerIntegrationTests(unittest2.TestCase):
 
             # Since we are consuming from the latest offset,
             # produce more messages to consume.
-            for i in range(1000):
+            for i in range(2):
                 self.prod.produce('msg {num}'.format(num=i).encode())
 
             # Consume from both a few times
--- a/tests/pykafka/test_balancedconsumer.py
+++ b/tests/pykafka/test_balancedconsumer.py
@@ -4,10 +4,9 @@ import time
 import unittest2
 
 from pykafka import KafkaClient
-from pykafka.balancedconsumer import BalancedConsumer
+from pykafka.balancedconsumer import BalancedConsumer, OffsetType
 from pykafka.test.utils import get_cluster, stop_cluster
 from pykafka.utils.compat import range
-
 
 def buildMockConsumer(num_partitions=10, num_participants=1, timeout=2000):
     consumer_group = 'testgroup'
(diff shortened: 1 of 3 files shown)


tests/pykafka/test_balancedconsumer.py
```diff
@@ -139,6 +140,43 @@ def test_consume(self):
...
+                auto_offset_reset=OffsetType.LATEST
+            )
+            consumer_b = self.client.topics[self.topic_name].get_balanced_consumer(
+                b'test_consume_latest', zookeeper_connect=self.kafka.zookeeper,
+                auto_offset_reset=OffsetType.LATEST
+            )
+
+            # Since we are consuming from the latest offset,
+            # produce more messages to consume.
+            for i in range(1000):
```
emmett9001: No need to produce so much here - 2 should be sufficient.

tests/pykafka/test_balancedconsumer.py
```diff
@@ -4,11 +4,10 @@
 import unittest2
```
emmett9001: These tests should actually go in `test_simpleconsumer`, since the change was made there.

ottomata: ```test_simpleconsumer``` already has a test for EARLIEST and LATEST offsets.  Just thought I'd add one for balanced consumer as well.

PR #249, opened 2015-09-03
prevents islice error on floating point numbers when using pypy

sammerry:
Both start and end integers are floating point when using pypy. This forces it to drop the decimal place so islice doesn't error.


yungchin:
Hi, thanks for this. Even though your pullreq seems to be based against current master, could it be that you're addressing a bug that you encountered in an earlier version?

I believe that the changes that came in with 4dc9be3e1b010708b75d40c11e0dea1162210ab7 should  ensure that both these vars are `int` already.

yungchin:
@sammerry is this still an issue for you? I'm going to close it, believing that it's fixed on master, but feel free to reopen.

PR #254, opened 2015-09-03
handle broker failure in producer

### LINKED ISSUE #250: producer should be resilient to broker failure
When a broker in a multi-node cluster with replicated topics goes down, the producer should not crash. Instead, it should continue producing messages to all available brokers. The producer should also be able to start when at least `<replication factor>` brokers in the cluster are up.

emmett9001:
This pull request resolves #250 by making the producer resilient to certain broker failure modes. In particular, this pull request makes the producer able to handle the following cases.

Given a multi-node cluster with a topic with replication factor `N`:
* When `< N - 1` brokers become unavailable, the producer seamlessly updates its metadata and produces messages to the updated partition leaders.
* When `>= N - 1` brokers become unavailable, some partitions may not have enough available replicas, and thus will have no leader. In this case, the best a producer can do is to continue retrying messages destined for that partition until it becomes available. This pull request implements this behavior.
* When a broker in the connect string supplied to `Producer.__init__` is unavailable, the producer crashes with an error.

Non-producer changes (exception resilience):
* `Broker.request_metadata` continues its retry loop if it encounters `SocketDisconnected`, which can happen if a broker disconnects during the loop
* `Topic.update` and `Partition.update` raise `LeaderNotAvailable` when they encounter `KeyError`. This error happens when the metadata returned from the cluster indicates a broker that is not held in the `Cluster`'s registry, a common failure mode when brokers disconnect suddenly. `Cluster.update` catches `LeaderNotAvailable` and warns about it.

Producer changes:
* Add `running` to `OwnedBroker` to allow its thread to be stopped by its parent `Producer`. Stop worker threads when the `Producer` shuts down. This is not strictly related to handling broker failures.
* `NotLeaderForPartition` now causes the producer to `_update`, which refreshes cluster metadata and reinstantiates the `OwnedBroker`s.
* `SocketDisconnectedError` also causes the producer to `_update`.
* If the maximum number of retries is reached in `_send_request`, the messages are re-produced, placing them into the appropriate queues for their partition keys. This means that a partition with no available leader will continue to have its messages retried until the leader becomes available.

### DIFF
--- a/pykafka/cluster.py
+++ b/pykafka/cluster.py
@@ -26,7 +26,8 @@ import weakref
 from .broker import Broker
 from .exceptions import (ConsumerCoordinatorNotAvailable,
                          KafkaException,
-                         UnknownTopicOrPartition)
+                         UnknownTopicOrPartition,
+                         LeaderNotAvailable)
 from .protocol import ConsumerMetadataRequest, ConsumerMetadataResponse
 from .topic import Topic
 from .utils.compat import iteritems, range
@@ -287,4 +288,8 @@ class Cluster(object):
                         'will NOT work. You need to create at least one topic '
                         'manually using the Kafka CLI tools.')
         self._update_brokers(metadata.brokers)
-        self._update_topics(metadata.topics)
+        try:
+            self._update_topics(metadata.topics)
+        except LeaderNotAvailable:
+            log.warning("LeaderNotAvailable encountered. This is "
+                        "because one or more partitions have no available replicas.")
--- a/pykafka/cluster.py
+++ b/pykafka/cluster.py
@@ -277,19 +277,23 @@ class Cluster(object):
 
     def update(self):
         """Update known brokers and topics."""
-        metadata = self._get_metadata()
-        if len(metadata.brokers) == 0 and len(metadata.topics) == 0:
-            log.warning('No broker metadata found. If this is a fresh cluster, '
-                        'this may be due to a bug in Kafka. You can force '
-                        'broker metadata to be returned by manually creating '
-                        'a topic in the cluster. See '
-                        'https://issues.apache.org/jira/browse/KAFKA-2154 '
-                        'for information. Please note: topic auto-creation '
-                        'will NOT work. You need to create at least one topic '
-                        'manually using the Kafka CLI tools.')
-        self._update_brokers(metadata.brokers)
-        try:
-            self._update_topics(metadata.topics)
-        except LeaderNotAvailable:
-            log.warning("LeaderNotAvailable encountered. This is "
-                        "because one or more partitions have no available replicas.")
+        max_retries = 3
+        for i in xrange(max_retries):
+            metadata = self._get_metadata()
+            if len(metadata.brokers) == 0 and len(metadata.topics) == 0:
+                log.warning('No broker metadata found. If this is a fresh cluster, '
+                            'this may be due to a bug in Kafka. You can force '
+                            'broker metadata to be returned by manually creating '
+                            'a topic in the cluster. See '
+                            'https://issues.apache.org/jira/browse/KAFKA-2154 '
+                            'for information. Please note: topic auto-creation '
+                            'will NOT work. You need to create at least one topic '
+                            'manually using the Kafka CLI tools.')
+            self._update_brokers(metadata.brokers)
+            try:
+                self._update_topics(metadata.topics)
+            except LeaderNotAvailable:
+                log.warning("LeaderNotAvailable encountered. This is "
+                            "because one or more partitions have no available replicas.")
+            else:
+                break
--- a/pykafka/partition.py
+++ b/pykafka/partition.py
@@ -131,7 +131,7 @@ class Partition():
             # Check leader
             if metadata.leader != self._leader.id:
                 log.info('Updating leader for %s from broker %s to broker %s', self,
-                          self._leader.id, metadata.leader)
+                         self._leader.id, metadata.leader)
                 self._leader = brokers[metadata.leader]
             # Check Replicas
             if sorted(r.id for r in self.replicas) != sorted(metadata.replicas):
@@ -142,6 +142,6 @@ class Partition():
                 log.info('Updating in sync replicas list for %s', self)
                 self._isr = [brokers[b] for b in metadata.isr]
         except KeyError:
-            raise LeaderNotAvailable("Leader for partition %s not available. This is "
+            raise LeaderNotAvailable("Replica for partition %s not available. This is "
                                      "probably because none of its replicas are "
                                      "available.", self.id)
--- a/pykafka/partition.py
+++ b/pykafka/partition.py
@@ -20,6 +20,7 @@ __all__ = ["Partition"]
 import logging
 
 from .common import OffsetType
+from .exceptions import LeaderNotAvailable
 from .protocol import PartitionOffsetRequest
 
 log = logging.getLogger(__name__)
@@ -130,7 +131,7 @@ class Partition():
             # Check leader
             if metadata.leader != self._leader.id:
                 log.info('Updating leader for %s from broker %s to broker %s', self,
-                         self._leader.id, metadata.leader)
+                          self._leader.id, metadata.leader)
                 self._leader = brokers[metadata.leader]
             # Check Replicas
             if sorted(r.id for r in self.replicas) != sorted(metadata.replicas):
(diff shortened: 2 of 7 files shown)


emmett9001:
@kbourgoin let me know how these changes look to you. I've tested all of the cases I can think of with three local brokers running a topic with RF=2. It's not going to be truly possible to know that these will fix the issue we were seeing in production, but we can at least get it to behave properly in all of our local testing before trying it in the real world.

pykafka/cluster.py
```diff
@@ -287,4 +288,8 @@ def update(self):
                         'will NOT work. You need to create at least one topic '
                         'manually using the Kafka CLI tools.')
         self._update_brokers(metadata.brokers)
-        self._update_topics(metadata.topics)
+        try:
+            self._update_topics(metadata.topics)
+        except LeaderNotAvailable:
```
yungchin: If we're going to discard the exception here, there might be a case for just discarding it directly in `Partition.update`, and logging it from there: that way, an exception on one Partition wouldn't stop us from updating remaining Topics/Partitions, and the leaderless Partition may not be one that this client cares about.

pykafka/partition.py
```diff
@@ -140,4 +142,6 @@ def update(self, brokers, metadata):
                 log.info('Updating in sync replicas list for %s', self)
                 self._isr = [brokers[b] for b in metadata.isr]
         except KeyError:
-            raise Exception("TODO: Type this exception")
+            raise LeaderNotAvailable("Leader for partition %s not available. This is "
```
yungchin: Apart from L135, couldn't the `KeyError` also come from L143, in which case it's not a missing leader but actually another replica?

pykafka/cluster.py
yungchin: Actually alternatively, shouldn't we retry `Cluster.update` altogether, when this is caught? I mean, we just got sent `metadata` which seems to already be out of date while we're trying to use it.

pykafka/producer.py
```diff
@@ -317,30 +344,12 @@ def _get_partition_msgs(partition_id, req):
...
-            owned_broker.enqueue(new_queue_contents[owned_broker.broker.id],
-                                 self._block_on_queue_full)
-            owned_broker.resolve_event_state()
-            owned_broker.lock.release()
+                log.error("Failed to produce messages to broker %s:%s after %s attempts. "
+                          "Re-enqueuing %s messages.", owned_broker.broker.host,
+                          owned_broker.broker.port, attempt, len(to_retry))
+                owned_broker.increment_messages_pending(-1 * len(to_retry))
+                for tup in to_retry:
+                    self._produce(tup)
```
yungchin: Doesn't this break the `max_retries` setting? I would have thought `ProduceFailureError` is fine here.

pykafka/producer.py
```diff
@@ -395,9 +405,26 @@ def queue_reader():
                     # surface all exceptions to the main thread
                     self.producer._worker_exception = sys.exc_info()
                     break
+            try:
```
yungchin: I think I prefer the way you did this before, through `Producer._update_leaders`. Here, there's no control over when exactly these messages make it into `_waiting_messages` (they might wait for the lock for a while?), so chances are they only make it into there after `Producer._update` finishes - in which case they could be there a long time.

pykafka/producer.py
emmett9001: The issue here is that this `else` case is reached whenever a broker dies during message production. Raising an exception here causes the `Producer` to raise that exception to the main thread.

yungchin: I would argue that if the user didn't want that, they should have set a higher `max_retries` or a longer `retry_backoff_ms`, so as to bridge the time it takes the cluster to elect a new leader.

(I'm thinking especially of the use case where message re-ordering is a problem - so a user would want to be able to set `max_retries=0` and get `ProduceFailureError` right away.)

yungchin:
Ok, I think I've looped through this in my head until it exploded now :)

There's one more thing that the old `Producer._update_leaders` did that hasn't made it over: because it held locks on all `OwnedPartitions`, it would block new `enqueue` calls while an update was going. That is important for message order, I think. With the new `_update`, you could potentially call `produce` in the middle of it running, and the newly submitted message would queue-jump past all the messages that were temporarily in `_waiting_messages`.

pykafka/producer.py
emmett9001: I agree. The problem is that this branch is meant to make the producer not crash when a broker fails. If we raise an exception here, the client will need to handle that exception to avoid a crash. Maybe that's ok, but it at least complicates the code a client needs to write:

```python
with Producer() as p:
    try:
        p.produce(message)
    except ThisException:
        # what goes here? p.produce(message)?
```

pykafka/producer.py
emmett9001: This way just seems cleaner to me, since it doesn't involve freezing the entire producer to do a leader reassignment.

pykafka/producer.py
yungchin: Ah right - I didn't sufficiently realise that the `ProduceFailureError` would crash the worker thread.

Sorry, yes that was silly - so I should have argued that instead of `raise ProduceFailureError` it should simply say `pass` and drop the message. That wouldn't crash the worker, and it's consistent with what the user asks, ie give up after `max_retries`.

(And then when we start returning futures from `produce()`, we can then push a `ProduceFailureError` onto the future instead, but that's a separate issue.)

pykafka/producer.py
yungchin: Yeah I agree that if it can be done without blocking `produce()` (at least for `async=True`), that would be neat. But how do you guarantee that the stopping threads aren't going to push their messages onto `_waiting_messages` too late? If that happens, I think they might sit there for ages until some other event triggers `_update`.

pykafka/producer.py
```diff
@@ -317,30 +350,15 @@ def _get_partition_msgs(partition_id, req):
...
-            owned_broker.lock.release()
+                log.error("Failed to produce messages to broker %s:%s after %s attempts. "
+                          "Re-enqueuing %s messages.", owned_broker.broker.host,
+                          owned_broker.broker.port, attempt, len(to_retry))
+                owned_broker.increment_messages_pending(-1 * len(to_retry))
+                for kv, partition_id, msg_attempt in to_retry:
+                    if msg_attempt >= self._max_message_retries:
+                        raise ProduceFailureError("Message failed to send after %d "
+                                                  "retries.", self._max_message_retries)
+                    self._produce((kv, partition_id, msg_attempt + 1))
```
yungchin: Thanks for this update, that looks perfect to re-enable the `max_retries=0` option. And seeing how this works with the attempts-tracking, it suddenly "clicked" in my head, what you said about why L347-351 don't help much if a broker has gone down or a leader has changed.

I kept misreading those lines before, thinking the fact that there's now an `_update` in L339 should fix everything, but now I get that it would still resend to the old destination, as per `owned_broker` in L351.

So given that the former retry loop doesn't really work so well, and that this does, would it make sense to drop that other one altogether? So what I mean is, you'd be left with:
```python
         if to_retry:
                owned_broker.increment_messages_pending(-1 * len(to_retry))
                for kv, partition_id, msg_attempt in to_retry:
                    if msg_attempt >= self._max_retries:
                        raise ProduceFailureError("Message failed to send after %d "
                                                  "retries.", self._max_retries)
                    self._produce((kv, partition_id, msg_attempt + 1))
```
(with possibly some copy-paste errors, and I've replace `_max_message_retries` by `_max_retries`, so you'd be back to one parameter for this).

yungchin:
(Just a note: I've struck through my earlier comment above about locks and ordering, which was on second thought, plain and simply, wrong :) )

pykafka/partition.py
yungchin: Sorry, that was sort of only half of the question I meant to put there: what I meant was, if the `KeyError` is due to a non-leader broker going missing in action, should we raise `LeaderNotAvailable` and thereby block producing to that partition, or could we swallow the exception as long as 1) there is still a leader, and maybe 2) `len(_isr) > required_acks`?

But I think that's also half off-topic for this pullreq, so we can make it a separate ticket, for if the problem comes up in practice.

emmett9001:
@yungchin thanks for the thoughts. I believe my [last commit](https://github.com/Parsely/pykafka/commit/c70a97924f8ac45aa5226c905f9ccdb6fcb1038a) addresses the ordering and dropped messages issues. Now I'm just testing a bunch to try and catch edge cases.

emmett9001:
@yungchin I'm inclined to punt on perfect ordering for the moment. Do the JVM clients guarantee ordering in this way? I know that Kafka itself guarantees that messages within a partition will be ordered, but is the client expected to do that too?

PR #256, opened 2015-09-04
fix consumer thread leak

### LINKED ISSUE #255: Thread leakages in simple consumer and producer
I have a program which creates one of my own threads when a certain call happens. That thread of mine creates a simpleconsumer.

The def run() of my thread does something like this:

```
	topic = kafka_client.topics['test_topic']
	consumer = topic.get_simple_consumer( 'group_id' )
```

Note that `kafka_client` is global -- outside of my thread. That is, there is only one of them for all threads. But, consumer is local: there's one per thread.  When my thread's run() stops cleanly it calls:

```
	consumer.stop()
```

Yet when I look at the thread count I see that it grows by one every time I start and stop one of my threads.  If I take out the call to get the consumer then it doesn't. It appears as though there is a thread leak in the simple_consumer.

emmett9001:
This pull request resolves #255 by fixing a race condition around connecting the offsets channel in the broker.

### DIFF
--- a/pykafka/broker.py
+++ b/pykafka/broker.py
@@ -89,6 +89,10 @@ class Broker():
         self._socket_timeout_ms = socket_timeout_ms
         self._offsets_channel_socket_timeout_ms = offsets_channel_socket_timeout_ms
         self._buffer_size = buffer_size
+        # when multiple consumers are started simultaneously on separate threads,
+        # they all race to connect the offsets channel at once. ensure this
+        # only happens once to avoid race condition.
+        self._offsets_channel_lock = self._handler.Lock()
         self.connect()
 
     def __repr__(self):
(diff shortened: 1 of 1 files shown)


yungchin:
I was going to comment that I didn't understand why this doesn't work correctly without locking: it would seem the worst that could happen is that you create a connection plus request handler, and then drop them in favour of another, new connection plus request handler. So there would be some wasted work, but not necessarily thread leakage.

But having had another look at `RequestHandler`, I strongly suspect that the issue is (again) with its use of `atexit`: because it runs `atexit.register(self.stop)`, it ensures (via the closure over `self`) that no instance of `RequestHandler` will ever be gc'd until program exit. So the real leak, I would argue, is in `RequestHandler`.

Now if that `atexit` is really unavoidable, then the fix here does of course make sense, but in that case we should make it explicit that `RequestHandler` instances are always non-gc'able. If it can be avoided though, then we should instead fix that and we wouldn't need this one.

PR #258, opened 2015-09-07
Get rid of zombie RequestHandler threads

### LINKED ISSUE #257: forget to disconect socket ?
recently, I met a problem about kafka connections, the phenomenon is that the TCP connections number on Kafka server was more than 20,000, and I reviewed my Producer and Consumer code, also the pyKafka code, I found a potential connection leak problem. 

Here's my test code and result:

1. build one  KafkaClient with two connections

        if __name__ == '__main__':
               client = KafkaClient(hosts)
                raw_input()

        ➜  ~  netstat | grep 9092
        tcp        0      0 10.8.192.*:51778       10.8.255.*:9092       ESTABLISHED
        tcp        0      0 10.8.192.*:51777       10.8.255.*:9092       ESTABLISHED

2. build three  KafkaClient with six connections

        if __name__ == '__main__':
            for i in xrange(3):
                client = KafkaClient(hosts)
            raw_input()

        ➜  ~  netstat | grep 9092
        tcp        0      0 10.8.192.*:58035       **********************:9092 ESTABLISHED
        tcp        0      0 10.8.192.*:58037       **********************:9092 ESTABLISHED
        tcp        0      0 10.8.192.*:58039       **********************:9092 ESTABLISHED
        tcp        0      0 10.8.192.*:58036       **********************:9092 ESTABLISHED
        tcp        0      0 10.8.192.*:58034       **********************:9092 ESTABLISHED
        tcp        0      0 10.8.192.*:58038       **********************:9092 ESTABLISHED
 
forgot disconnect after _get_metadata(where I write TODO)  ?

    def _get_metadata(self):
            """Get fresh cluster metadata from a broker."""
            # Works either on existing brokers or seed_hosts list
            brokers = [b for b in self.brokers.values() if b.connected]
            if brokers:
                for broker in brokers:
                    response = broker.request_metadata()
                    if response is not None:
                        return response
            else:  # try seed hosts
                brokers = self._seed_hosts.split(',')
                for broker_str in brokers:
                    try:
                        h, p = broker_str.split(':')
                        broker = Broker(-1, h, p, self._handler,
                                        self._socket_timeout_ms,
                                        self._offsets_channel_socket_timeout_ms,
                                        buffer_size=1024 * 1024)
                        response = broker.request_metadata()
                        # TODO: **broker.disconnect** ?
                        if response is not None:
                            return response

                    except:
                        log.exception('Unable to connect to broker %s',
                                      broker_str)
            # Couldn't connect anywhere. Raise an error.
            raise Exception('Unable to connect to a broker to fetch metadata.')

sorry about my English and how can we release client correctly.

yungchin:
This fixes #257 by making sure `RequestHandler` instances are gc'able.

Because it removes the atexit flow completely, it also indirectly deals with #218 - with the changes here, `RequestHandler.stop` will run later in the teardown process, so that the consumers are already gone, so they no longer keep pressure on the request queue.

### DIFF
--- a/pykafka/handlers.py
+++ b/pykafka/handlers.py
@@ -133,6 +133,7 @@ class RequestHandler(object):
     def _start_thread(self):
         """Run the request processor"""
         self = weakref.proxy(self)
+
         def worker():
             try:
                 while not self.ending.is_set():
@@ -152,6 +153,6 @@ class RequestHandler(object):
                     finally:
                         self._requests.task_done()
             except ReferenceError:  # dead weakref
-                pass
+                log.warning("ReferenceError in handler - dead weakref")
             log.info("RequestHandler worker: exiting cleanly")
         return self.handler.spawn(worker)
--- a/pykafka/handlers.py
+++ b/pykafka/handlers.py
@@ -17,11 +17,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 __all__ = ["ResponseFuture", "Handler", "ThreadingHandler", "RequestHandler"]
-import atexit
+
+from collections import namedtuple
 import functools
 import threading
+import weakref
+
 from .utils.compat import Queue, Empty
-from collections import namedtuple
 
 
 class ResponseFuture(object):
@@ -96,7 +98,9 @@ class RequestHandler(object):
         self.connection = connection
         self._requests = handler.Queue()
         self.ending = handler.Event()
-        atexit.register(self.stop)
+
+    def __del__(self):
+        self.stop()
 
     def request(self, request, has_response=True):
         """Construct a new request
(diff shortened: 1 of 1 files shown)


pykafka/handlers.py
```diff
@@ -124,17 +128,25 @@ def stop(self):
...
+                        if task.future:
+                            res = self.connection.response()
+                            task.future.set_response(res)
+                    except Exception as e:
+                        if task.future:
+                            task.future.set_error(e)
+                    finally:
+                        self._requests.task_done()
+            except ReferenceError:  # dead weakref
+                pass
```
emmett9001: We should have some warning message here - perhaps just at loglevel `INFO`.

yungchin: Thanks yes, coming up!

emmett9001:
This change is kind of scary in terms of how much it touches, but I've done a bunch of testing of both the consumer and producer on this branch, and I can't find any broken behavior. I think it's important to get `atexit` out of this module, since it really doesn't make sense to assume that any `RequestHandler` will live until the interpreter itself dies.

PR #259, opened 2015-09-07
travis: add back testinstances install

yungchin:
This was dropped in the test-suite upgrades as part of the py3
compatibility push.  As a result, the before-scripts fail, because there
is no testinstances installation outside the virtualenvs, and so every
TestCase has to instantiate a new test cluster.  Adding this back makes
things run a bit smoother.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ notifications:
 
 
 install:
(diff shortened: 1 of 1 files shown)


yungchin:
Judging from the output, travis-ci is content with this config update. The speedup is more modest than I'd hoped.

Issue #175, opened 2015-05-29
topic is not created when TopicDict is empty


Wombasta:
The new topic is not auto-created with error when no topics exist on kafka brokers (the TopicDict is empty {} ) 
```  
  File "/usr/local/lib/python2.7/dist-packages/pykafka/cluster.py", line 45, in __missing__
    if self._create_topic(key):
  File "/usr/local/lib/python2.7/dist-packages/pykafka/cluster.py", line 63, in _create_topic
    self._cluster.brokers[self._cluster.brokers.keys()[0]].request_metadata(topics=[topic_name])
IndexError: list index out of range
```
As soon as the random topic was created on broker manually, the new topic can be easily auto-created as designed.

kbourgoin:
The index error is actually from the broker list being empty, not the topic dict. It's part of a [known issue](https://issues.apache.org/jira/browse/KAFKA-2154) with Kafka. I put a warning [here](https://github.com/Parsely/pykafka/blob/35ad8c77c4ec75ac9eec5824b2368f0439273fb8/pykafka/cluster.py#L263-L268) about it, but I'll update the text to include that topic auto-creation won't work.

PR #262, opened 2015-09-10
Break Cluster-Topic refcycle

yungchin:
This partially addresses #257, as discussed there. I figured it was best to make a separate pullreq, as it doesn't have much to do with the changes in #258.

### DIFF
--- a/pykafka/broker.py
+++ b/pykafka/broker.py
@@ -270,8 +270,8 @@ class Broker():
     def request_metadata(self, topics=None):
         """Request cluster metadata
 
-        :param topics: The topic ids for which to request metadata
-        :type topics: Iterable of int
+        :param topics: The topic names for which to request metadata
+        :type topics: Iterable of `bytes`
         """
         max_retries = 3
         for i in range(max_retries):
--- a/pykafka/cluster.py
+++ b/pykafka/cluster.py
@@ -35,11 +35,23 @@ log = logging.getLogger(__name__)
 
 
 class TopicDict(dict):
-    """Dictionary which will attempt to auto-create unknown topics."""
+    """Lazy dict, which will also attempt to auto-create unknown topics"""
 
-    def __init__(self, cluster, *args, **kwargs):
+    def __init__(self, cluster, exclude_internal_topics, *args, **kwargs):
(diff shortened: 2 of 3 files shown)


yungchin:
As noted in the last commit msg, this now also happens to deal with #175 - topic auto-creation is enabled even on a fresh, empty cluster (there's still no broker metadata, but it can fall back on the bootstrap brokers).

emmett9001:
@yungchin thanks for making this totally separate from #258 - it makes them both a lot easier to think about.

emmett9001:
I've spent a while trying to grok this, and I understand it satisfactorily. Even outside of the specific object-leak bugfix, lazily creating `Topic`s is a good idea to mitigate memory usage. I'd like to run a few more tests on my own, but I'm close to totally convinced.

yungchin:
@emmett9001 thanks for checking this out! I was a bit spooked by the two test failures on travis just now:
https://travis-ci.org/Parsely/pykafka/jobs/81294673#L516

Have not seen these in local testing, but to save me some RAM I often run a two-node cluster (whereas we run a three-node cluster on travis), and I guess the smaller cluster is quicker to finish setting up a new topic.

In the failing tests, I think what goes wrong is this: `test_topic_autocreate` tries to auto-create a topic but gives up on it too soon (the cluster is still electing a leader and stuff), and so the test fails. Then  we enter `test_topic_updates`, and in the middle of that our new topic from `test_topic_autocreate` finally appears, so now this test errors too because it sees more new topics than it expected. (And both tests are new, which is why this has never come up before, but this seems to be the kind of failure that will keep coming up randomly....)

I'm not really sure what the best fix would be - I suppose `TopicDict._create_topic` needs to wait longer or should really be smarter about the wait, but I've no good suggestions on that :(

yungchin:
Out of curiosity, I restarted the test on travis earlier (the previous run's log output is [here][0] for posterity and such), and indeed it passed just fine the second time around. Of course that doesn't mean it's all ok - the test would continue to occasionally fail. The commit I've just pushed should fix that.

[0]: https://gist.github.com/yungchin/1ee48453b23a1319d4f4

Issue #253, opened 2015-09-03
Deadlock when using reset_offset


zsimpson:
In simpleconsumer I was passing a list of all partitions to reset. Sometimes (often but not always) it would lock up.  Adding traces I found that the lock up was at the line that reads:

```
if partition.fetch_lock.acquire(True):
```

In the `reset_offsets()` function.

I changed my code to call the reset_offsets function multiple times, once for each partition. I don't know if this just slows it down enough or what exactly, but it made the deadlock disappear.

My reproduction case is in a somewhat complex environment so I'm unable to provide a simple repro case, but if you'd like me to add traces or other debugging I'd be happy to do so.

emmett9001:
Thanks @zsimpson - turns out sharing resources between threads is hard. It might help to see what your debug traces look like when you hit this deadlock.

zsimpson:
Here's the trace. If you want I'll be happy to add more traces in. I can reproduce it every time right now it seems.

```
2015-09-04 16:01:05,232 - pykafka.simpleconsumer - INFO - Resetting offsets for 16 partitions
2015-09-04 16:01:05,238 - pykafka.simpleconsumer - INFO - Flushed queue for partition 6
2015-09-04 16:01:05,244 - pykafka.simpleconsumer - INFO - Flushed queue for partition 12
2015-09-04 16:01:05,249 - pykafka.simpleconsumer - INFO - Flushed queue for partition 8
2015-09-04 16:01:05,386 - pykafka.simpleconsumer - DEBUG - Fetched 4054 messages for partition 1
2015-09-04 16:01:05,416 - pykafka.simpleconsumer - DEBUG - Partition 1 queue holds 4054 messages
2015-09-04 16:01:05,422 - pykafka.simpleconsumer - DEBUG - Fetched 4168 messages for partition 3
2015-09-04 16:01:05,450 - pykafka.simpleconsumer - DEBUG - Partition 3 queue holds 4168 messages
2015-09-04 16:01:05,457 - pykafka.simpleconsumer - DEBUG - Fetched 4239 messages for partition 5
2015-09-04 16:01:05,487 - pykafka.simpleconsumer - DEBUG - Partition 5 queue holds 4239 messages
2015-09-04 16:01:05,492 - pykafka.simpleconsumer - DEBUG - Fetched 4205 messages for partition 9
2015-09-04 16:01:05,521 - pykafka.simpleconsumer - DEBUG - Partition 9 queue holds 4205 messages
[...]

PR #265, opened 2015-09-17
enforce id-based ordering before attempting to obtain partition locks.

emmett9001:
for #253

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -585,7 +585,8 @@ class SimpleConsumer():
 
         for broker, owned_partitions in iteritems(self._partitions_by_leader):
             partition_reqs = {}
-            for owned_partition in owned_partitions:
+            sorted_offsets = sorted(owned_partitions, key=lambda k: k[0].partition.id)
+            for owned_partition in sorted_offsets:
                 # attempt to acquire lock, just pass if we can't
                 if owned_partition.fetch_lock.acquire(False):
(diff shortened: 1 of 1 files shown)


pykafka/simpleconsumer.py
```diff
@@ -507,10 +507,13 @@ def _handle_success(parts):
 
         log.info("Resetting offsets for %s partitions", len(list(owned_partition_offsets)))
 
+        # sort offsets to avoid deadlocks
+        sorted_offsets = sorted(iteritems(owned_partition_offsets), key=lambda k: k[0].partition.id)
```
ltucker: I think you may want to recompute this within your retry loop since owned_partition_offsets is mutated by the loop.

emmett9001:
@yungchin This could use a sanity check.

yungchin:
@emmett9001 is the sorting in `fetch()` strictly necessary? As that code just skips over any partitions it cannot grab a lock on, doesn't that mean there shouldn't be an issue with deadlocks there? Other than that, this looks good to me.

emmett9001:
No, I don't think it's strictly necessary. I figured sorting is a good idea in general because it makes the whole system more predictable.

yungchin:
Ok, yeah that seems fine to me - I don't imagine the sorting is very expensive.

I've not been able to reproduce the bug here, so can't confirm that it's fixed, but I'm convinced that sorting before taking all the locks is a very real improvement either way.

PR #268, opened 2015-09-19
producer: avoid messages stuck in queues on stop

yungchin:
Because stop() called _wait_all() after stopping the OwnedBroker threads
(and because the OwnedBroker threads stop as soon as they see
 self.running=False, without trying to check for remaining messages), a
situation could occur where the OwnedBroker threads had already exited
but a message was still sitting in the queue - resulting in _wait_all()
and thus stop() looping forever.  We ended up with occasional hangs in
test_producer.test_async_produce_context.

This is a minimal fix, that just swaps those steps.

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -216,10 +216,10 @@ class Producer(object):
     def stop(self):
         """Mark the producer as stopped"""
         self._running = False
+        self._wait_all()
         if self._owned_brokers is not None:
             for owned_broker in self._owned_brokers.values():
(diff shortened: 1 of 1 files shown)


yungchin:
Just ran into this while starting work on producer futures, but now noticed it also happened on travis: https://travis-ci.org/Parsely/pykafka/jobs/81100168

I think this is fairly trivial so will merge right away.

emmett9001:
Looks fine, thanks.

PR #269, opened 2015-09-21
Producer futures

### LINKED ISSUE #238: Return futures from Producer.produce()
As per earlier discussion [here][0], I'm spinning this out into a separate issue. 
[0]: https://github.com/Parsely/pykafka/pull/234#discussion_r37731037

The proposal here would be that both `Producer.produce` and `RdKafkaProducer.produce` would return a future, enabling users to check the delivery status of produced messages to the kafka cluster.

I originally drafted a rough idea that used a `pykafka.handlers.ResponseFuture`, but I would now really propose that we return a (backported) `concurrent.futures.Future` instead. The former has never really been part of our public interface, and it wouldn't make sense to start leaking it now, if there's a more standard and more featureful alternative available.

The returned future would evaluate to `None` if delivery was successful (or optionally, it could report back both the partition_id and offset of the produced message, but I think we should save that for another day). If delivery wasn't successful, the future would re-raise the exception that occurred on the worker thread -- be it `InvalidMessageSize` or `RequestTimedOut` or any other that currently only appear in logging output.

@emmett9001 you're much more familiar with `Producer` internals, but I've been working on this anyway on the rdkafka side, so would also be happy to pick this up if you like.

yungchin:
This closes #238.

Error responses from kafka previously only appeared in log output. With the changes here, `produce()` returns a `concurrent.futures.Future`, through which any errors that persist after `max_retries` can be communicated to user code.

For `sync=True`, we simply evaluate the `Future` immediately, so that kafka error responses are raised directly from `produce()`.

This touches a few more lines than a minimal implementation would have done, I hope that's ok. I felt that slapping yet another item (ie the `delivery_future`) onto `message_partition_tup` was just going to get too unwieldly, so instead of `message_partition_tup` we're now moving instances of `Message` around. The second structural change is in `_send_request()`, where `to_retry` previously was an iterable of messages, but given that we were now going to add exceptions to that, and the error responses always apply to a whole `MessageSet`, `to_retry` is now an iterable of `MessageSet, Exception`.

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -295,13 +295,15 @@ class Producer(object):
                 if p_id == partition_id
             )
 
+        def mark_as_delivered(message_batch):
+            owned_broker.increment_messages_pending(-1 * len(message_batch))
+            for msg in message_batch:
+                msg.delivery_future.set_result(None)
+
         try:
             response = owned_broker.broker.produce_messages(req)
             if self._required_acks == 0:  # and thus, `response` is None
-                owned_broker.increment_messages_pending(
-                    -1 * len(message_batch))
-                for msg in message_batch:
-                    msg.delivery_future.set_result(None)
+                mark_as_delivered(message_batch)
                 return
 
             # Kafka either atomically appends or rejects whole MessageSets, so
@@ -311,12 +313,7 @@ class Producer(object):
             for topic, partitions in iteritems(response.topics):
                 for partition, presponse in iteritems(partitions):
                     if presponse.err == 0:
-                        # mark messages as successfully delivered
-                        delivered = req.msets[topic][partition].messages
-                        owned_broker.increment_messages_pending(
-                            -1 * len(delivered))
-                        for msg in delivered:
-                            msg.delivery_future.set_result(None)
+                        mark_as_delivered(req.msets[topic][partition].messages)
                         continue  # All's well
                     if presponse.err == UnknownTopicOrPartition.ERROR_CODE:
                         log.warning('Unknown topic: %s or partition: %s. '
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -315,28 +315,16 @@ class Producer(object):
                     if presponse.err == 0:
                         mark_as_delivered(req.msets[topic][partition].messages)
                         continue  # All's well
-                    if presponse.err == UnknownTopicOrPartition.ERROR_CODE:
-                        log.warning('Unknown topic: %s or partition: %s. '
-                                    'Retrying.', topic, partition)
-                    elif presponse.err == NotLeaderForPartition.ERROR_CODE:
-                        log.warning('Partition leader for %s/%s changed. '
-                                    'Retrying.', topic, partition)
+                    if presponse.err == NotLeaderForPartition.ERROR_CODE:
                         # Update cluster metadata to get new leader
                         self._update()
-                    elif presponse.err == RequestTimedOut.ERROR_CODE:
-                        log.warning('Produce request to %s:%s timed out. '
-                                    'Retrying.', owned_broker.broker.host,
-                                    owned_broker.broker.port)
-                    elif presponse.err == LeaderNotAvailable.ERROR_CODE:
-                        log.warning('Leader not available for partition %s.'
-                                    'Retrying.', partition)
-                    elif presponse.err == InvalidMessageError.ERROR_CODE:
-                        log.warning('Encountered InvalidMessageError')
-                    elif presponse.err == InvalidMessageSize.ERROR_CODE:
-                        log.warning('Encountered InvalidMessageSize')
-                    elif presponse.err == MessageSizeTooLarge.ERROR_CODE:
-                        log.warning('Encountered MessageSizeTooLarge')
-                    exc = ERROR_CODES[presponse.err]
+                    info = "Produce request for {}/{} to {}:{} failed.".format(
+                        topic,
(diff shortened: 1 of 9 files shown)


yungchin:
This needs an update after we're done with #279.

yungchin:
@rduplain @emmett9001 @kbourgoin I've just updated the description above. Would be grateful if you can check that this makes sense in terms of interface - and implementation of course.

kbourgoin:
Related to #238.

kbourgoin:
Connected to #238.

yungchin:
@emmett9001 can I bug you to review yet another one? :)

emmett9001:
It's on my list!

pykafka/protocol.py
```diff
@@ -137,8 +137,9 @@ class Message(Message, Serializable):
           Value => bytes
 
     :class:`pykafka.protocol.Message` also contains `partition` and
-    `partition_id` fields. Both of these have meaningless default values when
-    :class:`pykafka.protocol.Message` is used by the producer.
+    `partition_id` fields. Both of these have meaningless default values. When
+    :class:`pykafka.protocol.Message` is used by the producer, `partition_id`
+    identifies the Message's destination partition.
```
emmett9001: Since you're using `Message.delivery_future` in the producer, I think it should be declared in `__init__` with `None` as its initial value. Since we're going to be keeping a lot of `Message` instances in memory now, we may want to consider adding `__slots__` to this class. `__slots__` would serve a dual purpose: explicitly documenting all of the attributes a `Message` can ever have, and reducing the memory footprint of each message instance.

pykafka/producer.py
```diff
@@ -237,24 +238,27 @@ def produce(self, message, partition_key=None):
...
+            return msg.delivery_future.result()
+        return msg.delivery_future
 
-    def _produce(self, message_partition_tup):
+    def _produce(self, message):
         """Enqueue a message for the relevant broker
 
-        :param message_partition_tup: Message with partition assigned.
-        :type message_partition_tup: ((bytes, bytes), int) tuple
+        :param message: Message with partition assigned.
```
emmett9001: This docstring could use an update - something like "message with x, y, z assigned, ready to be enqueued"

pykafka/protocol.py
yungchin: Thanks, yes, that should make a really nice improvement. I'd never had a good use for `__slots__` before and had all but forgotten about them. Shall I just do that within this pull request? (I guess this is touching so many bits already, I might as well?)

emmett9001: Yes, I recommended it here because I think the change is small and pertinent enough to be included in this PR.

yungchin: Cool, will do that. Thanks!

emmett9001:
This looks great, thanks for the effort @yungchin! I suppose a common use case will be something like this:
```python
with topic.get_producer() as producer:
    count = 0
    futures = []
    while True:
        count += 1
        future = producer.produce('%s' % count)
        futures.append(future)
        if count % 80000 == 0:
            for future in futures:
                if future.exception() is not None:
                    raise future.exception()
            futures = []
            print "produced %s" % count
```
Is that about what you had in mind?

This request is big enough that I'd like to do another close pass over the code, but my impression after a readthrough is that this is pretty close.

pykafka/producer.py
```diff
@@ -272,84 +276,80 @@ def _send_request(self, message_batch, owned_broker):
...
-                self._produce((kv, partition_id, msg_attempt + 1))
+            for mset, exc in to_retry:
+                # XXX arguably, we should try to check these non_recoverables
+                # for individual messages in _produce and raise errors there
+                # right away, rather than failing a whole batch here?
+                non_recoverable = type(exc) in (InvalidMessageSize,
+                                                MessageSizeTooLarge)
+                for msg in mset.messages:
+                    if (non_recoverable
+                            or msg.produce_attempt >= self._max_retries):
```
emmett9001: [nitpick] No need for this linebreak, I think. I've been using 90 characters as a length limit, since github windows hold at least that without scrolling left to right.

yungchin:
Thanks, yep, that's it! There are also some [convenience functions][0] in the `futures` module, with which in your example you could split the list up in completed and pending futures, and throw the pending futures back on the list for the next round (otherwise checking `future.exception()` will block until the future is completed). I should probably steal that example and stick it somewhere in the docs, actually.

[0]: https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait

emmett9001:
Yes, I think it's a good idea to include a code sample like this right above your paragraph on how to use the futures.

yungchin:
Ok, I was poking around a bit with `sys.getsizeof` and the `__slots__` change seems to reduce `Message` overhead by 10x (`Message.__dict__` was over 1kB on py2 at least). However.... `Future` and its members weigh in at about twice that. The BDFL giveth, and the BDFL taketh away, or something like that.

emmett9001:
Oh wow. In that case we may want a way to turn off futures with a kwarg or something. We should keep that in mind if memory usage ever becomes a problem, but I don't think we necessarily need to add it now.

yungchin:
With the changes I made as part of f4f6df1, if the user ignores the future, it's gone immediately (the message only holds a weakref to it). So you'd still have the overhead of creating and destroying one, but you wouldn't have one lying around for every message in your outbound-queue.

(The weakref is luckily not in possession of a `__dict__`, so it's a lot smaller)

PR #293, opened 2015-10-07
reset_offset deadlock fix

### LINKED ISSUE #287: Potential for out-of-range error to overwrite a user's reset_offsets instruction?
( @emmett9001 I got thinking about this while reviewing #265, but since I shouldn't be heaping multiple issues into the same thread as I've done in the past, here's a fresh ticket :))

I believe that in the usage model that @zsimpson describes in #253 -- where you'd start a new consumer, then immediately call `reset_offsets()` on it -- there's a special case where you might end up with the default `auto_offset_reset` value, rather than whatever you just passed to `reset_offsets()`. Here's the sequence of events (or the theory, anyway):

  * user starts a new consumer
  * in `start()`, `fetch_offsets()` runs, which happens to obtain a bunch of stale offsets (this is the special case -- maybe it's because the topic's just been compacted; whatever)
  * the fetch-worker starts, enters `fetch()`, grabs a bunch of fetch locks
  * user calls `reset_offsets()`
  * `reset_offsets()` blocks trying to get a lock that the fetch-worker holds
  * the fetch-worker gets a response with `OffsetOutOfRangeError` on all partitions (not necessarily all, but that's the easiest case to reason about)
  * the fetch-worker then [releases locks on those partitions][0]
  * this means the user's `reset_offsets()` call can now grab those locks, and do its work
  * the fetch-worker waits for that to finish, then [continues with its own resets][1]
  * end result: the user-provided offsets got clobbered

I suspect it's really hard to get a test case written that demonstrates this reliably, but if we're happy to merge a fix without that test case, I'd think not releasing the locks before resetting, and instead using an `RLock` should do the trick. Happy to take a shot at that.

[0]: https://github.com/Parsely/pykafka/blob/770c4fc114dcb4ba0752240c81a56ae717ae50ab/pykafka/simpleconsumer.py#L619-L624
[1]: https://github.com/Parsely/pykafka/blob/770c4fc114dcb4ba0752240c81a56ae717ae50ab/pykafka/simpleconsumer.py#L627-L630

emmett9001:
This pull request fixes #287 by using an `RLock`. The logic this replaces was actually written before I'd learned about `RLock`s, so I think this is pretty sensible. It's still core enough to the operation of the consumer that it needs some fairly careful testing before a merge.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -615,20 +615,6 @@ class SimpleConsumer():
                         raise
 
                 parts_by_error = build_parts_by_error(response, self._partitions_by_id)
-                # release the lock in these cases, since resolving the error
-                # requires an offset reset and not releasing the lock would
-                # lead to a deadlock in reset_offsets. For successful requests
-                # or requests with different errors, we still assume that
-                # it's ok to retain the lock since no offset_reset can happen
-                # before this function returns
-                out_of_range = parts_by_error.get(OffsetOutOfRangeError.ERROR_CODE, [])
-                for owned_partition, res in out_of_range:
-                    owned_partition.fetch_lock.release()
-                    # remove them from the dict of partitions to unlock to avoid
-                    # double-unlocking
-                    partition_reqs.pop(owned_partition)
-                # handle the rest of the errors that don't require deadlock
-                # management
                 handle_partition_responses(
                     self._default_error_handlers,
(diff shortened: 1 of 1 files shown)


yungchin:
This looks perfectly good to me, but Travis disagrees: https://travis-ci.org/Parsely/pykafka/jobs/84204893 (I've not seen that particular test stalling before, I think). It may not be down to these changes of course.

PR #294, opened 2015-10-09
SimpleConsumer/RdKafkaSimpleConsumer switching options

yungchin:
This moves the rdkafka switch (which wasn't functioning yet, and which was written in a past where the rdkafka branch had its own implementations of `Cluster`, `Topic`, `Partition`, etc) out of `pykafka.client` and into `pykafka.topic`. You would then select an rdkafka-backed consumer like so: `topic.get_simple_consumer(use_rdkafka=True)` (and similarly for a balanced consumer).

Note this is a pullreq based on the rdkafka_extension branch, I split it out as it's probably useful to discuss separately.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -73,7 +73,7 @@ class BalancedConsumer():
                  zookeeper=None,
                  auto_start=True,
                  reset_offset_on_start=False,
-                 use_rdkafka=True):
+                 use_rdkafka=False):
         """Create a BalancedConsumer instance
 
         :param topic: The topic this consumer should consume
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -33,6 +33,10 @@ from .exceptions import (KafkaException, PartitionOwnedError,
                          ConsumerStoppedException)
 from .simpleconsumer import SimpleConsumer
 from .utils.compat import range, get_bytes
+try:
+    from . import rdkafka
+except ImportError:
+    rdkafka = False
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


pykafka/balancedconsumer.py
```diff
@@ -267,7 +275,9 @@ def _setup_internal_consumer(self, start=True):
             # _setup_internal_consumer. subsequent calls should not
             # reset the offsets, since they can happen at any time
             reset_offset_on_start = False
-        self._consumer = SimpleConsumer(
+        Cls = (rdkafka.RdKafkaSimpleConsumer
+               if self._use_rdkafka else SimpleConsumer)
```
emmett9001: This is pretty slick - it will be awesome if the interfaces can stay identical.

pykafka/client.py
```diff
@@ -59,8 +57,8 @@ def __init__(self,
             milliseconds) when reading responses for offset commit and
             offset fetch requests.
         :type offsets_channel_socket_timeout_ms: int
-        :param ignore_rdkafka: Don't use rdkafka, even if installed.
-        :type ignore_rdkafka: bool
+        :param ignore_rdkafka: Deprecated, and in fact ignored. Instead, you
```
emmett9001: It's annoying that to conform strictly to semantic versioning we'd have to increment the version again to truly get rid of this.

emmett9001:
I like this interface. It's important that `get_balanced_consumer` and `get_simple_consumer` as close an interface as possible, and the way you've done that here is good. I think this is fine to merge into the main rdkafka branch, which i also hope to review soon.

pykafka/client.py
yungchin: Argh, yes, I really should have thought of making this removal a separate pullreq. It would have been in time for 2.0.0. Somehow I totally didn't think of that :|

yungchin:
Thanks @emmett9001!!

pykafka/client.py
emmett9001: Know what, I thought about this some more and realized that it's pointless to just deprecate this since it's never done anything to begin with. This is the rare case where I'd actually prefer to break old code that uses this kwarg (if any such code exists) without incrementing the major version.

yungchin: Ok cool, I'll get rid of it!

Issue #295, opened 2015-10-09
Cluster does not reconnect to disconnected brokers


cscheffler:
Behaviour: a KafkaClient cluster does not reconnect to disconnected brokers that have become available again.

Test environment: a 3 broker Kafka cluster with a replication factor of 2 on the test topic.

To reproduce:
* Connect a producer to the cluster and produce messages continuously.
* The producer connects to 2 out of 3 brokers in the cluster, since the replication factor is 2. For this example, let's say we are connected to brokers 1 and 2, but not to broker 3.
* Stop broker 1.
* The producer notices the disconnect and updates the leader from 1 to 2 or 3 on some of its partitions.
* Everything keeps running smoothly.
* Start broker 1 again.
* Everything keeps running smoothly.
* Stop broker 2.
* The producer notices the disconnect and updates the leader from 2 to 1 or 3 on some of its partitions.
* The producer never tried to reconnect to broker 1, so sending messages fails.
* After 3 retries, the producer dies with an exception.

Error output:
```
2015-10-09 15:29:33,445 ERROR Exception encountered in worker thread:
  File "/workspace/pykafka/pykafka/producer.py", line 403, in queue_reader
    self.producer._send_request(batch, self)
  File "/workspace/pykafka/pykafka/producer.py", line 351, in _send_request
    "retries.", self._max_retries)

2015-10-09 15:29:33,448 INFO Blocking until all messages are sent
Traceback (most recent call last):
  File "produce-pykafka.py", line 46, in <module>
    break
  File "/workspace/pykafka/pykafka/producer.py", line 169, in __exit__
    self.stop()
  File "/workspace/pykafka/pykafka/producer.py", line 219, in stop
    self._wait_all()
  File "/workspace/pykafka/pykafka/producer.py", line 363, in _wait_all
    self._raise_worker_exceptions()
  File "/workspace/pykafka/pykafka/producer.py", line 154, in _raise_worker_exceptions
    raise ex
pykafka.exceptions.ProduceFailureError: ('Message failed to send after %d retries.', 3)
```

Proposed fix: Try to reconnect to all available brokers on a call to cluster.update(). Some connections might fail, but just leave those disconnected until the next call to update().

cscheffler:
This pull request attempts to solve the problem: https://github.com/Parsely/pykafka/pull/296

Tested on my local cluster.

emmett9001:
Excellent, I'll take a look at this on Monday.

[...]

PR #296, opened 2015-10-09
Attempt reconnect to disconnected brokers on cluster update.

cscheffler:
See the following issue for details: https://github.com/Parsely/pykafka/issues/295

### DIFF
--- a/pykafka/cluster.py
+++ b/pykafka/cluster.py
@@ -253,6 +253,13 @@ class Cluster(object):
                     source_host=self._source_host,
                     source_port=self._source_port
                 )
+            elif not self._brokers[id_].connected:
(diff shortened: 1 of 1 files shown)


amontalenti:
Assigning to @emmett9001 who was hardening the broker reconnect logic. This looks like a good improvement. Thank you for your contribution, @cscheffler.

cscheffler:
Happy to contribute.

We're in the process of upgrading to Kafka 0.8.2 (from 0.8.1) and to pykafka. We'd like to do the upgrade with a rolling restart and zero service interruption, so lots of testing happening at the moment.

emmett9001:
@cscheffler I've been able to replicate both the failure of this case on master and the solution in your branch. Well done, and thanks for the contribution!

Issue #289, opened 2015-10-07
minimum consumer lag is 1, not 0


dribler:
I noticed that, even if consumer consumed all messages, lag value is 1. I expected that after consuming ALL message, lag value should be 0.

I thought this is Kafka issue, but i've made experiment and below what i got.

So i have 3 brokers and small topic:

```
Topic:test     PartitionCount:6        ReplicationFactor:3     Configs:
        Topic: test    Partition: 0    Leader: 0       Replicas: 0,2,1 Isr: 0,2,1
        Topic: test    Partition: 1    Leader: 1       Replicas: 1,0,2 Isr: 1,0,2
        Topic: test    Partition: 2    Leader: 2       Replicas: 2,1,0 Isr: 2,1,0
        Topic: test    Partition: 3    Leader: 0       Replicas: 0,1,2 Isr: 0,1,2
        Topic: test    Partition: 4    Leader: 1       Replicas: 1,2,0 Isr: 1,2,0
        Topic: test    Partition: 5    Leader: 2       Replicas: 2,0,1 Isr: 2,0,1
```

I produced 4 messages and consumed with kafka-console-consumer:

/opt/kafka/bin/kafka-consumer-offset-checker.sh --topic test --group test --zookeeper localhost:2181
```
Group           Topic                          Pid Offset          logSize         Lag             Owner
test            test                          0   0               0               0               none
test            test                          1   0               0               0               none
test            test                          2   4               4               0               none
test            test                          3   0               0               0               none
test            test                          4   0               0               0               none
test            test                          5   0               0               0               none
```

Lag is 0. I expected that.

And now 4 messages consumed with pykafka 2.0.0

 /opt/kafka/bin/kafka-consumer-offset-checker.sh --topic test --group test2 --zookeeper localhost:2181
```
Group           Topic                          Pid Offset          logSize         Lag             Owner
test           test3                          0   -1              0               unknown         none
test           test3                          1   -1              0               unknown         none
test           test3                          2   3               4               1               none
test           test3                          3   -1              0               unknown         none
test           test3                          4   -1              0               unknown         none
test           test3                          5   -1              0               unknown         none
```
Lag is 1. I had same on pykafka 1.1.0

We have this at production and i able to reproduce it locally everytime. Is it pykafka issue? Maybe i do something wrong?
This actually is not a big problem - there are no missed messages, this just confused a bit.

emmett9001:
[...]

Issue #216, opened 2015-07-27
Proposal for changes in manual offset management


yungchin:
While working on #213 I ran into a confusing quirk, where `held_offsets` may return an offset of -1 (ie we're at the start of the partition), which seems fine until you do the following: 1) store the offsets from `held_offsets` in your application's external offset store, 2) feed them back from there into `reset_offsets`. The problem is that Kafka defines -1 to mean LATEST (it would not have been a problem if LATEST was -2 instead, and EARLIEST was -1). So you may get:

```
In [18]: cons = top.get_simple_consumer(reset_offset_on_start=True, auto_offset_
reset=pykafka.common.OffsetType.EARLIEST)                                       

In [19]: held = cons.held_offsets

In [20]: held
Out[20]: {0: -1, 1: -1, 2: -1}

In [21]: cons.reset_offsets([(cons.partitions[pid], offset) for pid, offset in h
eld.items()])

In [22]: cons.held_offsets
Out[22]: {0: -1, 1: 99, 2: -1}
```

We can avoid the magic numbers altogether by doing something like

 * creating a property `next_offsets` which would just return `held_offsets` + 1, so to speak
 * (plus slap a deprecation warning on `held_offsets`)
 * creating a kwarg `next_offsets` for `reset_offsets()`, which accepts the same as its `partition_offsets` kwarg but expects all values + 1
 * (plus slap a deprecation warning on the `partition_offsets` kwarg)

If that makes sense, I'm happy to write the pullreq.

kbourgoin:
I'm generally +1 to the idea, but need to think on it a bit more. I can see the issue with the current setup, but I'm confused why in the example provided only partition `1` has its offset changed whereas the other two are unchanged.

yungchin:
Ah, yes sorry that's a bit confusing: only that partition held any messages. I quickly created the test-topic using the KafkaInstance producer, which dumps all its messages on a single partition if you only call `produce_messages` once.

yungchin:
If we're headed for a major version bump soon, would that be a good time to move on this one? Would save us all the deprecation warnings :)

yungchin:
(Just ran into an issue with this again.)

As an alternative fix, that wouldn't change as much of the interface, how about if we only change the implementation of `held_offsets` so that it returns `-2` (ie `OffsetType.EARLIEST) when it holds `-1` internally? So

```python
    @property
    def held_offsets(self):
        """Return a map from partition id to held offset for each partition"""
[...]

PR #298, opened 2015-10-12
consumer off-by-one

emmett9001:
This pull request changes the way `SimpleConsumer` maintains its internal offset counters. In `OffsetCommitRequest`s, the consumer now sends the *next* offset to consume rather than the last one consumed, fixing the issue raised in #289.

Based on the test cases in #289, this seems to be what Kafka expects. I can't find anywhere in the documentation whether this is technically correct or not, but I suppose hoping for that is a bit of a long shot given the state of their docs. This pull request also adjusts some existing consumer tests based on the assumption that Kafka expects this.

### DIFF
--- a/tests/pykafka/test_simpleconsumer.py
+++ b/tests/pykafka/test_simpleconsumer.py
@@ -6,7 +6,7 @@ from uuid import uuid4
 from pykafka import KafkaClient
 from pykafka.simpleconsumer import OwnedPartition, OffsetType
 from pykafka.test.utils import get_cluster, stop_cluster
-from pykafka.utils.compat import range
+from pykafka.utils.compat import range, iteritems
 
 
 class TestSimpleConsumer(unittest2.TestCase):
@@ -57,7 +57,7 @@ class TestSimpleConsumer(unittest2.TestCase):
             offsets_committed = consumer.held_offsets
             consumer.commit_offsets()
 
-            offsets_fetched = dict((r[0], r[1].offset)
+            offsets_fetched = dict((r[0], r[1].offset - 1)
                                    for r in consumer.fetch_offsets())
             self.assertEquals(offsets_fetched, offsets_committed)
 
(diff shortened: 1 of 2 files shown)


emmett9001:
@yungchin what do you think about this change? The change in `build_offset_commit_request` fixes the issue form #289, and the change in `fetch_offsets` seems to me the sanest way to make `test_offset_resume` pass. Since that test is ensuring that `held_offsets` doesn't change when a new consumer is created, the change has to be made to the consumer itself.

yungchin:
@emmett9001 this looks like a complete fix to me, but I may well be missing something - I mean I surely missed all of this before, even though we had [this conversion in place already][0] in one other place. That is, I wonder if it won't come back to bite us later, with these +1s and -1s that we might forget. If kafka uses "next offset" rather than "last offset" in all its messages, aren't we better off replacing `last_offset_consumed` by `next_offset` (I know, that name's taken, so some other name) internally altogether?

[0]: https://github.com/Parsely/pykafka/blob/1907feb07bfa3531c0213949c32a734d36274bfa/pykafka/simpleconsumer.py#L472-L474

tests/pykafka/test_simpleconsumer.py
```diff
@@ -44,7 +44,8 @@ def _get_simple_consumer(self, **kwargs):
         consumer.stop()
 
     def test_consume(self):
-        with self._get_simple_consumer() as consumer:
+        with self._get_simple_consumer(auto_offset_reset=OffsetType.EARLIEST,
+                                       reset_offset_on_start=True) as consumer:
```
yungchin: Was this test broken by these changes? Mostly curious, but also it makes me feel like I don't fully understand the changes.

emmett9001: This test wasn't broken by the changes, but I wanted to make sure that if this one ever runs after `test_consumer_lag` that it doesn't fail. Since both of these tests consume the entire topic, I wanted to be sure this one starts from the beginning. Thinking about it though, they use different consumer groups, so I suppose this change shouldn't be necessary.

emmett9001:
@yungchin `last_offset_consumed` and `next_offset` are both necessary for the consumer to work properly. `next_offset` is always the next offset to fetch, but `last_offset_consumed` indicates the most recent message that was actually returned from a `consume` call. I tried a few different ways to remove one of these, but I realized that the current design of the consumer actually necessitates two different counters. Are you suggesting a name change for one of them?

yungchin:
Yes, sorry, that's what I meant - I know `next_offset` is already in use, so indeed I meant we'd need another name for `last_offset_consumed`, if - and that would be the proposal - we'd no longer store "last offset consumed" in it, but "next offset to consume".

emmett9001:
@yungchin the current state of this branch uses logic that's consistent with master but requires less `-1`/`+1` offset logic. Let me know if you see anything that you think should be changed.

yungchin:
I've added three comments in-line. Additionally, I think there's still one `last_offset_consumed` in `balancedconsumer.py`. No other issues that occur to me!

In fact I think that, as a happy side-effect, this implements #216 too?

emmett9001:
My goal was to avoid changing a ton of unrelated tests and `held_offsets`. If you think it's preferable to make wide-reaching changes to those, I'd go along with it. I'm just torn on whether your suggestions would actually be an improvement.

Issue #299, opened 2015-10-13
get_simple_consumer() TypeError: object of type 'NoneType' has no len()


asifhj:
Getting break on last message read from queue.

2015-10-13 11:20:54,438.438.999891281:pykafka.simpleconsumer:8868:DEBUG:1916:Partition 25 queue holds 5 messages
Traceback (most recent call last):
  File "C:/Users/asifj/PycharmProjects/Python_SOAP_OSSJ_SAP_Fusion_Kafka_Spark_HBase/KafkaConsumerDEV.py", line 28, in <module>
    for message in consumer:
  File "C:\Python27\lib\site-packages\pykafka\simpleconsumer.py", line 288, in __iter__
    message = self.consume(block=True)
  File "C:\Python27\lib\site-packages\pykafka\simpleconsumer.py", line 311, in consume
    while not message:
  File "C:\Python27\lib\site-packages\pykafka\protocol.py", line 169, in __len__
    size = 4 + 1 + 1 + 4 + 4 + len(self.value)
TypeError: object of type 'NoneType' has no len()
0 ?????

emmett9001:
Hi @asifhj, thanks for reporting this. Are you able to post a code snippet that shows how you're triggering this error? Also, what version of pykafka are you running?

asifhj:
Here is the code snippet

client = KafkaClient(hosts="xxx.22.147.232:9092,xxx.22.147.242:9092,xxx.22.147.243:9092")
print client.topics
topic = client.topics['SAPEvent']
consumer = topic.get_simple_consumer()
for message in consumer:
    if message is not None:
        print "Offset: "+str(message.offset)+"\tMessage: "+str(message.value)

asifhj:
Yes this fix resolves it.

emmett9001:
How about the code that uses the producer? Is it possible that you're putting a `None` into a message without realizing it?

asifhj:
Producer is not able to send "None", I am reading from already queued messages.

emmett9001:
Can you create a test case that reliably reproduces this issue? It's unclear how the consumer is getting into a state where this error becomes possible.

yungchin:
I can't immediately see what the struct helpers return when a message is received with a null payload, but I suspect this is what would cause `None` to appear in `Message.value`. Our producer cannot produce such messages yet, so this would have to mean @asifhj used some other client to produce these messages?

[...]

PR #300, opened 2015-10-13
Fix for SimpeConsumer TypeError: object of type 'NoneType' has no len()

asifhj:
I have added if condition to check the object type before evaluating the length of it.

(edited for waffle.io benefit: connected to #299)

### DIFF
--- a/pykafka/protocol.py
+++ b/pykafka/protocol.py
@@ -172,10 +172,13 @@ class Message(Message, Serializable):
         self.produce_attempt = produce_attempt
 
     def __len__(self):
-        size = 4 + 1 + 1 + 4 + 4 + len(self.value)
-        if self.partition_key is not None:
-            size += len(self.partition_key)
-        return size
+        if not self.value==None:
+            size = 4 + 1 + 1 + 4 + 4 + len(self.value)
+            if self.partition_key is not None:
+                size += len(self.partition_key)
+            return size
+        else:
+            return 0
 
     @classmethod
     def decode(self, buff, msg_offset=-1, partition_id=-1):


emmett9001:
Related to #299

emmett9001:
I think this fix is likely to have some side effects. A cleaner solution to avoid `None` message values could be to perform a `None` check [here](https://github.com/Parsely/pykafka/blob/64df140a062ec0bb921eabf56180090d9b4a0b67/pykafka/protocol.py#L165) and raise an exception. Still, I have no idea how the consumer could receive a `None` there in the first place.

asifhj:
I have tested it with almost 1Gig of data of different types of formats and
with empty value, it has not broken anywhere.

On Tue, Oct 13, 2015 at 7:18 PM, Emmett J. Butler <notifications@github.com>
wrote:

> I think this fix is likely to have some side effects. A cleaner solution
> to avoid None message values could be to perform a None check here
> <https://github.com/Parsely/pykafka/blob/64df140a062ec0bb921eabf56180090d9b4a0b67/pykafka/protocol.py#L165>
> and raise an exception. Still, I have no idea how the consumer could
> receive a None there in the first place.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Parsely/pykafka/pull/300#issuecomment-147720466>.
>



-- 
*Asif J.*
*Cell:      +91 97385 47018*
*Github: asifhj*

*Skype:  asif.h.j*

asifhj:
I have tested it with almost 1Gig of data of different types of formats and with empty value, it has not broken anywhere.

pykafka/protocol.py
```diff
@@ -172,10 +172,13 @@ def __init__(self,
...
-        if self.partition_key is not None:
-            size += len(self.partition_key)
-        return size
+        if not self.value==None:
+            size = 4 + 1 + 1 + 4 + 4 + len(self.value)
+            if self.partition_key is not None:
+                size += len(self.partition_key)
+            return size
+        else:
+            return 0
```
yungchin: It could be problematic that this returns the wrong length: if `value` is `None`, there's still the length of the headers plus the length of `partition_key` (it wouldn't be very useful to send a null payload without sending a key, so we should expect `partition_key` to be populated).

asifhj: Then we will return like `return 4 + 1 + 1 + 4 + 4`

yungchin:
It would be great if there could be a test case with this - as Emmett said it's not obvious how `None` values appear. If these are indeed messages with null payloads, we'd have to either use the command line tools to produce such test messages, or quickly extend our `Producer` to enable null payloads there too. I'll check how hard it would be to add that feature.

pykafka/protocol.py
asifhj: Below code will consider everything except for `None` length
`size = 4 + 1 + 1 + 4 + 4
if self.partition_key is not None:
      size += len(self.partition_key)
return size`

yungchin:
@asifhj I had a quick go patching `Producer` to allow it to send `None` messages, thinking that that would make it easier to test this pullreq, but of course that also uses `protocol.Message`, so I ended up fixing this as a part of that. See #306. Thanks for pointing out the problem spot!

asifhj:
Great! Thanks for the update.

Issue #271, opened 2015-09-28
BalancedConsumer.stop() needs to delay releasing its zookeeper nodes


yungchin:
(This also out of #206)

Currently, `stop()` begins with stopping its zookeeper handle, but as long as we haven't stopped the internal consumer, we should still keep the consumed partitions to ourselves.

PR #304, opened 2015-10-14
BalancedConsumer: remove znodes when stopping

### LINKED ISSUE #276: BalancedConsumer.stop() should remove its ephemeral nodes if it doesn't own the zookeeper handle
This was reported by @fortime in a comment on #227.

yungchin:
These changes resolve #276 and #271.

As noted in 6abec8f I couldn't come up with a simple, satisfactory solution to clean up the watches we set. Having had a quick browse through the implementation of kazoo's `ChildrenWatch`, I suspect it may even get in the way of a `BalancedConsumer` ever getting gc'd. The workaround for that though would be to drop `ChildrenWatch` altogether, which would be a much bigger change - not one I thought we ought to tackle right now.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -517,7 +517,7 @@ class BalancedConsumer():
 
     def _brokers_changed(self, brokers):
         if not self._running:
-            return False  # disables the watcher
+            return False  # `False` tells ChildrenWatch to disable this watch
         if self._setting_watches:
             return
         log.debug("Rebalance triggered by broker change")
@@ -525,7 +525,7 @@ class BalancedConsumer():
 
     def _consumers_changed(self, consumers):
         if not self._running:
-            return False  # disables the watcher
+            return False  # `False` tells ChildrenWatch to disable this watch
         if self._setting_watches:
             return
         log.debug("Rebalance triggered by consumer change")
@@ -533,7 +533,7 @@ class BalancedConsumer():
 
     def _topics_changed(self, topics):
         if not self._running:
-            return False  # disables the watcher
+            return False  # `False` tells ChildrenWatch to disable this watch
         if self._setting_watches:
             return
         log.debug("Rebalance triggered by topic change")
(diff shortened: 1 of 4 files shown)


pykafka/balancedconsumer.py
```diff
@@ -392,12 +400,15 @@ def _add_self(self):
         if len(self._topic.partitions) <= len(participants):
             raise KafkaException("Cannot add consumer: more consumers than partitions")
 
-        path = '{path}/{id_}'.format(
+        self._zookeeper.create(
+            self._path_self(), self._topic.name, ephemeral=True, makepath=True)
+
+    def _path_self(self):
```
emmett9001: This seems like a pretty good case to use `@property`

pykafka/balancedconsumer.py
```diff
@@ -504,18 +515,24 @@ def _check_held_partitions(self):
             self._rebalance()
 
     def _brokers_changed(self, brokers):
+        if not self._running:
+            return False  # disables the watcher
```
emmett9001: Why `False` instead of `None`?

yungchin: `ChildrenWatch` explicitly checks for `False`: https://github.com/python-zk/kazoo/blob/bbf581e3db992e22b82f1264b1a21308595699fa/kazoo/recipe/watchers.py#L334

It's really weird though, if you look at that function - I don't think `return False` kills the watcher thread, it just sets it `_stopped`.

emmett9001: That could definitely use a comment explanation, because it just looks wrong next to the succeeding lines.

yungchin: Yes, do you think I should also add `return True` at the end? It should work fine without that, because the caller distinguishes between `None` and `False`, but maybe that makes it look more even?

emmett9001: If `True` doesn't actually have any meaning to the caller, I think it's actually more readable to leave it out.

yungchin: It doesn't. Ok cool.

emmett9001:
This looks good to me if you address the two comments.

PR #306, opened 2015-10-14
protocol: enable null payloads

### LINKED ISSUE #299: get_simple_consumer() TypeError: object of type 'NoneType' has no len()

yungchin:
These changes enable both the producer and consumer to handle null payloads (represented as `None` in python), which are meaningful in kafka as "delete markers" for log compaction. I've added a test that sends and receives `None`.

Resolves #299.

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -224,13 +224,13 @@ class Producer(object):
     def produce(self, message, partition_key=None):
         """Produce a message.
 
-        :param message: The message to produce
+        :param message: The message to produce (use None to send null)
         :type message: bytes
         :param partition_key: The key to use when deciding which partition to send this
             message to
         :type partition_key: bytes
         """
-        if not isinstance(message, bytes):
+        if not (isinstance(message, bytes) or message is None):
             raise TypeError("Producer.produce accepts a bytes object, but it "
                             "got '%s'", type(message))
         if not self._running:
--- a/pykafka/protocol.py
+++ b/pykafka/protocol.py
@@ -172,7 +172,9 @@ class Message(Message, Serializable):
         self.produce_attempt = produce_attempt
 
     def __len__(self):
-        size = 4 + 1 + 1 + 4 + 4 + len(self.value)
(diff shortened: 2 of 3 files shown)


emmett9001:
Wow, TIL:
```
In [2]: class A():
    def __len__(self):
        print "hey"
        return 2
   ...:     

In [3]: a = A()

In [4]: bool(a)
hey
Out[4]: True
```
Also, verified that this PR is sound with this:
```
In [22]: buff = bytearray(10)                                                                                                                                               
In [24]: struct.pack_into("i", buff, 0, *(-1,))
In [25]: buff
Out[25]: bytearray(b'\xff\xff\xff\xff\x00\x00\x00\x00\x00\x00')
In [26]: buff = bytearray(10)                                                                                                                                                                  
In [27]: struct.pack_into("i0s", buff, 0, *(-1, b""))                                                                                                                                          
In [28]: buff
Out[28]: bytearray(b'\xff\xff\xff\xff\x00\x00\x00\x00\x00\x00')
```

yungchin:
Thanks for the fast review! Heh, I had the same thing re the `__len__` and `bool` bit - at first I thought the traceback in the issue report had been truncated somehow.

PR #307, opened 2015-10-15
[subject]: In handlers.py, add except SocketDisconnectedError; In con…

nashuiliang:
When the socket is closed, The program will reconnect.
When the program executes `BrokerConnection.reconnect`, the program throws exception.

```
Exception in thread Thread-1:
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/vagrant/github.com/.../pykafka/handlers.py", line 153, in worker
    self.connection.reconnect()
  File "/vagrant/github.com/.../pykafka/connection.py", line 101, in reconnect
    self.connect()
TypeError: connect() takes exactly 2 arguments (1 given)
```

The code of the raw program `BrokerConnection.reconnect`:
```
def reconnect(self):
        """Disconnect from the broker, then reconnect"""
        self.disconnect()
        self.connect()
```

I change the `BrokerConnection.connect`
```
 def connect(self, timeout):
        """Connect to the broker."""
        ...
```
to:
```
def connect(self, timeout=10 * 1000):
        """Connect to the broker."""
        ...
```

### DIFF
--- a/pykafka/connection.py
+++ b/pykafka/connection.py
@@ -73,7 +73,7 @@ class BrokerConnection(object):
         """Returns true if the socket connection is open."""
         return self._socket is not None
 
-    def connect(self, timeout):
+    def connect(self, timeout=10 * 1000):
         """Connect to the broker."""
         log.debug("Connecting to %s:%s", self.host, self.port)
         self._socket = socket.create_connection(
--- a/pykafka/handlers.py
+++ b/pykafka/handlers.py
@@ -24,6 +24,7 @@ import logging
 import threading
 import weakref
 
+from .exceptions import SocketDisconnectedError
 from .utils.compat import Queue, Empty
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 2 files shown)


yungchin:
Thanks for putting this in! Given that @emmett9001 is working on related changes under #223 and #224,  I'll assign this to him too.

emmett9001:
Hi @nashuiliang, thanks for this. I'm not sure it's safe to simply `reconnect()` in response to a `SocketDisconnectedError`. How does this handle the case where the service on the other end of the connection is down permanently? It seems like this could create an infinite loop of `reconnect()` calls as more requests are made and fail.

I assume you're encountering this issue in the consumer, since the producer is pretty good at handling unresponsive brokers. If that's the case, I'd love to hear a bit about the motivation behind this pull request.

(By the way, I've fixed the interpreter error you found [here](https://github.com/Parsely/pykafka/commit/3de3eb1e854f18bb92babe4927edac568972a3f4) - thanks for that!)

emmett9001:
@nashuiliang You may want to try out [this branch](https://github.com/Parsely/pykafka/pull/322) and see if it solves the problem you're attempting to solve here.

emmett9001:
@nashuiliang It looks like current master fully handles the case you've pointed out here. I'm going to close this pull request - if you have further concerns about broker reconnection, feel free to comment here or open a new pull request. Thanks for your contribution!

PR #308, opened 2015-10-16
Recover after zk session loss

### LINKED ISSUE #204: BalancedConsumer incorrectly handles loss of its ephemeral zookeeper nodes

yungchin:
This changes `_rebalance()` to always use zk as the source of truth for what partitions we hold, rather than `self._partitions`. The main reason to do so is that `self._partitions` may be out of date, if `_rebalance()` is triggered after a zk session expiration. This was originally reported by @fortime in #215 - I'm however opting for a slightly more conservative solution (that is, querying zk every time).

This resolves #204. The new test in 8faf6c1 demonstrates the issue by simulating a session expiration. There are also two other fixes in 01368f6 and 23698e3 that were small enough that I thought I might as well slip them in here.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -271,13 +271,15 @@ class BalancedConsumer():
         self._zookeeper = KazooClient(zookeeper_connect, timeout=timeout / 1000)
         self._zookeeper.start()
 
-    def _setup_internal_consumer(self, partitions=[], start=True):
+    def _setup_internal_consumer(self, partitions=None, start=True):
         """Instantiate an internal SimpleConsumer.
 
         If there is already a SimpleConsumer instance held by this object,
         disable its workers and mark it for garbage collection before
         creating a new one.
         """
+        if partitions is None:
+            partitions = []
         reset_offset_on_start = self._reset_offset_on_start
         if self._consumer is not None:
             self._consumer.stop()
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -242,7 +242,11 @@ class BalancedConsumer():
 
         This method should be called as part of a graceful shutdown process.
         """
-        self._running = False
+        with self._rebalancing_lock:
+            # We acquire the lock in order to prevent a race condition where a
+            # rebalance that is already underway might re-register the zk
+            # nodes that we remove here
+            self._running = False
         self._consumer.stop()
         if self._owns_zookeeper:
             # NB this should always come last, so we do not hand over control
@@ -425,6 +429,8 @@ class BalancedConsumer():
         if self._consumer is not None:
             self.commit_offsets()
         with self._rebalancing_lock:
+            if not self._running:
+                raise ConsumerStoppedException
             log.info('Rebalancing consumer %s for topic %s.' % (
                 self._consumer_id, self._topic.name)
             )
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -505,7 +505,7 @@ class BalancedConsumer():
             owner_id, stat = self._zookeeper.get(
                 '{path}/{slug}'.format(
                     path=self._topic_path, slug=partition_slug))
-            if owner_id == self._consumer_id:
+            if owner_id == get_bytes(self._consumer_id):
                 zk_partition_ids.add(int(partition_slug.split('-')[1]))
         # build a set of partition ids we think we own
         internal_partition_ids = set([p.id for p in self._partitions])
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -525,7 +525,7 @@ class BalancedConsumer():
         if zk_partitions != self._partitions:
             log.warning("Internal partition registry doesn't match ZooKeeper!")
             log.debug("Internal partition ids: %s\nZooKeeper partition ids: %s",
-                      internal_partition_ids, zk_partition_ids)
+                      self._partitions, zk_partitions)
             return False
         return True
 
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -501,11 +501,14 @@ class BalancedConsumer():
         zk_partition_ids = set()
         all_partitions = self._zookeeper.get_children(self._topic_path)
         for partition_slug in all_partitions:
-            owner_id, stat = self._zookeeper.get(
-                '{path}/{slug}'.format(
-                    path=self._topic_path, slug=partition_slug))
-            if owner_id == get_bytes(self._consumer_id):
-                zk_partition_ids.add(int(partition_slug.split('-')[1]))
+            try:
+                owner_id, stat = self._zookeeper.get(
+                    '{path}/{slug}'.format(
+                        path=self._topic_path, slug=partition_slug))
+                if owner_id == get_bytes(self._consumer_id):
+                    zk_partition_ids.add(int(partition_slug.split('-')[1]))
+            except NoNodeException:
+                pass  # disappeared between ``get_children`` and ``get``
         return set(self._topic.partitions[_id] for _id in zk_partition_ids)
 
     def _check_held_partitions(self):
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -442,14 +442,17 @@ class BalancedConsumer():
                     new_partitions = self._decide_partitions(participants)
 
                     # Update zk with any changes:
-                    old_partitions = self._get_held_partitions()
-                    self._remove_partitions(old_partitions - new_partitions)
-                    self._add_partitions(new_partitions - old_partitions)
+                    # Note that we explicitly fetch our set of held partitions
+                    # from zk, rather than assuming it will be identical to
+                    # `self.partitions`.  This covers the (rare) situation
+                    # where due to an interrupted connection our zk session
+                    # has expired, in which case we'd hold zero partitions on
+                    # zk, but `self._partitions` may be outdated and non-empty
+                    current_zk_parts = self._get_held_partitions()
+                    self._remove_partitions(current_zk_parts - new_partitions)
+                    self._add_partitions(new_partitions - current_zk_parts)
 
                     # Only re-create internal consumer if something changed.
-                    # Note that `old_partitions` may be different from
-                    # `self._partitions` if our zk session expired (in which
-                    # case the former may be empty while the latter isn't)
                     if new_partitions != self._partitions:
                         self._setup_internal_consumer(list(new_partitions))
 
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -202,7 +202,8 @@ class BalancedConsumer():
                 time.sleep(120)
                 if not self._running:
                     break
-                self._check_held_partitions()
+                if not self._check_held_partitions():
+                    self._rebalance()
             log.debug("Checker thread exiting")
         log.debug("Starting checker thread")
         return self._cluster.handler.spawn(checker)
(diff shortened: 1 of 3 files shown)


pykafka/balancedconsumer.py
```diff
@@ -460,22 +473,16 @@ def _path_from_partition(self, p):
...
 
-        Also remove these partitions from the consumer's internal
-        partition registry.
-
         :param partitions: The partitions to remove.
         :type partitions: Iterable of :class:`pykafka.partition.Partition`
         """
         for p in partitions:
-            assert p in self._partitions
+            # TODO pass zk node version to make sure we still own this node
```
yungchin: Couldn't make up my mind about whether that's really worth it - we would have to emit node versions from `_get_held_partitions()` and pass those into this. Maybe a lower-priority thing for a separate ticket?

pykafka/balancedconsumer.py
```diff
@@ -428,17 +434,24 @@ def _rebalance(self):
...
+                    self._remove_partitions(old_partitions - new_partitions)
+                    self._add_partitions(new_partitions - old_partitions)
 
                     # Only re-create internal consumer if something changed.
-                    if old_partitions or new_partitions:
-                        self._setup_internal_consumer()
+                    # Note that `old_partitions` may be different from
+                    # `self._partitions` if our zk session expired (in which
+                    # case the former may be empty while the latter isn't)
+                    if new_partitions != self._partitions:
```
emmett9001: I don't think this line makes sense given that you've removed all mutations to `self._partitions` elsewhere in the file.

pykafka/balancedconsumer.py
```diff
@@ -428,17 +434,24 @@ def _rebalance(self):
...
+                        participants.append(self._consumer_id)
 
-                    old_partitions = self._partitions - partitions
-                    self._remove_partitions(old_partitions)
+                    new_partitions = self._decide_partitions(participants)
 
-                    new_partitions = partitions - self._partitions
-                    self._add_partitions(new_partitions)
+                    # Update zk with any changes:
+                    old_partitions = self._get_held_partitions()
```
emmett9001: Maybe consider changing this variable name to `current_partitions` or something like that.

pykafka/balancedconsumer.py
```diff
@@ -486,41 +493,46 @@ def _add_partitions(self, partitions):
...
-            owner_id, stat = self._zookeeper.get(
-                '{path}/{slug}'.format(
-                    path=self._topic_path, slug=partition_slug))
-            if owner_id == self._consumer_id:
-                zk_partition_ids.add(int(partition_slug.split('-')[1]))
-        # build a set of partition ids we think we own
-        internal_partition_ids = set([p.id for p in self._partitions])
-        # compare the two sets, rebalance if necessary
-        if internal_partition_ids != zk_partition_ids:
+        zk_partitions = self._get_held_partitions()
```
emmett9001: This can probably just be combined with the following line.

pykafka/balancedconsumer.py
```diff
@@ -202,7 +201,8 @@ def checker():
                 time.sleep(120)
                 if not self._running:
                     break
-                self._check_held_partitions()
+                if not self._check_held_partitions():
+                    self._rebalance()
```
emmett9001: This is definitely cleaner than before.

pykafka/balancedconsumer.py
```diff
@@ -428,17 +434,24 @@ def _rebalance(self):
...
+                    old_partitions = self._get_held_partitions()
+                    self._remove_partitions(old_partitions - new_partitions)
+                    self._add_partitions(new_partitions - old_partitions)
 
                     # Only re-create internal consumer if something changed.
-                    if old_partitions or new_partitions:
-                        self._setup_internal_consumer()
+                    # Note that `old_partitions` may be different from
+                    # `self._partitions` if our zk session expired (in which
+                    # case the former may be empty while the latter isn't)
```
emmett9001: I think this is an important comment. Is this just an explanation of the use of `new_partitions` here? It might be more clearly stated as "if this is the first rebalance of a new zk session" or simply "if there is no node for this consumer in zookeeper".

pykafka/balancedconsumer.py
emmett9001: I like the idea of making pykafka znode-version-aware. What would we do here, though? I guess we'd raise an error if we found that we didn't own the node? Warn and pass? Trigger a new rebalance?
It's definitely not for this pull request, but it seems worthwhile to explore.

tests/pykafka/test_balancedconsumer.py
```diff
@@ -39,7 +39,7 @@ def setUpClass(cls):
     def test_consume_returns(self):
         """Ensure that consume() returns in the amount of time it's supposed to
         """
-        self._mock_consumer._setup_internal_consumer(start=False)
+        self._mock_consumer._setup_internal_consumer([], start=False)
```
emmett9001: I half think that this `[]` should be a kwarg (as `None`, of course) to avoid code like this, but maybe it doesn't matter since it's private anyway.

tests/pykafka/test_balancedconsumer.py
```diff
@@ -195,6 +201,36 @@ def test_external_kazoo_client(self):
...
+    def test_zk_conn_lost(self):
+        """Check we restore zookeeper nodes correctly after connection loss
+
+        See also github issue #204.
+        """
+        zk = KazooClient(self.kafka.zookeeper)
+        zk.start()
+        try:
+            topic = self.client.topics[self.topic_name]
+            consumer = topic.get_balanced_consumer(b'test_zk_conn_lost',
```
emmett9001: Consumer group should go into a variable.

tests/pykafka/test_balancedconsumer.py
```diff
@@ -195,6 +201,36 @@ def test_external_kazoo_client(self):
...
+                                                   zookeeper=zk)
+            self.assertTrue(consumer._check_held_partitions())
+            zk.stop()  # expires session, dropping all our nodes
+
+            # Start a second consumer on a different zk connection
+            other_consumer = topic.get_balanced_consumer(b'test_zk_conn_lost')
+
+            zk.start()
```
emmett9001: Does this actually trigger one of the watches that `consumer` has registered? If so, I totally understand what this test is doing, but I can't see exactly which watch it would trigger.

tests/pykafka/test_balancedconsumer.py
```diff
@@ -195,6 +201,36 @@ def test_external_kazoo_client(self):
...
+            time.sleep(.3)  # allow consumers time to begin rebalancing
+            with consumer._rebalancing_lock:  # wait until rebalancing finishes
+                self.assertTrue(consumer._check_held_partitions())
+            with other_consumer._rebalancing_lock:
+                self.assertTrue(other_consumer._check_held_partitions())
+        finally:
+            try:
+                consumer.stop()
+                other_consumer.stop()
```
emmett9001: I think you may want to `zk.stop()` here.

emmett9001:
This is a really exciting changeset. Using `self._partitions` as the source of truth was a bad idea of mine that originally stemmed from lack of familiarity with zookeeper, and this change actually simplifies the rebalancing process a bit.
The new test actually hangs my local `tox` when I run it against all interpreters:
```
tox tests/pykafka/test_balancedconsumer.py::BalancedConsumerIntegrationTests::test_zk_conn_lost
```
I thought it had to do with the zookeeper connection being left open, but that doesn't actually appear to fix it. Aside from that, I left a bunch of comments about small things. Once those are addressed, I can test this out and hopefully get it merged. Nice job!

yungchin:
Thanks for looking at this - again so quickly!

> The new test actually hangs my local tox when I run it against all interpreters:

Is that only when the test fails (ie at 8faf6c1) or also after applying the fixes? If it's with the fixes, I've no idea why that happens and would be really worried - it doesn't happen here.

If it's at 8faf6c1, what I get here is that the "FAILED" shows up, and the traceback and stderr logging, but then the test tear-down takes a really long time to exit. The reason is that there's also an exception in the `finally` clause of the test (which I've hidden with the `except: pass` so I could see the original exception), and that prevents the consumer from getting destructed. The failure happens because in `stop()` we're trying to remove nodes that don't exist after the `zk.stop() zk.start()` cycle.

tests/pykafka/test_balancedconsumer.py
yungchin: Yes, `ChildrenWatch` internally registers a listener on the zk connection state, so that all three watches trigger when the connection comes back (so that also means we start a `_rebalance` three times instead of just once, but apart from hammering the zk cluster that's probably harmless?).

pykafka/balancedconsumer.py
yungchin: Yeah, I was really struggling with the wording of this. What I mean to highlight is that this line is not the same as
```python
if new_partitions != old_partitions:
```
.... even though 99% of the time it is actually the same - it's only different after we recover from a zk connection loss. Hmm, does that get closer to what I should have written? I'll have another go at it now.

pykafka/balancedconsumer.py
yungchin: Hah, cool, wasn't intended as an improvement, just that I'm lazy so I wanted to reuse that method for the test :)

pykafka/balancedconsumer.py
yungchin: I'm not sure - I mostly thought of this because I had to remove the `assert p in self._partitions` here, and wanted something similar in its place - so if we'd follow the original behaviour of the `assert`, then it would raise an exception here.

I think that's probably also the safest option: it would mean the rebalance would abort, and a new rebalance would probably be underway anyway (because anything that caused this loss of ownership would trigger the watches?), and the new rebalance would reset everything we think we know about which nodes we own? .... ugh, that was the longer version of "I'm not sure" :)

pykafka/balancedconsumer.py
yungchin: Was just about to make this change when I realised I forgot to update L525 below, which should be, but wasn't, reusing this var.

yungchin:
@emmett9001 ok, I think all updates are in. I'm curious/anxious about the hangs in tox you mentioned - let me know if there's any more I can check on this end.

emmett9001:
The only remaining issue here is that the new test often hangs. Once that's resolved, we should be able to merge this.

yungchin:
@emmett9001 the commit I just added removes the hanging on my box.

(For everyone else's understanding: we discussed this elsewhere, and established that the test hangs only happen if you test without setting up a kafka cluster first. The travis-ci config sets up the cluster before starting the test suite, which is why it all passed fine on travis-ci.)

pykafka/balancedconsumer.py
```diff
@@ -261,7 +271,7 @@ def _setup_zookeeper(self, zookeeper_connect, timeout):
         self._zookeeper = KazooClient(zookeeper_connect, timeout=timeout / 1000)
         self._zookeeper.start()
 
-    def _setup_internal_consumer(self, start=True):
+    def _setup_internal_consumer(self, partitions=[], start=True):
```
emmett9001: We shouldn't use mutable kwargs - I think `partitions=None` and some logic in the function to set it to `[]` in that case should do the trick. I can add that.

yungchin: Ouch, thanks for catching this - I'm a repeat offender with this one. Somehow my brain and python really don't agree on the semantics of kwargs :/

PR #310, opened 2015-10-21
handlers: fix RequestHandler queue flush on stop

### LINKED ISSUE #309: race condition in RequestHandler?
I believe there to be a possible race condition in RequestHandler which could cause lockage during GC of the object. Specificically, if the RequestHandler object is cleaned up before the final task is marked as done [here](https://github.com/Parsely/pykafka/blob/master/pykafka/handlers.py#L154), then the thread will block [here](https://github.com/Parsely/pykafka/blob/master/pykafka/handlers.py#L130). A possible manifestation of this issue is sometimes occurring on creation of a client.
* Broker is created [here](https://github.com/Parsely/pykafka/blob/master/pykafka/cluster.py#L212) and sets up a request shortly thereafter.
* That broker soon goes out of scope, so the only reference 
 to its handler is the weakref in the request thread.
* The race condition occurs when a ReferenceError is thrown before or when the task is marked as done, meaning that the
 join call [here](https://github.com/Parsely/pykafka/blob/2.0.0/pykafka/handlers.py#L130) blocks forever.

The issue is difficult to reproduce, but I've seen it occur a number of times in our testing. I'm not really sure how one would go about fixing it, but if anyone can come up with a suggestion, I would be happy to implement and test.

yungchin:
This resolves #309 by removing the weakref solution introduced in #258,
which would potentially throw a ReferenceError before the worker was
quite done emptying the request queue.

### DIFF
--- a/pykafka/handlers.py
+++ b/pykafka/handlers.py
@@ -103,7 +103,7 @@ class RequestHandler(object):
         # _start_thread(), so be careful not to rebind it
         self.shared = self.Shared(connection=connection,
                                   requests=handler.Queue(),
-                                  ending=False)
+                                  ending=handler.Event())
 
     def __del__(self):
         self.stop()
@@ -133,7 +133,7 @@ class RequestHandler(object):
         self.shared = None
         log.info("RequestHandler.stop: about to flush requests queue")
         shared.requests.join()
-        shared.ending = True
+        shared.ending.set()
 
     def _start_thread(self):
         """Run the request processor"""
@@ -144,7 +144,7 @@ class RequestHandler(object):
         shared = self.shared
 
         def worker():
-            while not shared.ending:
+            while not shared.ending.is_set():
                 try:
                     # set a timeout so we check `ending` every so often
                     task = shared.requests.get(timeout=1)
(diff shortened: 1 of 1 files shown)


yungchin:
Not to be too much of a codecov cheerleader, but their patch-coverage feature just allowed me to spot a bug that was hidden from the test suite: https://codecov.io/github/Parsely/pykafka/compare/f0f12a3...2d4a35a showed that there's only partial branch coverage for the `while not shared.ending:` line, which was due to me introducing a bug in the finaliser (and exceptions in finalisers are suppressed, so it went otherwise unnoticed).

@stevepeak

stevepeak:
:+1: Thank you @yungchin Very happy to provide this solution for you!

emmett9001:
@yungchin this looks good. I'm pretty happy with any fix that removes a `weakref`, as those generally make the code harder to reason about. I spent a bit trying to come up with a better name for `shared` that's more descriptive of what it's doing, but I can't.

yungchin:
@emmett9001 thanks as ever for the speediness!! I actually think that `shared` is the worst name possible, but still couldn't think of a better one. It's like they say http://martinfowler.com/bliki/TwoHardThings.html

PR #311, opened 2015-10-22
Register atexit handlers for balancedconsumer and producer

rogaha:
This makes sure that the ``stop()`` method is called properly at ``program termination``.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -35,6 +35,7 @@ from .exceptions import (KafkaException, PartitionOwnedError,
                          ConsumerStoppedException, ZookeeperConnectionLost)
 from .simpleconsumer import SimpleConsumer
 from .utils.compat import range, get_bytes, itervalues
+from .utils.common import unregister_cleanup_func, register_cleanup_func
 
 
 log = logging.getLogger(__name__)
@@ -189,6 +190,13 @@ class BalancedConsumer():
         if auto_start is True:
             self.start()
 
+        def cleanup(obj):
+            obj.stop()
+        self._cleanup_func = cleanup
+
+        # Register a cleanup handler
+        register_cleanup_func(self._cleanup_func, self)
+
     def __repr__(self):
         return "<{module}.{name} at {id_} (consumer_group={group})>".format(
             module=self.__class__.__module__,
@@ -269,6 +277,11 @@ class BalancedConsumer():
             # additionally we'd want to remove watches here, but there are no
             # facilities for that in ChildrenWatch - as a workaround we check
             # self._running in the watcher callbacks (see further down)
+
+        # Unregister the cleanup handler
+        unregister_cleanup_func(self._cleanup_func, self)
+
+        del self._cleanup_func
 
     def _setup_zookeeper(self, zookeeper_connect, timeout):
         """Open a connection to a ZooKeeper host.
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -39,6 +39,7 @@ from .exceptions import (
 from .partitioners import random_partitioner
 from .protocol import Message, ProduceRequest
 from .utils.compat import iteritems, range, itervalues
+from .utils.common import unregister_cleanup_func, register_cleanup_func
 
 log = logging.getLogger(__name__)
 
@@ -137,6 +138,14 @@ class Producer(object):
         self._update_lock = self._cluster.handler.Lock()
         self.start()
 
+        def cleanup(obj):
+            if obj._running:
+                obj.stop()
+        self._cleanup_func = cleanup
+
+        # Register a cleanup handler
+        register_cleanup_func(self._cleanup_func, self)
+
     def _raise_worker_exceptions(self):
         """Raises exceptions encountered on worker threads"""
         if self._worker_exception is not None:
@@ -217,6 +226,10 @@ class Producer(object):
         if self._owned_brokers is not None:
             for owned_broker in self._owned_brokers.values():
                 owned_broker.stop()
+        # Unregister the cleanup handler
+        unregister_cleanup_func(self._cleanup_func, self)
+
+        del self._cleanup_func
 
     def produce(self, message, partition_key=None):
         """Produce a message.
--- a/tests/pykafka/test_producer.py
+++ b/tests/pykafka/test_producer.py
@@ -2,12 +2,20 @@ from __future__ import division
 
 import time
 import unittest2
+import mock
 from uuid import uuid4
 
 from pykafka import KafkaClient
 from pykafka.exceptions import MessageSizeTooLarge, ProducerQueueFullError
 from pykafka.protocol import Message
 from pykafka.test.utils import get_cluster, stop_cluster
+
+
+class MockAtexit(mock.MagicMock):
+    _exithandlers = []
+
+    def unregister(self, func):
+        raise ValueError
 
 
 class ProducerIntegrationTests(unittest2.TestCase):
@@ -156,6 +164,30 @@ class ProducerIntegrationTests(unittest2.TestCase):
         prod.produce(b"")  # empty string should be distinguished from None
         self.assertEqual(b"", self.consumer.consume().value)
 
+    def test_cleanup_stop_is_not_called_on_stopped_object(self):
+        """Test that the stop() method is not called twice"""
+        prod = self.client.topics[self.topic_name].get_producer()
+        prod.stop()
+        self.assertFalse(hasattr(prod, '_cleanup_func'))
+
+    def test_cleanup_stop_is_called_on_not_stopped_object(self):
+        """Test that the cleanup() method flushes the queue properly"""
+        payload = uuid4().bytes
+
+        prod = self.client.topics[self.topic_name].get_producer()
+        prod.produce(payload)
+        # simulates a program termination
+        prod._cleanup_func(prod)
+        # confirm that message was received
+        message = self.consumer.consume()
+        self.assertEqual(message.value, payload)
+
+    @mock.patch('pykafka.utils.common.atexit', new=MockAtexit())
+    def test_failing_to_remove_cleanup_func(self):
+        """Test ValueError is ignored """
+        prod = self.client.topics[self.topic_name].get_producer()
+        prod.stop()
+        self.assertFalse(hasattr(prod, '_cleanup_func'))
 
 if __name__ == "__main__":
     unittest2.main()
--- a/pykafka/utils/common.py
+++ b/pykafka/utils/common.py
@@ -1,0 +1,23 @@
+import atexit
+
+
+def unregister_cleanup_func(cleanup, obj):
+    """Unregister the cleanup handler"""
+    if hasattr(obj, '_cleanup_func'):
+        # Remove cleanup handler now that we've stopped
+        try:
+            # py3 supports unregistering
+            if hasattr(atexit, 'unregister'):
+                atexit.unregister(cleanup) # pylint: disable=no-member
+            # py2 requires removing from private attribute...
+            else:
+                # ValueError on list.remove() if the exithandler no longer exists
+                # but that is fine here
+                atexit._exithandlers.remove((cleanup, (obj,), {}))
+        except ValueError:
+            pass
+
+
+def register_cleanup_func(cleanup, obj):
+    """Register a cleanup handler"""
+    atexit.register(cleanup, obj)
--- a/tests/pykafka/test_balancedconsumer.py
+++ b/tests/pykafka/test_balancedconsumer.py
@@ -31,6 +31,13 @@ def buildMockConsumer(num_partitions=10, num_participants=1, timeout=2000):
                             consumer_timeout_ms=timeout), topic
 
 
+class MockAtexit(mock.MagicMock):
+    _exithandlers = []
+
+    def unregister(self, func):
+        raise ValueError
+
+
 class TestBalancedConsumer(unittest2.TestCase):
     @classmethod
     def setUpClass(cls):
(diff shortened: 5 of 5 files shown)


rogaha:
PR rebased.

rogaha:
Can you please re-run the tests? It looks like that tests passed for python versions ``2.7, 3.4 and pypy``, but timeout for python ``3.5``.

rogaha:
All tests are passing now!

rduplain:
Thanks for this pull request! Floor is open for discussion.

yungchin:
@rogaha Thanks very much for this. It's definitely addressing something that's been a problem for quite a while now - the fact that if you forget to issue a `stop()` anywhere at all, it would never be called (one issue with that being that you wouldn't know if your offsets got committed, another that a stale consumer might continue with auto-committing offsets that overwrite your newer consumer's offsets).

I had a go at trying to break reference cycles between the producer/consumer and their worker threads  at some point, and at the time that wasn't sufficient to deal with it (because there were further reference cycles). However, we've since cleaned up a few ref cycles (see #258, #262, and 583087bb25), and so I wonder if it would be doable now.

Would you be interested in looking into that as an alternative solution? There would be a benefit in that it would allow instances to stop as they go out of scope, rather than lingering (and continuing to auto-commit) until program exit.

I do think your implementation in this pullreq is very neat - also careful to remove registered callbacks during `stop()` - and it would perhaps be nice to bring it under `pykafka.utils` (that is, as an external facility, instead of as internal consumer/producer logic) for users who want to be absolutely certain that instances are stopped cleanly. (One reason not to make it internal to the producer/consumer is that the `atexit` handler holding a reference to `self` will always keep the reference count at at least 1 until program exit.)

Thanks!

rogaha:
Hi @rduplain, we are using ``python-kafka`` library in production for producers and ``pykafka`` for consumers due the ``BalancedConsumer`` feature! You guys did a great job on that. The missing piece for us to move completely to ``pykafka`` is in this PR. It makes sure the queue is flushed at ``program termination``.

rogaha:
Hi @yungchin, thanks for the feedback! It makes a lot of sense to move that to ``utils``, I'm open for alternatives, but this approach seems to be working well, we have been using in production for a while. 

I'll move that code to ``utils``, thanks again for reviewing the PR.

rogaha:
@yungchin PR updated!

rogaha:
@yungchin just to let you know, I tested the context manager ``__exit__`` & ``__enter__``, but it doesn't give us the same guarantees as the ``atexit`` approach.

tests/pykafka/test_producer.py
```diff
@@ -118,6 +126,30 @@ def test_null_payloads(self):
...
         self.assertEqual(b"", self.consumer.consume().value)
 
+    def test_cleanup_stop_is_not_called_on_stopped_object(self):
+        """Test that the stop() method is not called twice"""
+        prod = self.client.topics[self.topic_name].get_producer()
+        prod.stop()
+        self.assertFalse(hasattr(prod, '_cleanup_func'))
+
+    def test_cleanup_stop_is_called_on_not_stopped_object(self):
+        """Test that the cleanup() method flushes the queue properly"""
```
yungchin: I really like this docstring putting it in these terms. Even though this is implicitly also tested in `test_async_produce_context`, it is actually a requirement that deserves the emphasis given here.

pykafka/balancedconsumer.py
```diff
@@ -259,6 +268,11 @@ def stop(self):
             # facilities for that in ChildrenWatch - as a workaround we check
             # self._running in the watcher callbacks (see further down)
 
+        # Unregister the cleanup handler
+        unregister_cleanup_func(self._cleanup_func, self)
```
yungchin: One nitpick would be that this stores a reference to `self` elsewhere, which would prevent garbage collection. That's a bit academic currently, because we already have other reasons why gc doesn't happen, but this would add one more.

yungchin:
@rogaha thanks very much for the big refactor! For my understanding of the results you're after, can you expand a bit on what you see when you say "it doesn't give us the same guarantees as the atexit approach"?

@kbourgoin @emmett9001 what are your thoughts on enabling this by default on the producer/consumer? I think there's definitely a lot of value in having this facility under `utils.common`, making it easy to register your instances, but my concern with enabling it all the time is that it would mask the finalisation-issues that we'd still have. I note that even with this patch in place an old consumer, after going out of scope, would continue to auto-commit potentially stale offsets regularly.

To be clear, I do think being able to ensure clean finalisation at program exit is a big improvement over the status quo, and the pullreq looks good. I just wonder if we shouldn't address the root cause of the problems first.

emmett9001:
@yungchin @rogaha before reading the diff, my take on these comments is that it's probably preferable to improve the behavior of the producer's context manager, since this issue (terminating producer worker threads) is exactly what that context manager is meant to solve. What is the context manager not doing now in master that it needs to (what are the "guarantees"?)

pykafka/producer.py
```diff
@@ -140,6 +141,14 @@ def __init__(self,
         self._update_lock = self._cluster.handler.Lock()
         self.start()
 
+        def cleanup(obj):
+            if obj._running:
+                obj.stop()
```
emmett9001: Since this function is nearly identical to `self.__exit__`, it seems that we could solve this issue by simply using `with` in the client code. Am I missing something?

yungchin:
Tangentially related to this... I took another shot at making `SimpleConsumer` gc-able, and it turns out I just missed one thing in the previous attempt: that we store closures over `self` in `self._default_error_handlers`. So [throwing in some weakrefs][0] is now sufficient to make instances get gc'd automatically as they go out of scope. I could go see if the same sort of treatment works in `Producer` and `BalancedConsumer`. @rogaha do you think that that fix could provide sufficient guarantees for your use cases? 

[0]: https://github.com/Parsely/pykafka/commit/a65acdcb1b8e23e8739d4a6bd9df0ba972d8e5fd

pykafka/producer.py
rogaha: @emmett9001 I had missed this part, so it seems that ``self.__exit__`` behaves the same way when we instantiate the class using ``with``, but it would require a refactor of our client code to use that, we use the same class instance in different parts of the code.

rogaha: Would it work? 
<pre>
def get_kafka_async_producer(topic_name):
    """
    Return a real kafka producer for dev, stage and prod. For unit-tests it returns None
    """
    try:
        with KafkaProducer(topic_name, async=True, async_queue_maxsize=settings.KAFKA_ASYNC_QUEUE_MAXSIZE) as producer:
            return producer
    except KafkaUnavailableError:
        log.exception("Connection to Kafka is unavailable")
        return None
</pre>

In the ``KafkaProducer()`` class we instantiate the pykafka ``SimpleProducer()`` class.

rogaha:
@emmett9001 @yungchin sorry about the delayed reply, I was on vacation. So, I ran the experiment without using ``with``, I was not familiar with the ``context manager`` usage. It seems to work similarly as the ``atexit`` approach.

rogaha:
@yungchin thanks for adding those extra handlers.

rogaha:
@emmett9001 I'll remove the ``atexit`` implementation since the ``context manger`` approach already solve the issue. Should I refactor the PR and implement the ``context manager`` for the ``balanced consumer`` as well?

emmett9001:
@rogaha if you'd like to, please do. A context manager for the `BalancedConsumer` is something we'd like to add either way.

rogaha:
Sounds good. I'll close this PR since we are not going to use the ``atexit`` approach! Thanks @yungchin for reviewing the PR.

rogaha:
I'm reopening it, the ``context manager`` didn't work for my use case. It would require opening a new connection for each message. Basically we instantiate the ``Producer`` at beginning and we use the same connection to send the messages as they arrive by calling ``kafka_producer.send(paylod)``. 

The only way it worked well was using the ``atexit`` approach. I think we can keep both approaches since they are not conflicting with each other.

pykafka/balancedconsumer.py
rogaha: I see, how do you think we could do that to avoid it?

tests/pykafka/test_producer.py
rogaha: Thanks @yungchin

rogaha:
PR rebased.

pykafka/balancedconsumer.py
yungchin: One way would be to have `register_cleanup_func` wrap `self` in a `weakref.proxy` and wrap `self._cleanup_func` in a thin wrapper that just catches and ignores `ReferenceError`. That makes it all awfully ugly to read and understand, but I don't know a better option for this.

yungchin:
Thanks for expanding on the use case @rogaha. Basically then, for use cases where the context manager is cumbersome, we currently do a bad job because `stop()` will never run (not even on interpreter exit).

I would like to pursue #329 first, which I think would help handle your use case, and has an advantage over the `atexit` approach, in that cleanups may happen as soon as an instance goes out of scope, rather than only on interpreter exit. I'll try to push ahead with that, so you can hopefully see if it's sufficient for your uses soon. Thanks!

yungchin:
@rogaha I believe #329 is in a workable state, and I'd be grateful if you would have time to take it for a spin and see if it works for your uses.

I should note that #329 relies on the interpreter, under normal circumstances, running `__del__()` before exiting. If and when `__del__()` is called is however not strictly guaranteed, and so it may not offer as strong a guarantee that `stop()` is always called as the `atexit` approach.

bcwalrus:
@emmett9001, et al -- I have a basic question about how people commonly use the Producer's context manager. Sounds like the context manager requires instantiating a new Producer for every (batch of) message to send. Right? Is that a bit heavy for normal use?

Would it be easier to create *just one* Producer, keep it around, and have it send stuff when I have stuff to send? How should I use the context manager in this case?

rogaha:
@yungchin, I think ``__del__()`` doesn't give the same guarantees as ``atexit``. Is there any side effect of using ``atexit`` on your point of view?

pykafka/producer.py
emmett9001: Unfortunately I don't think that method will work. The `with` block will stop the producer as soon as the function returns.

rogaha: Yeap, indeed!  :)

emmett9001:
@rogaha @bcwalrus I don't understand the use case that requires a new `Producer` to be instantiated for each message batch when the context manager is in use. It was written to enable code like this:
```python
with topic.get_producer() as producer:
    while True:
        producer.produce('whatever')
```
This code allows a single producer to remain active while it sends an arbitrarily long string of messages. Under what circumstances are you required to make a new producer for each message or message batch?

I'm opposed to using `atexit` handlers because they assume that the pykafka components will always be running in the main thread. This is not a restriction that we want to place on users of the library, since we've seen use cases where people start multiple consumers in their own threads.

The preferred way to use the producer if you need to share one instance across multiple code blocks is to call `stop()` manually before the program exits. `stop()` has always been part of the `Producer`'s public API and is meant to be used this way. Is there a reason that `atexit` is preferable to simply calling `stop()` prior to interpreter shutdown?

yungchin:
@rogaha I certainly agree `atexit` gives you stronger guarantees at exit than just `__del__`. Nevertheless, I find that `__del__` usually works fine (as long as one's careful to avoid any circular references) and it delivers the most intuitive behaviour (where with "intuitive" I guess I mean "pythonic").

I'd like to see that we can write `p = my_topic.get_producer()` in any context, and that we may then expect that soon after `p` goes out of scope, it gets flushed, stopped and deleted. That would make it most similar to `f = open("filename")`: that is also usually taught with the caveat that one should always `f.close()` (just like `p.stop()`) or that if possible one should use `with open("x") as f:`, but nevertheless few people heed that advice and everyone sort of expects that `f` gets closed anyway after it goes out of scope.

Having `atexit` baked into the class breaks that expectation, because the registry will keep the instance alive - so with that, `p` no longer gets flushed and stopped if it goes out of scope. If you register it yourself, because you want those extra-strong guarantees, then it's different, because it's a conscious choice and so you won't expect automatic garbage-collection to work. That would only add an extra line when you create your instance:
```
producer = topic.get_producer()
register_cleanup_func(producer)
```
(where `register_cleanup_func` is a slightly different implementation of what you have here). This is roughly what I meant with having it as an external facility.

bcwalrus:
@emmett9001 I'd like to leave `atexit` out of our discussion. That's just means to an end. My motivation is to understand the `while True` structure you mentioned (which is also what's in the examples).

Context: I have a Django app, and doesn't control the `main`.

Suppose I do that `while True` loop. *How does my application logic tell the producer what to produce?* I'd probably need a queue that the producer reads from. And my application code would put things into the queue. Now I have another queue that I need to manage and drain. And I have extra locks and cond vars associated with this queue, that eats compute cycles. I also have to set up a new thread so that the producer can run a `while True` loop. All those seem undesirable. Or is there a better way?

Suppose I don't have a `while True` loop, and I don't set up a thread with a long-lived producer. *Do I have to instantiate a new `Producer` every time my application code wants to send a message?* That seems inefficient, as the setup of a `Producer` looks expensive.

So I think I'm missing something obvious. Do you have any suggestions?

yungchin:
@bcwalrus that doesn't sound like a scenario where you would use a context manager? Indeed, you'd be best off keeping a long-lived producer around.

emmett9001:
@bcwalrus Do you want to share a single `Producer` instance across multiple Django views? If there's a way to pass around in-memory python objects between views, you can just use that. If you're just talking about sending many messages from a single Django view, you can just hold a reference to your `Producer` instance and call `producer.produce()` whenever you'd like to send another message.

Also, I'm currently merging #329 into master, which was also set up to fix the auto-stop issue that prompted this discussion. Maybe give current master a try, it may fix the `atexit` issue in a cleaner way that doesn't require client code changes.

rogaha:
@bcwalrus thanks for the describing side effects of using that ``while true``  loop. I also don't another way of using the ``context manager`` besides having an extra ``queue``.

bcwalrus:
@emmett9001 @yungchin #329 looks promising. I'll looking into it. Thanks for working on that!

Django views are short-lived. So there is no good way to hold a long-lived `Producer`, inside an open context manager. I can certainly use a vanilla `Producer` without a context manager. But I'd lose the auto-cleanup ability.

I *could* spin off a new thread that runs a `while True` loop, and be able to use the context manager. But it's probably not worth it, for reasons stated in my previous comment https://github.com/Parsely/pykafka/pull/311#issuecomment-154572215. (Mainly I have to worry about draining and cleaning up yet another queue.)

PR #315, opened 2015-10-26
Empty partition assignment

### LINKED ISSUE #313: Race condition when creating multi BalancedConsumer
```python
    def _add_self(self):
        """Register this consumer in zookeeper.
        This method ensures that the number of participants is at most the
        number of partitions.
        """
        # NOTE: Race condition here
        if len(self._topic.partitions) <= len(participants):

        self._zookeeper.create(
            self._path_self, self._topic.name, ephemeral=True, makepath=True)
```
As shown above, [there is](https://github.com/Parsely/pykafka/blob/master/pykafka/balancedconsumer.py#L412) a race condition when multi processes create BalancedConsumer at the same time with the same `consumer_group`.

When a race condition occurred, the consumer instance holds [empty partition](https://github.com/Parsely/pykafka/blob/master/pykafka/balancedconsumer.py#L450), and the internal `_consumer ` is [not inited](https://github.com/Parsely/pykafka/blob/master/pykafka/balancedconsumer.py#L464).

Moreover, because of the [null](https://github.com/Parsely/pykafka/blob/master/pykafka/balancedconsumer.py#L250) `_consumer`, there seems to be no way to stop it.

May be we can stop the consumer and raise an exception when [calling _decide_partitions](https://github.com/Parsely/pykafka/blob/master/pykafka/balancedconsumer.py#L450) returns empty partition?

emmett9001:
This pull request causes a `BalancedConsumer` that has no assigned partitions to immediately `stop()` itself. It fixes #313.

### DIFF
--- a/tests/pykafka/test_balancedconsumer.py
+++ b/tests/pykafka/test_balancedconsumer.py
@@ -7,6 +7,7 @@ from kazoo.client import KazooClient
 
 from pykafka import KafkaClient
 from pykafka.balancedconsumer import BalancedConsumer, OffsetType
+from pykafka.exceptions import NoPartitionsForConsumerException
 from pykafka.test.utils import get_cluster, stop_cluster
 from pykafka.utils.compat import range
 
@@ -210,8 +211,9 @@ class BalancedConsumerIntegrationTests(unittest2.TestCase):
                 auto_start=False)
         consumer._decide_partitions = lambda p: set()
         consumer.start()
-        time.sleep(1)
         self.assertFalse(consumer._running)
+        with self.assertRaises(NoPartitionsForConsumerException):
+            consumer.consume()
 
 
     def test_zk_conn_lost(self):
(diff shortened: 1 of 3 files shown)


tests/pykafka/test_balancedconsumer.py
```diff
@@ -201,6 +201,19 @@ def test_external_kazoo_client(self):
...
+
+    def test_no_partitions(self):
+        """Ensure a consumer assigned no partitions immediately exits"""
+        consumer = self.client.topics[self.topic_name].get_balanced_consumer(
+                b'test_no_partitions',
+                zookeeper_connect=self.kafka.zookeeper,
+                auto_start=False)
+        consumer._decide_partitions = lambda p: set()
+        consumer.start()
+        time.sleep(1)
```
yungchin: Because the rebalance during `start()` runs on the same thread, it also works without this sleep. Not that it adds anything noticeable to the time our integration tests take anyway ;)

yungchin:
This looks good to me. I've one issue I encountered while testing it, which is its own bug in a way, but also ties in with this one: if I take the steps that you wrote in the test, and then try to `consume()`, I get an `AttributeError: 'NoneType' object has no attribute 'consume'`. I suppose that should have been a `ConsumerStoppedException` at least, but maybe we could define a specific exception for this situation, so the user can tell why it stopped (something like a `class TooManyConsumers(ConsumerStoppedException):` maybe)?

PR #317, opened 2015-10-27
Suspend internal consumer during rebalancing and zk connection lapses

### LINKED ISSUE #273: BalancedConsumer.consume() should block/fail during rebalance or if zk connection is pending/dead
This comes out of #206. Clearly there are use cases where, if the zookeeper connection goes away and so you can't know if you still own your partitions, you should not be consuming from them. That's probably the safest option, ie to block/fail completely, and is what's the core of this commit (which would need redoing because it depends on a lot of other commits on that branch): https://github.com/Parsely/pykafka/commit/3b23fadeacfec18b91d22aaf121a349a329d02cc

yungchin:
This resolves #273, borrowing a number of ideas initially proposed by @emmett9001 and @fortime in #206 and #215. On current master, the internal consumer will be left running until the very end of the rebalancing process, which could potentially stretch out over several retries (and which may fail, yet leaving the internal consumer live, auto-committing offsets on partitions it no longer owns).

With these changes, we suspend the internal consumer as soon as it becomes clear that it no longer holds the right set of partitions, and only then start further zookeeper back-and-forths to effect the rebalance.

The second change is that this adds a zk connection-status listener, so that if we "lose sight" of our zk nodes, we again suspend the internal consumer. (When the connection comes back, this will always trigger our watches, and thus a rebalance, so that didn't need implementing.)

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -324,6 +324,8 @@ class BalancedConsumer():
         something more sophisticated here.
         """
         if self._consumer is not None:
+            log.debug(
+                "Suspending internal consumer ({})".format(self._consumer_id))
             self._consumer.stop()
 
     def _decide_partitions(self, participants):
@@ -557,9 +559,11 @@ class BalancedConsumer():
         return True
 
     def _get_zk_state_listener(self):
+        """Callback to suspend internal consumer when zk connection drops"""
         ref = weakref.ref(self)
 
         def listener(zk_state):
+            log.info("zk_state_listener: {}".format(zk_state))
             self = ref()
             if self is None:
                 return
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -209,7 +209,8 @@ class BalancedConsumer():
 
     @property
     def partitions(self):
-        return self._consumer.partitions if self._consumer else None
+        return self._consumer.partitions if (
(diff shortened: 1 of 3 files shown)


yungchin:
@emmett9001 I'm not completely sure if it's very pretty to introduce a new exception here, but I was looking for some way to inform the user when a `consume()` is attempted while the zk connection is down. Hope this makes sense!

pykafka/balancedconsumer.py
```diff
@@ -209,7 +213,8 @@ def checker():
 
     @property
     def partitions(self):
-        return self._consumer.partitions if self._consumer else None
+        return self._consumer.partitions if (
+            self._consumer and self._consumer._running) else None
```
emmett9001: I'm not sure I like checking `self._consumer._running` here. The consumer can still be holding partitions even if it's not running (though they may not match zookeeper). I'm pretty sure this check is what's making the tests fail after my master merge, too.

emmett9001:
@yungchin I had some concerns about how this works with current master, but I believe I've figured them out with my handful of commits on this branch. I like this approach a lot, since it feels like one of the cleanest possible ways to do what we're trying to do. I instinctively want to test it manually before merging, but thinking about it, I'm not sure I'd be doing anything the integration tests aren't already doing.

emmett9001:
Tested locally and all looks good!

Issue #318, opened 2015-10-27
KeyError in commit_offsets


emmett9001:
The following stacktrace was encountered when our internal Kafka server was overloaded by many concurrent consumers (running `2.0.1`) that were all heavily lagged:

```
KeyError: 12897
  File "kazoo/recipe/watchers.py", line 333, in _get_children
    result = self._func(children)
  File "pykafka/balancedconsumer.py", line 532, in _consumers_changed
    self._rebalance()
  File "pykafka/balancedconsumer.py", line 441, in _rebalance
    self._setup_internal_consumer()
  File "pykafka/balancedconsumer.py", line 273, in _setup_internal_consumer
    self._consumer.stop()
  File "pykafka/simpleconsumer.py", line 286, in stop
    self.commit_offsets()
  File "pykafka/simpleconsumer.py", line 397, in commit_offsets
    partitions_by_id=self._partitions_by_id)
  File "pykafka/utils/error_handlers.py", line 50, in handle_partition_responses
    parts_by_error = build_parts_by_error(response, partitions_by_id)
  File "pykafka/utils/error_handlers.py", line 75, in build_parts_by_error
    owned_partition = partitions_by_id[partition_id]
```
`partitions_by_id` in this case was

```
{
50: <pykafka.simpleconsumer.OwnedPartition object at 0x7f509021b410>, 
55: <pykafka.simpleconsumer.OwnedPartition object at 0x7f509021b550>
}
```
I'm fairly sure that `12897` is not a valid partition ID for this cluster, but it's unclear exactly where that number came from. It's in the `OffsetCommitResponse` itself, so it's technically possible that this points to an error in `utils.unpack_from`. More likely, it seems, is that Kafka returned a bit of corrupt data.

`parts_by_error` contains one key: 7, which indicates a request timeout, likely in the previous iteration of the loop.

yungchin:
Getting a number like `12897` might also point to a reading frame shift (that may not be the correct term in networking jargon, if not I guess I borrowed it from genetics ;)). This could also explain the occurrence of seemingly truncated messages in #312.

[...]

Issue #312, opened 2015-10-23
PyKafka NoMessagesConsumed Errors


jasonrhaas:
PyKafka throws an error occasionally when trying to iterate through messages in a Kafka queue.  I imagine the issue is some consumer or zookeeper timeout issue, since I know that I am not consuming faster than my Kafka queue is being produced to.  I made a simple script to manually iterate through messages and after a while it throws this error:

```
Traceback (most recent call last):
  File "pykafka_test.py", line 50, in <module>
    k.consume()
  File "pykafka_test.py", line 37, in consume
    print self.balanced_consumer.consume().value
  File "/usr/local/lib/python2.7/site-packages/pykafka/balancedconsumer.py", line 575, in consume
    message = self._consumer.consume(block=block)
  File "/usr/local/lib/python2.7/site-packages/pykafka/simpleconsumer.py", line 345, in consume
    self._raise_worker_exceptions()
  File "/usr/local/lib/python2.7/site-packages/pykafka/simpleconsumer.py", line 199, in _raise_worker_exceptions
    raise ex
pykafka.exceptions.NoMessagesConsumedError
```

I imagine its seeing some *null* value or something and just raising that exception.  Is this the desired behavior?  Should it return `None` instead?  In a gist that @amontalenti wrote, he had a code block that did this:

```
        msg = self.consumer.consume()
        if not msg:
            # nothing here
            return
```

This would work if `self.consumer.consume()` was returning `None`, but instead it raises an error.  You can catch this error with `except: NoMessagesConsumed` but I think this error message may be misleading and something else is going on.

Here's a simple script I'm using to test this:

https://gist.github.com/jasonrhaas/2281296a538908cfa85b

amontalenti:
Thanks for reporting this Jason. We discussed this a little today and one of either @kbourgoin or @emmett9001 will look at this soon!

yungchin:
Thanks very much for this. This exception was introduced with ac5fda0d, and indicates that a truncated message was received from kafka. We can probably change how we handle the error (as you've found, it currently aborts the fetcher thread).

I'm hoping @emmett9001 (who wrote the commit I referenced) may be able provide some extra context: would this error be recoverable if we re-fetch from the same offset? That is, is it just a truncated network transmission, or is there really only a partial message stored in kafka?

jasonrhaas:
```
if len(buff) - offset < size:
    break
```

```
if len(messages) == 0 and attempted:
    raise NoMessagesConsumedError()
```
[...]

PR #320, opened 2015-10-28
connection: ensure reading the full Size field

yungchin:
BrokerConnection.response() did not check that it received all four
bytes of the Size field that precedes a ResponseMessage.  The result of
reading fewer bytes would be 1) an exception being raised when trying to
`struct.unpack()` it, and 2) the next caller of response() receiving
bogus data (because the previous call didn't `recv` the rest of its message).

This never tripped up any tests, as it would likely only ever fail on a
heavily loaded network.  It is a good candidate for explaining the kind
of problems reported in issues #318 and #312.

This commit only fixes our reading of the Size field.  Arguably, a more
complete fix would also make sure that the handler, when catching
exceptions from response(), would discard the connection, to prevent
the ensuing bogus data reads.

### DIFF
--- a/pykafka/connection.py
+++ b/pykafka/connection.py
@@ -109,11 +109,15 @@ class BrokerConnection(object):
 
     def response(self):
         """Wait for a response from the broker"""
-        size = self._socket.recv(4)
-        if len(size) == 0:
-            # Happens when broker has shut down
-            self.disconnect()
-            raise SocketDisconnectedError
+        size = b""  # Size => int32
+        while len(size) != 4:
+            r = self._socket.recv(4 - len(size))
+            if len(r) == 0:
+                # Happens when broker has shut down
+                self.disconnect()
+                raise SocketDisconnectedError
+            size += r
         size = struct.unpack('!i', size)[0]
         recvall_into(self._socket, self._buff, size)
+        # Drop CorrelationId => int32
         return buffer(self._buff[4:4 + size])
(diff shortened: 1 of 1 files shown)


pykafka/connection.py
```diff
@@ -109,11 +109,15 @@ def request(self, request):
 
     def response(self):
         """Wait for a response from the broker"""
-        size = self._socket.recv(4)
-        if len(size) == 0:
-            # Happens when broker has shut down
-            self.disconnect()
-            raise SocketDisconnectedError
+        size = b""  # Size => int32
+        while len(size) != 4:
```
emmett9001: This `4` and the `4` on the next line should be put into a variable like `expected_len` or similar, since this code breaks if they're not the same number.

emmett9001:
I guess this case will only ever happen on a cluster that's so heavily loaded that it is unable to send all four bytes of the message size?

yungchin:
I think it's the latter - it all got sent, but the rest of it is still somewhere in an operating system buffer somewhere along the way. While writing the patch I was checking myself against the python docs and the how-to actually mentions this type of bug, in the second-to-last paragraph of this section: https://docs.python.org/2/howto/sockets.html#using-a-socket :)

yungchin:
It just occurred to me, we could also reuse `recvall_into` which we use for fetching the rest of the response, like so:
```python
size = bytearray(4)
recvall_into(self._socket, size, 4)
```
which cleans this code up a lot.

emmett9001:
Thanks for the link, I just learned a lot. It's funny that they mention this exact bug, too. Once the tests pass I'll merge this.

yungchin:
Hah, it's gone already! I'll leave that refactor for another day :)

Thanks!!

PR #321, opened 2015-10-28
raise worker exceptions inside of consume() loop.

### LINKED ISSUE #319: pykafka 2.0.1 silently stalling on disconnect from kafka broker
Hi, 

As suggested by @yungchin here:  https://groups.google.com/forum/#!topic/pykafka-user/zmK5h7Mt2Wo reporting the following: 

Balanced consumer in  case of one kafka broker goes down will silently stuck in version of pykafka 2.0.1, while in version 1.1.1 it throws SocketDisconnectedError in some thread. 

Setup: 2 kafka brokers, two zookeeper nodes; one topic with two partitions; replication factor 2.
```
$ pip show pykafka
---
Name: pykafka
Version: 2.0.1
Location: /usr/local/lib/python2.7/dist-packages
Requires: kazoo, tabulate
```
Code snippet: http://pastebin.com/z0sssrvb
Sample output: 
```
...
735 | :Message1522.                                                                     
                        {0: 767, 1: 735}                                           
768 | :Message1523.                                                         
                        {0: 768, 1: 735}                            
No handlers could be found for logger "kazoo.recipe.watchers" 
```

[and silence]

Thanks

emmett9001:
This pull request fixes #319 by raising `SocketDisconnectedError` in cases where no more messages are fetchable from the kafka cluster. This case arises when the total number of brokers in the cluster drops below the replication factor, causing the `while True` loop in `consume()` to spin indefinitely. This happens because when kafka cannot guarantee proper replication of data, it won't return it in fetch requests. Thus, when the cluster does not have enough nodes in it to meet the topic's replication factor, fetch requests return no messages.


emmett9001:
@tigran-a does this pull request fix the hanging behavior you're seeing?

tigran-a:
yes, now I constantly catch an exception if try to consume the message if one broker went down (does not matter whether if it is back after or not )

PR #322, opened 2015-10-28
Handle broker failure in consumer

### LINKED ISSUE #223: Consumer hanging after a broker gets down

emmett9001:
This pull request fixes #223 by making the `SimpleConsumer` resilient to changes in broker availability.

The two cases it handles are 1) a currently connected broker losing its connection, and 2) an election causing the leaders of some partitions on the consumed topic to change. The first case is handled by updating the cluster metadata in response to a `SocketDisconnectedError` on `fetch()` instead of raising that exception. The second case is handled by updating the cluster metadata in response to a `NotLeaderForPartition` error on `fetch()`. Interestingly, the protocol documentation [words](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ErrorCodes) the description of `NotLeaderForPartition` in a way that implies that it can only be encountered in response to a `ProduceRequest`. Even so, the method in this pull request still appears to work.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -342,9 +342,8 @@ class SimpleConsumer():
             else:
                 timeout = 1.0
 
-        self._raise_worker_exceptions()
-
         while True:
+            self._raise_worker_exceptions()
             if self._messages_arrived.acquire(blocking=block, timeout=timeout):
                 # by passing through this semaphore, we know that at
                 # least one message is waiting in some queue.
@@ -626,7 +625,12 @@ class SimpleConsumer():
                               owned_partition.partition.id,
                               owned_partition.message_count)
 
-        sorted_by_leader = sorted(iteritems(self._partitions_by_leader), key=lambda k: k[0].id)
+        def unlock_partitions(parts):
+            for owned_partition in parts:
+                owned_partition.fetch_lock.release()
+
+        sorted_by_leader = sorted(iteritems(self._partitions_by_leader),
+                                  key=lambda k: k[0].id)
         for broker, owned_partitions in sorted_by_leader:
             partition_reqs = {}
             sorted_offsets = sorted(owned_partitions, key=lambda k: k.partition.id)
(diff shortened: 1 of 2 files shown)


pykafka/simpleconsumer.py
```diff
@@ -646,20 +655,18 @@ def _handle_success(parts):
                         min_bytes=self._fetch_min_bytes
                     )
                 except (IOError, SocketDisconnectedError):
+                    if self._running:
+                        unlock_partitions(iterkeys(partition_reqs))
```
yungchin: I think that if an `owned_partition` was found too full-up in L646 above, it won't end up in `partition_reqs`, and so may remain locked?

pykafka/simpleconsumer.py
```diff
@@ -646,20 +655,18 @@ def _handle_success(parts):
...
-
                 parts_by_error = build_parts_by_error(response, self._partitions_by_id)
                 handle_partition_responses(
                     self._default_error_handlers,
                     parts_by_error=parts_by_error,
                     success_handler=_handle_success)
-                # unlock the rest of the partitions
-                for owned_partition in iterkeys(partition_reqs):
-                    owned_partition.fetch_lock.release()
+                unlock_partitions(iterkeys(partition_reqs))
```
yungchin: The same as in the other comment will apply here. I find it really difficult to track what's locked and what's not, so I'm wondering if maybe - even if that doesn't look very elegant - it would help to just keep an explicit `list()` of fetch-locks held.

yungchin:
@emmett9001 this is surprisingly concise and works well I think - I was expecting it would take a much bigger diff to accomplish all this. I've only the concern over locking noted earlier inline, nothing else.

I wrote a little test while reviewing this, mostly copied over from one I wrote in the producer-futures branch yesterday. It's a bit crude and also not fancy enough to catch real problems such as that locking issue, but it makes sure some of the cluster-update path gets some automatic exercise. I'll push it to this branch but feel free to pop it off again if you hate it - it does add a few seconds to the test suite I think.

pykafka/simpleconsumer.py
emmett9001: As soon as the lock is acquired, the partition is added as a key to `partition_reqs` with `None` as its value. See [these lines](https://github.com/Parsely/pykafka/blob/3026e9c1858ed1844cf66530c05bc8353087a6b6/pykafka/simpleconsumer.py#L635-L636) and the fact that they are right next to each other. I've been working under the assumption that since we don't further mutate that dict until the next loop iteration, we're guaranteed that we're unlocking exactly the set of locked partitions. Maybe the style of this code could use some clarifying, but I don't see a bug in it at least.

yungchin: Ah, yes, sorry! I need to bring my glasses obviously. No bugs here, all good to go :)

Issue #288, opened 2015-10-02
Notification / callback on rebalance?


ryan-stateless:
I have N consumers of M partitions in a Kafka topic. Kafka messages are updates to an in-memory data structure that is initialized from (and periodically written back out to) a 3rd party DB.

I'm trying to figure out a robust solution to the following problem:
Consumer  'A' owns some subset of the partitions. Consumer 'B' comes online, initializes, and takes ownership of one (or more) partitions from 'A' and begins processing messages. Consumer 'B' then goes down and the Consumer 'A' resumes consuming the messages.

From consumer A's perspective, there was a gap in updates for a while, but eventually they resumed. In reality, A has failed to apply a subset of the updates to it's data structure because they were being handled by 'B'

If there were a mechanism to notify Consumer 'A' that it had lost ownership of the partition, it could do the appropriate flushing/retiring of data and likewise resume cleanly when it regained ownership.

Does such a mechanism exist? Is there any reason that such a mechanism wouldn't work cleanly in practice? It seems like I could pass a function pointer to the BalancedConsumer at init() and call that function at the end of _rebalance(). Thoughts? Pitfalls? Worthwhile PR?

emmett9001:
At first blush, a callback after `_rebalance` sounds like it wouldn't be too problematic to implement. As long as we kept it very generic and didn't try to give the callback too much information about any state, I think it would be fine. Sounds like a worthwhile pull request.

I'm wondering about your use case. Does each consumer have its own separate in-memory data structure it's responsible for modifying, or are all of the consumers pushing updates to a single structure? If you want all updates to go to the same structure, it seems like you could use a single consumer or a group of balanced consumers without any issue. If you're modifying multiple structures (one per consumer), I think separate consumer groups would be more suited to your use case.

It's extremely likely I'm misunderstanding something about your use case - please enlighten me!

ryan-stateless:
In my use case, you might as well think of the topic as "user activity updates". The number of users is much greater than the number of partitions. User updates are hashed into partitions by unique user id. Each consumer in the group is responsible for aggregating user updates into an up-to-date user data structure per user. The data structure is large and the update heuristic moderately complex, precluding a database read before each update or an atomic in-database stored procedure style update.

[...]

Issue #274, opened 2015-09-28
Should BalancedConsumer.consume() block/fail during rebalancing?


yungchin:
This was raised by @fortime in comments on #215: "The consumer will consume message during the time of rebalancing. It is an expected behavior, but it may cause messages in 30 seconds consumed by multiple consumers."

emmett9001:
Folding this into #273 since they're really about the same thing: when the zookeeper partition registry is changing or inaccessible, what should the consumer do about it?

PR #324, opened 2015-10-29
rebalance callbacks

emmett9001:
This pull request adds callback notification hooks to the rebalance process. It's related to #288.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -35,7 +35,7 @@ from .common import OffsetType
 from .exceptions import (KafkaException, PartitionOwnedError,
                          ConsumerStoppedException, NoPartitionsForConsumerException)
 from .simpleconsumer import SimpleConsumer
-from .utils.compat import range, get_bytes, itervalues
+from .utils.compat import range, get_bytes, itervalues, iteritems
 
 
 log = logging.getLogger(__name__)
@@ -331,7 +331,7 @@ class BalancedConsumer(object):
         self._zookeeper = KazooClient(zookeeper_connect, timeout=timeout / 1000)
         self._zookeeper.start()
 
-    def _setup_internal_consumer(self, partitions=None, start=True):
+    def _get_internal_consumer(self, partitions=None, start=True):
         """Instantiate an internal SimpleConsumer.
 
         If there is already a SimpleConsumer instance held by this object,
@@ -344,10 +344,10 @@ class BalancedConsumer(object):
         if self._consumer is not None:
             self._consumer.stop()
             # only use this setting for the first call to
-            # _setup_internal_consumer. subsequent calls should not
+            # _get_internal_consumer. subsequent calls should not
             # reset the offsets, since they can happen at any time
             reset_offset_on_start = False
-        self._consumer = SimpleConsumer(
+        return SimpleConsumer(
             self._topic,
             self._cluster,
             consumer_group=self._consumer_group,
@@ -549,7 +549,25 @@ class BalancedConsumer(object):
 
                     # Only re-create internal consumer if something changed.
                     if new_partitions != self._partitions:
-                        self._setup_internal_consumer(list(new_partitions))
+                        cns = self._get_internal_consumer(list(new_partitions))
+                        if self._partitions_revoked_callback is not None:
+                            self._partitions_revoked_callback(
+                                self, self._get_held_offsets_for_partitions(
+                                    current_zk_parts - new_partitions, cns))
+                        if self._partitions_assigned_callback is not None:
+                            new_offsets = self._partitions_assigned_callback(
+                                self,
+                                self._get_held_offsets_for_partitions(
+                                    new_partitions, cns)
+                            )
+                            if new_offsets:
+                                cns.reset_offsets(
+                                    partition_offsets=[
+                                        (part, new_offsets[part.id])
+                                        for part in itervalues(cns.partitions)
+                                        if part.id in new_offsets]
+                                )
+                        self._consumer = cns
 
                     log.info('Rebalancing Complete.')
                     break
@@ -560,12 +578,21 @@ class BalancedConsumer(object):
                         raise
                     log.info('Unable to acquire partition %s. Retrying', ex.partition)
                     time.sleep(i * (self._rebalance_backoff_ms / 1000))
-            if self._partitions_assigned_callback is not None:
-                self._partitions_assigned_callback(self, new_partitions)
-            if self._partitions_revoked_callback is not None:
-                self._partitions_revoked_callback(self, current_zk_parts - new_partitions)
         if should_stop:
             self.stop()
+
+    def _get_held_offsets_for_partitions(self, partitions, cns):
+        """Return held offsets from a consumer for a set of partitions
+
+        :param partitions: The partitions for which to return currently held offsets
+        :type partitions: Iterable of :class:`pykafka.partition.Partition`
+        :param cns: The consumer from which to fetch held offsets
+        :type cns: :class:`pykafka.simpleconsumer.SimpleConsumer`
+        """
+        if cns.held_offsets is None:
+            return {}
+        ids = [partition.id for partition in partitions]
+        return {id_: offset for id_, offset in iteritems(cns.held_offsets) if id_ in ids}
 
     def _path_from_partition(self, p):
         """Given a partition, return its path in zookeeper.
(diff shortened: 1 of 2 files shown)


yungchin:
@emmett9001 I was just looking at the callback interface that the new [JVM consumer][0] has for this, looking to steal some ideas, and I wonder if it might be worthwhile to do something similar? They define two callbacks, one for partitions that are to be removed, and one for partitions that are to be newly added, with the guarantee that the underlying consumer is not running while the callbacks run. 

The purpose of all that seems to be that that helps users who have some external offsets store: if you know a partition is about to be dropped, and you know the consumer is not servicing consume() calls currently, that makes it a safe moment to store that partition's offset in your external store. And for the second callback it's the same -  a safe moment to call reset_offsets(), basically.

[0]: http://people.apache.org/~nehanarkhede/kafka-0.9-consumer-javadoc/doc/org/apache/kafka/clients/consumer/ConsumerRebalanceCallback.html

emmett9001:
@yungchin I've implemented something comparable to the JVM consumer's interface, and I think it's ready for master. One possible difference that I don't actually think is a big deal is the fact that both of these callbacks are triggered after the rebalance is completed. Given that they're both still inside of the rebalance lock, we maintain the guarantee that consumption will not start until the callbacks return. Let me know what you think, and feel free to merge if you like.

emmett9001:
Actually I just realized - let me know if you think this is ready, but we'll wait to merge until right before cutting a minor version bump. That will make it easier to keep doing bugfix releases if other things on the 2.1.0 milestone get delayed.

pykafka/balancedconsumer.py
```diff
@@ -547,6 +561,10 @@ def _rebalance(self):
                         raise
                     log.info('Unable to acquire partition %s. Retrying', ex.partition)
                     time.sleep(i * (self._rebalance_backoff_ms / 1000))
+            if self._partitions_assigned_callback is not None:
+                self._partitions_assigned_callback(self, new_partitions)
+            if self._partitions_revoked_callback is not None:
+                self._partitions_revoked_callback(self, current_zk_parts - new_partitions)
```
yungchin: If the user wants to extract the `held_offsets` for the revoked partitions (to store them in their own custom offset store), I think this callback comes too late, because the internal consumer has at this point been replaced by the new one already.

emmett9001: It's called with the set of partitions that were owned and aren't anymore. The offset for each partition will be the same as it was before the rebalance. Using `held_offsets` would involve figuring out which partitions were revoked and then filtering those out of the output of `held_offsets` - a two-step process - right?

pykafka/balancedconsumer.py
```diff
@@ -547,6 +561,10 @@ def _rebalance(self):
                         raise
                     log.info('Unable to acquire partition %s. Retrying', ex.partition)
                     time.sleep(i * (self._rebalance_backoff_ms / 1000))
+            if self._partitions_assigned_callback is not None:
+                self._partitions_assigned_callback(self, new_partitions)
```
yungchin: If the user wants to use this callback to load offsets from their custom store via `reset_offsets()`, we would need some way to prevent `BalancedConsumer.consume()` from serving messages between L553 and here, otherwise it would in the mean time return messages at whatever offset we pulled from kafka.

emmett9001: Isn't that what https://github.com/Parsely/pykafka/issues/274 was about?

emmett9001:
@yungchin I've changed around the rebalance logic to support saving and fetching offsets from an external store, as well as resetting offsets via the partitions_assigned callback. Let me know what you think of the second try.

pykafka/balancedconsumer.py
```diff
@@ -157,6 +159,18 @@ def __init__(self,
             internal offset counter to `self._auto_offset_reset` and commit that
             offset immediately upon starting up
         :type reset_offset_on_start: bool
+        :param partitions_assigned_callback: Function accepting two arguments to be called
+            after a rebalance has completed. The arguments are a reference to the
+            consumer that just completed its rebalance and a sequence of Partition
+            instances that the consumer owns after rebalance. This function may also
```
yungchin: Just a docstring thingie here: the second argument is now an offsets dict.

pykafka/balancedconsumer.py
```diff
@@ -157,6 +159,18 @@ def __init__(self,
...
+            return a dictionary of `{partition_id: new_offset}` to have the consumer
+            reset its offsets to these new values.
+        :type partitions_assigned_callback: function
+        :param partitions_revoked_callback: Function accepting two arguments to be called
+            after a rebalance has completed. The arguments are a reference to the
+            consumer that just completed its rebalance and a sequence of Partition
+            instances that the consumer no longer owns after rebalance.
```
yungchin: Same here.

tests/pykafka/test_balancedconsumer.py
```diff
@@ -116,6 +116,49 @@ def setUpClass(cls):
...
+                b'test_consume_earliest', zookeeper_connect=self.kafka.zookeeper,
+                auto_offset_reset=OffsetType.EARLIEST,
+                partitions_assigned_callback=on_partitions_assigned,
+                partitions_revoked_callback=on_partitions_revoked
+            )
+            consumer_b = self.client.topics[self.topic_name].get_balanced_consumer(
+                b'test_consume_earliest', zookeeper_connect=self.kafka.zookeeper,
+                auto_offset_reset=OffsetType.EARLIEST
+            )
+            time.sleep(3)
```
yungchin: Just the other week I wrote a test helper for this, which should have been in its own pullreq - sorry - guess we can plug it in here after both get merged. See https://github.com/Parsely/pykafka/commit/6ef4199fff8b3778e96e6a0f75bd6772db33d4a9

pykafka/balancedconsumer.py
```diff
@@ -535,7 +566,24 @@ def _rebalance(self):
...
+                        cns = self._get_internal_consumer(list(new_partitions))
+                        if self._consumer is not None and \
+                                self._partitions_revoked_callback is not None:
+                            self._partitions_revoked_callback(
+                                self, _get_held_offsets_for_partitions(
+                                    current_zk_parts - new_partitions, self._consumer))
+                        if self._partitions_assigned_callback is not None:
+                            new_offsets = self._partitions_assigned_callback(
+                                self,
+                                _get_held_offsets_for_partitions(new_partitions, cns))
```
yungchin: Should this be `new_partitions - current_zk_parts`?

yungchin: Oh wait - the docstring says it is all partitions. I'm wondering which is more convenient though? If you really wanted all partitions you could get those off the `self` that also gets passed, but figuring out which ones are new is more tricky?

yungchin: Just took another look at what the JVM consumer docs say (not that we should necessarily do the same), and they pass all partitions here, but they also pass all the old partitions (not just the revoked ones) to the other callback. Maybe that's actually the most flexible way - if you wanted all the old offsets, there's no other way to get them.

yungchin: Apologies for the sort of stream-of-consciousness comment flood. I'm now thinking, the JVM consumer defines two callbacks because they call them in two different states, but now that we call these back-to-back.... so wouldn't we provide the same utility if we had.... one callback, with signature `rebal_callback(self, all_old_partitions_offsets, all_current_partitions_offsets ) => {returning optionally a new set of offsets}`?

That way users would get access to all the offsets, from before and after, and if they wanted to know which ones had been dropped / newly assigned, they could easily diff the two sets.

Sorry for all the hemming and hawing - I'm actually quite excited about how powerful this interface is becoming!!

emmett9001:
@yungchin updated with a single callback instead of two. I agree it's significantly simpler and actually more powerful this way.

pykafka/balancedconsumer.py
```diff
@@ -535,7 +564,18 @@ def _rebalance(self):
 
                     # Only re-create internal consumer if something changed.
                     if new_partitions != self._partitions:
-                        self._setup_internal_consumer(list(new_partitions))
+                        cns = self._get_internal_consumer(list(new_partitions))
+                        old_offsets = _get_held_offsets(current_zk_parts, self._consumer)
```
yungchin: One last thing I just realised: this can now just be `old_offsets = self._consumer.held_offsets` (and something similar is true for `new_offsets`. That's also safer in case we recover after a zk session timeout, in which case maybe `current_zk_parts == {}`.

Other than that, I think this is all ready, and pretty awesome too.

Issue #314, opened 2015-10-26
Exception in thread Thread and hang


JaneVinWin:
when i consumed for a while, balanced consumer will raise a exception on screen, see below:
```
No handlers could be found for logger "kazoo.client"
Exception in thread Thread-15:
Traceback (most recent call last):
  File "/usr/lib64/python2.7/threading.py", line 551, in __bootstrap_inner
    self.run()
  File "/usr/lib64/python2.7/threading.py", line 504, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/lib/python2.7/site-packages/kazoo/recipe/watchers.py", line 36, in wrapper
    return func(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/kazoo/recipe/watchers.py", line 333, in _get_children
    result = self._func(children)
  File "/usr/lib/python2.7/site-packages/pykafka/balancedconsumer.py", line 532, in _consumers_changed
    self._rebalance()
  File "/usr/lib/python2.7/site-packages/pykafka/balancedconsumer.py", line 431, in _rebalance
    partitions = self._decide_partitions(participants)
  File "/usr/lib/python2.7/site-packages/pykafka/balancedconsumer.py", line 320, in _decide_partitions
    idx = participants.index(self._consumer_id)
ValueError: 'localhost.localdomain:dc19b88a-7233-4f44-8295-fd196f72a2b3' is not in list
 ```

However, the loop still hang there but not exit. and this exception can not be caught in code. See below:
for message in balanced_consumer:
            
            try:
                print "offset: " + str(message.offset)
                            
            except ex.SocketDisconnectedError,e:
                a = datetime.datetime.now()
                print "SocketDisconnectedError inside at: " + str(a) + ";",e
                break
            except ValueError,e:
                a = datetime.datetime.now()
                print "ValueError inside at: " + str(a) + ";",e
                break
            except Exception, e:
                a = datetime.datetime.now()
                print Exception, " inside at: " + str(a) + ";", e
                break
            else:
                pass

yungchin:
Thanks for reporting this. I believe this issue occurs when your zookeeper connection has been reset, which was very recently fixed in #308. Could you please check if you still see this if you use current master?

JaneVinWin:
@yungchin , thanks for your reply, could you tell me which code i should use? i'm confusing about a lot comment in #308

yungchin:
[...]

PR #327, opened 2015-11-04
SocketDisconnectedError in commit_offsets

### LINKED ISSUE #325: Make SimpleConsumer commit_offsets handle cluster updates too
This is a relatively minor issue if running a pure `SimpleConsumer`, because the fetch workers will almost always resolve a `SocketDisconnectedError` before the auto-commit worker gets there. With an `RdKafkaSimpleConsumer` however, we don't run those fetch workers, so the auto-commit worker will have to fend for itself.

emmett9001:
This pull request stops the `SimpleConsumer`'s offset committer thread from dying when it encounters a `SocketDisconnectedError` as a result of a broker going down.

Resolves #325.

### DIFF
--- a/pykafka/cluster.py
+++ b/pykafka/cluster.py
@@ -304,6 +304,7 @@ class Cluster(object):
                 log.error("Socket disconnected during offset manager "
                           "discovery. This can happen when using PyKafka "
                           "with a Kafka version lower than 0.8.2.")
+                self.update()
             else:
                 coordinator = self.brokers.get(res.coordinator_id, None)
                 if coordinator is None:
--- a/pykafka/cluster.py
+++ b/pykafka/cluster.py
@@ -259,10 +259,13 @@ class Cluster(object):
                 try:
                     self._brokers[id_].connect()
                 except socket.error:
-                    log.info('Failed to re-establish connection with broker id %s: %s:%s', id_, meta.host, meta.port)
+                    log.info('Failed to re-establish connection with broker id %s: %s:%s',
+                             id_, meta.host, meta.port)
             else:
                 broker = self._brokers[id_]
                 if meta.host == broker.host and meta.port == broker.port:
+                    log.info('Broker %s:%s metadata unchanged. Continuing.',
+                             broker.host, broker.port)
                     continue  # no changes
                 # TODO: Can brokers update? Seems like a problem if so.
                 #       Figure out and implement update/disconnect/reconnect if
@@ -298,9 +301,9 @@ class Cluster(object):
                 if i == MAX_RETRIES - 1:
                     raise
             except SocketDisconnectedError:
-                raise KafkaException("Socket disconnected during offset manager "
-                                     "discovery. This can happen when using PyKafka "
-                                     "with a Kafka version lower than 0.8.2.")
+                log.error("Socket disconnected during offset manager "
+                          "discovery. This can happen when using PyKafka "
+                          "with a Kafka version lower than 0.8.2.")
             else:
                 coordinator = self.brokers.get(res.coordinator_id, None)
                 if coordinator is None:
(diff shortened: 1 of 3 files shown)


pykafka/simpleconsumer.py
```diff
@@ -393,8 +394,17 @@ def commit_offsets(self):
...
+            try:
+                response = self._offset_manager.commit_consumer_group_offsets(
+                    self._consumer_group, 1, b'pykafka', reqs)
+            except (SocketDisconnectedError, IOError):
+                log.error("Error committing offsets for topic %s "
+                          "(SocketDisconnectedError)",
+                          self._topic.name)
+                if i >= self._offsets_commit_max_retries - 1:
+                    raise
+                continue
```
yungchin: Cool, thanks for this. Should this call `self._update()` before `continue` here?

yungchin:
While combing through the tracebacks posted on #314 I realised there may be one more thing here: `self._offset_manager` could potentially change if the cluster reconfigures?

pykafka/simpleconsumer.py
```diff
@@ -206,6 +206,7 @@ def _update(self):
         with self._update_lock:
             self._cluster.update()
             self._setup_partitions_by_leader()
+            self._discover_offset_manager()
```
emmett9001: `_update()` also needs to handle updating partition metadata. As it is now, stopping and then restarting a broker causes an infinite loop of failures in `fetch()` because the original `Broker` instance is still held and not connected.

yungchin: Ugh, that's nasty. I thought that originally, `_cluster.update()` was supposed to re-connect the existing `Broker` instance, not replace it, so that it was ok to reference that from `Partition`. Did we break that at some point?

emmett9001: In some cases that will happen, but in the one I'm testing now, [this case](https://github.com/Parsely/pykafka/blob/4ea463f18f74b38b1da72507ca54f6b0316c5afc/pykafka/cluster.py#L247) is hit instead. I think it's correct, it just makes this problem a lot more gnarly.

yungchin:
Just to cross-ref this info here: see Jason's comment at https://github.com/Parsely/pykafka/issues/189#issuecomment-154479063 - it looks like `BalancedConsumer._rebalance` (which begins with a call to `commit_offsets`) would also benefit from this pullreq. I think the specific situation there is that it is the offset-managing broker that has become unavailable, and so we'd need to update our cluster metadata before we even decide if we otherwise need a  `_rebalance`.

The other important thing is that Jason encountered `socket.timeout`, not `SocketDisconnectedError`. I wonder if that could be because we test on Debian whereas he runs CentOS, and maybe OS sockets are configured with different default behaviour? Certainly the python docs on `socket` suggest that you shouldn't need to expect timeouts unless you configured them, but here we are and they show up nevertheless. (This may need its own ticket actually?)

pykafka/simpleconsumer.py
emmett9001: Removing [this line](https://github.com/Parsely/pykafka/blob/2b3a9f14880f20649b076da156c00a4aefc2e55a/pykafka/cluster.py#L242) does solve the issue, though I don't think that's the best solution. Ideally we'd figure out what's holding a reference to the old `Broker` instance, but I'm having a really hard time doing that.

emmett9001:
@yungchin feel free to review (and merge) this when you can. I resolved the one issue I know about, and I think the `timeout` issue is probably better handled in #189

pykafka/cluster.py
```diff
@@ -298,9 +301,9 @@ def get_offset_manager(self, consumer_group):
                 if i == MAX_RETRIES - 1:
                     raise
             except SocketDisconnectedError:
-                raise KafkaException("Socket disconnected during offset manager "
-                                     "discovery. This can happen when using PyKafka "
-                                     "with a Kafka version lower than 0.8.2.")
+                log.error("Socket disconnected during offset manager "
```
yungchin: If we pass over `SocketDisconnectedError` here, the next loop iteration will still be talking to `broker` which we selected in L287, so wouldn't that remain disconnected? I wonder if a `self.update()` would be in order, but not entirely sure if that's safe in the middle of this.

PR #329, opened 2015-11-05
[RFC] simpleconsumer: allow automatic gc

yungchin:
With these changes, a SimpleConsumer is automatically finalised when it
goes out of scope.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -18,6 +18,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 __all__ = ["BalancedConsumer"]
+from functools import partial
 import itertools
 import logging
 import socket
@@ -189,6 +190,10 @@ class BalancedConsumer():
         if auto_start is True:
             self.start()
 
+    def __del__(self):
+        log.debug("Finalising {}".format(self))
+        self.stop()
+
     def __repr__(self):
         return "<{module}.{name} at {id_} (consumer_group={group})>".format(
             module=self.__class__.__module__,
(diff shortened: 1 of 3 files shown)


yungchin:
@emmett9001 this isn't a proposal for merging quite yet - mostly to collect some opinions/discussion. So we only need three weakrefs in `SimpleConsumer` to allow it to clean itself up, and I suspect it's a similar amount of changes for `BalancedConsumer` and `Producer`. The thing is though, we may not like spraying weakrefs all over the code? There are other solutions where we avoid leaking `self` into the threads - they'd just need relatively more modifications. What do you think?

emmett9001:
@yungchin In general I'd prefer to avoid using `weakref` wherever possible. It makes code tougher to reason about, and I don't know of a great benefit other than shrinking this diff. I'm not completely opposed to the use of `weakref`, but if there are other workable options I'd like to explore those first. Just my opinion - what do you think?

yungchin:
@emmett9001 I agree, I'm not generally a great fan of using `weakref` too liberally - especially `weakref.proxy` which makes it hard to spot where `ReferenceErrors` might pop up. 

But... I just now began an attempt to do this without weakrefs, and within about 5mins got convinced that in this case, the amount of changes I'd introduce would be so large that I'd likely introduce umpteen new bugs. E.g., for all the occurrences of `self` in `fetch()` we would need to find some other way to inject those members without injecting `self`. (That would also be ugly to read.)

Also, I think that for the specific case of `SimpleConsumer`, the risks of accidentally not catching a `ReferenceError` are limited: the only state that we need to ship out before destroying an instance is `held_offsets`, which we commit during `stop()`. Whatever happens to the fetch workers after that point, we don't really care about I think.

I'm not sure the same argument (that it's ok to be cavalier) applies for `Producer` and `BalancedConsumer`. Still, how about if I sprinkle some weakrefs on those too, just because that's a quick approach to begin with, so we can see where the changes need to go, and then we can discuss where we'd like different, non-weakref patches?

emmett9001:
@yungchin sounds good to me.

bcwalrus:
@yungchin Nice. Thanks for working on this. There may be a way to unittest this based on side effects. For example, when a `Producer` goes out of scope, its `OwnedBrokers` are stopped.

rogaha:
Awesome, thanks for working on this @yungchin ! I'll test it with our application as well!

yungchin:
Thanks @rogaha @bcwalrus it'd be great if this turns out sufficient for you (it may not be, of course - I find that it works well in a plain cpython interpreter, but the embedded interpreter in the WSGI server you have may not be quite as patient at program exit).

You've a very good point that this begs for tests because it's so easy to break again, and focusing on side effects seems a great idea. It will have to go with the integration tests, because if we send `self` references into mocks which discard them, then gc might seem to work perfectly, just only in the test suite. I'll create a new ticket for that.

PR #331, opened 2015-11-05
raise exceptions from balancedconsumer threads

emmett9001:
This pull request causes the `BalancedConsumer`'s public API to raise any exceptions encountered during threaded execution. These exceptions can occur either on the zookeeper checker thread or during threaded execution of zookeeper watch callbacks. The logic used is identical to that of the `SimpleConsumer`, making the case that Simple and Balanced should share code somehow. I'm still considering whether a subclass, utility function, or mixin makes the most sense, but that doesn't have to block this PR. Thoughts welcome.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -21,7 +21,9 @@ __all__ = ["BalancedConsumer"]
 import itertools
 import logging
 import socket
+import sys
 import time
+import traceback
 from uuid import uuid4
 import weakref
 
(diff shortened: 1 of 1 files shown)


yungchin:
I think it would probably make things really tidy as a mixin, but I totally agree there's no need to do that right now - should we make a tag "refactoring" so that we can throw it on the issue tracker backlog for a quiet week?

This all looks good to me. I was going to just hit "merge" but will leave it to you just in case you wanted to refactor first after all.

Issue #328, opened 2015-11-05
Cannot remove _zk_state_listener when creating BalancedConsumer failed


xqliang:
When an exception raised from [here](https://github.com/Parsely/pykafka/blob/master/pykafka/balancedconsumer.py#L438), the `_zk_state_listener` is [added](https://github.com/Parsely/pykafka/blob/master/pykafka/balancedconsumer.py#L239), but the `_running` flag is not set yet. So calling `stop()` is [ignored](https://github.com/Parsely/pykafka/blob/master/pykafka/balancedconsumer.py#L251), the `_zk_state_listener` and `_zookeeper` connection [which is owned by us](https://github.com/Parsely/pykafka/blob/master/pykafka/balancedconsumer.py#L235) are still there.

We can set `_running` to `True` before `_add_self`, but we will encounter [NoNodeError](https://github.com/Parsely/pykafka/blob/master/pykafka/balancedconsumer.py#L268) when stop it.

Another problem, when init `BalancedConsumer` with `auto_start=True`(the default setting), and an exception raised from [here](https://github.com/Parsely/pykafka/blob/master/pykafka/balancedconsumer.py#L438), we cannot get the  instance reference, so there is no way to do cleanups.

yungchin:
Thanks very much for this. I think we could solve this by 1) calling `stop()` if exceptions occur during `start()` (and maybe also adding a `__del__()` that runs `stop()`) and 2) instead of preventing `stop()` when not running, making `stop()` idempotent. Does that sound right?

If I may ask a favour: you seem to be able to trigger that exception out of `_add_self()` a lot - any chance you could turn that into a test case for `test_balancedconsumer.py`? Thanks!

xqliang:
@yungchin OK, I'm going to try it sometime.

emmett9001:
I don't think calling `stop()` in response to exceptions in `start()` entirely makes sense - isn't that the very thing that's causing the `NoNodeError`? `stop()` needs to be changed to not assume that everything in `start()` ran cleanly - possibly wrapping all of the zookeeper operations in `stop()` in a `try` that simply warns and passes on exception is good enough.

emmett9001:
This issue can be reproduced with test code like this
```python
_zookeeper = KazooClient('127.0.0.1:2181', timeout=6)
    _zookeeper.start()
    consumer = topic.get_balanced_consumer("foo",
                                        auto_offset_reset=OffsetType.EARLIEST,
[...]

PR #332, opened 2015-11-05
stop() balancedconsumer in response to errors in start()

emmett9001:
This pull request safely stops the `BalancedConsumer` if errors are encountered during `start()`.

Connected to #328

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -260,19 +260,19 @@ class BalancedConsumer():
         if self._consumer is not None:
             self._consumer.stop()
         self._zookeeper.remove_listener(self._zk_state_listener)
+        if self._owns_zookeeper:
+            # NB this should always come last, so we do not hand over control
+            # of our partitions until consumption has really been halted
+            self._zookeeper.stop()
+        else:
+            self._remove_partitions(self._get_held_partitions())
         try:
-            if self._owns_zookeeper:
-                # NB this should always come last, so we do not hand over control
-                # of our partitions until consumption has really been halted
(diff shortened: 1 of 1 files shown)


emmett9001:
@yungchin I'm not sure this is completely correct, but it's at least a start. Let me know if there are other things you think need to change here.

pykafka/balancedconsumer.py
```diff
@@ -259,16 +260,19 @@ def stop(self):
...
+                # of our partitions until consumption has really been halted
+                self._zookeeper.stop()
+            else:
+                self._remove_partitions(self._get_held_partitions())
+                self._zookeeper.delete(self._path_self)
+                # additionally we'd want to remove watches here, but there are no
+                # facilities for that in ChildrenWatch - as a workaround we check
+                # self._running in the watcher callbacks (see further down)
+        except NoNodeError:
+            log.warning("Error encountered while closing ZooKeeper connection")
```
yungchin: Minor nitpick: this makes it sound as if something unexpected happened, when in fact we know why this happens. If we'd wrap the `try .. except` around the `delete(_path_self)` line only, I'd argue the error would be so specific that you could just `pass` without logging it?

yungchin: (Just to check, I created a `BalancedConsumer` with an external `KazooClient` instance, and then called `stop()` on it twice - indeed the other lines here don't raise exceptions, just that one.)

yungchin:
Apart from the one inline comment, I've nothing to complain. I think this is all correct, and much better than preventing `stop()` if not `_running` (which this did before this patch).

PR #339, opened 2015-11-10
Load all topic values on values method.

rduplain:
Without this, print_topics can end up with a list of None values, as seen in testing a production Kafka instance with 200+ topics.


rduplain:
cc: @jasonrhaas

PR #341, opened 2015-11-10
Catch IOError during connection response.

rduplain:


pykafka/utils/socket.py
```diff
@@ -36,8 +36,11 @@ def recvall_into(socket, bytea, size):
     offset = 0
     while offset < size:
         remaining = size - offset
-        chunk = socket.recv(remaining)
-        if not len(chunk):
+        try:
+            chunk = socket.recv(remaining)
+        except IOError:
+            chunk = None
+        if chunk is None or len(chunk) == 0:
```
yungchin: Not a new bug in this pullreq, but a bug that affects what this wants to fix: when we raise from here, we can't call `BrokerConnection.disconnect()` (which we should, and which we do do in the other exception handler in `BrokerConnection.response()` above).

Issue #240, opened 2015-08-28
Synchronous batch produce() no longer possible in 2.0?


ltucker:
This doesn't apply to a released version, just the current trunk that is slated for 2.0.  

As far as I can tell, there is currently not a reasonable way to express a synchronous batch produce(...) at the discretion of the application in the Producer interface.

Whenever sync=True, you get one-by-one synchronous publishing.  It waits with arbitrary sleeps up to 300ms for empty queues after every single message ( produce(...) only accepts one message at a time. )

If sync=False, well then ... it's not synchronous and you aren't guaranteed that a flush has taken place no matter how you adjust batching parameters / timing parameters.

There is no public method to produce a batch of messages and wait for them to all be flushed.
There is no public method to force a synchronous flush of all pending messages.

This might be planned already, not sure.  Seems like this might also be helped by https://github.com/Parsely/pykafka/issues/238

yungchin:
Yes that's right - there was a discussion about this on #177 (and I've no idea how to reconstruct it, with all of github's magic about commits superseding older comments etc etc - maybe we ought to feed all github updates into the mailing list so we can at least read them back there) and in the end this got dropped, basically because it was unclear how to tell, in case of errors, which messages in the batch had been successfully enqueued and which ones hadn't been.

I see what you mean about how it fits in with #238. With that in place, you could probably make a synchronous `produce_batch()` which would just be a bunch of async `produce()` calls, plus a bulk `wait` for all of the futures.

ltucker:
yes, exactly -- ah, yes, sorry I missed the other discussion -- feel free to close or merge into others as you see fit.

yungchin:
Well maybe it's good to have this here as an independent issue. I reckon there's room for discussion to reintroduce it as a method `Producer.produce_batch()` or a name like that.

The thing is, I've never had a real use for batch-produce calls myself, which makes it harder to come up with the right semantics, but I guess you'll have some use and ideas for this?

ltucker:
Looks like would need some funkier signature to incorporate the key based partitioning parameter in any case (tuples of (key, message) or something)  For our cases, it would probably be sufficient / more flexible to have those futures in hand, but I'll think about it and see if I can propose something generally nice.

In some cases, for example, we don't want to commit the offset of a block of consumed messages until we can guarantee the results of processing the messages have been published (to some other topic(s)).

Today we use synchronous publication for that guarantee (with batching necessary for performance).

yungchin:
[...]

PR #344, opened 2015-11-11
Producer error reporting

emmett9001:
This pull request provides a mechanism by which the Producer can report errors on a per-message basis. Currently, this is implemented as Futures returned from produce(), though this method has some problems with memory inefficiency and will need to be changed. @yungchin had the idea of instead using a deque of errored messages that client code can consume.

See also #269, #240

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ notifications:
 
 
 install:
-    - pip install codecov kazoo tox testinstances
+    - pip install codecov kazoo tox testinstances futures
 
 before_script:
     - "python -m pykafka.test.kafka_instance 3 --download-dir /home/travis/kafka-bin &"
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ notifications:
 
 
 install:
-    - pip install codecov kazoo tox testinstances futures
+    - pip install codecov kazoo tox testinstances
 
 before_script:
     - "python -m pykafka.test.kafka_instance 3 --download-dir /home/travis/kafka-bin &"
--- a/README.rst
+++ b/README.rst
@@ -58,18 +58,41 @@ producing messages.
 
 .. sourcecode:: python
 
-    >>> with topic.get_producer() as producer:
+    >>> with topic.get_sync_producer() as producer:
     ...     for i in range(4):
-    ...         future = producer.produce('test message ' + i ** 2)
+    ...         producer.produce('test message ' + i ** 2)
 
-You're free to ignore the future returned from `produce()`, or you can later
-evaluate it to assure yourself that the message made it to disk on the cluster.
-This works as with any `concurrent.futures.Future` (`docs`_), by checking
-`future.result()` (which should be `None`) or `future.exception()`.
+The example above would produce to kafka synchronously, that is, the call only
+returns after we have confirmation that the message made it to the cluster.
 
-(If you would prefer your exceptions straight from the `produce()` call, you
-can create the producer with `sync=True`, but this is of course a lot slower,
-with network round-trips for every message produced.)
+To achieve higher throughput however, we recommend using the ``Producer`` in
+asynchronous mode.  In that configuration, ``produce()`` calls will return a
+``concurrent.futures.Future`` (`docs`_), which you may evaluate later (or, if
+reliable delivery is not a concern, you're free to discard it unevaluated).
+Here's a rough usage example:
+
+.. sourcecode:: python
+
+    >>> with topic.get_producer() as producer:
+    ...     count = 0
+    ...     pending = []
+    ...     while True:
+    ...         count += 1
+    ...         future = producer.produce('test message',
+    ...                                   partition_key='{}'.format(count))
+    ...         pending.append(future)
+    ...         if count % 10**5 == 0:  # adjust this or bring lots of RAM ;)
+    ...             done, not_done = concurrent.futures.wait(pending,
+                                                             timeout=.001)
+    ...             for future in done:
+    ...                 message_key = future.kafka_msg.partition_key
+    ...                 if future.exception() is not None:
+    ...                     print 'Failed to deliver message {}: {}'.format(
+    ...                         message_key, repr(future.exception()))
+    ...                 else:
+    ...                     print 'Successfully delivered message {}'.format(
+    ...                         message_key)
+    ...             pending = list(not_done)
 
 .. _docs: https://pythonhosted.org/futures/#future-objects
 
--- a/README.rst
+++ b/README.rst
@@ -67,6 +67,10 @@ evaluate it to assure yourself that the message made it to disk on the cluster.
 This works as with any `concurrent.futures.Future` (`docs`_), by checking
 `future.result()` (which should be `None`) or `future.exception()`.
 
+(If you would prefer your exceptions straight from the `produce()` call, you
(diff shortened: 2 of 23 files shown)


yungchin:
### Status update

I just pushed a first shot at replacing the cancelled futures interface with a queue that delivery reports need to be fetched from, see https://github.com/Parsely/pykafka/commit/f85e7434526251948cb5a04e83866c0931617ce4 - this is still a work in progress, but I just want to solicit some opinions on the usability of this interface, not to be botching it a second time. There are some rough performance numbers at the bottom of this update.

@emmett9001 @kbourgoin @rduplain @amontalenti 

### Design considerations / questions

 * For now, this just adds one new public method `Producer.get_delivery_report(self, block, timeout)` where the arguments are just the same as those to `queue.Queue.get()`, and the return value is a `(Message, Exception/None)` tuple (where `None` means success). You'd call this once for each call to `produce()` - see example code with benchmarks below
 * I will need to add an init parameter to enable/disable reporting (with reporting disabled it should then run as fast as current master). We may also want other modes of reporting than the currently implemented one, eg "report only failures", or "use one global queue" (rather than the current use of a thread-local queue). I think I'd like to do that sort of switching by just dynamically replacing the `get_delivery_report()` method based on an option passed to `__init__()`, rather than creating mixin classes that require bigger overall changes (users would then see different producer types), but I am open to all suggestions and ideas.
 * The main reason I started out with a thread-local queue here is that it nicely supports a use model that the futures interface also supported well: if you have something like a multi-threaded web server, with all threads sharing a single producer, but each thread checking its own delivery failures (or each running with `sync=True`), then you don't want them to de-queue each others reports. On the other hand, I suppose there may be use models where you'd `produce()` from multiple threads but have a single thread somewhere handling all delivery failures. This would not be possible with the currently proposed code, so maybe we really do need that "use one global queue" option too? (Not that I know of anyone with that particular use model...)

### Quick benchmark

This ran on my laptop, with the kafka cluster also on there, so I've kept only one significant decimal in the timings. They ran in ipython with `timeit bench_prod(10**5, linger_ms=100)`. All these were rebased against current master to make them otherwise similar.

 * current master at 66adb22a4d982fa61e206dd555573346c8b5f09d ran in **3.8s** with this benchmark:
```python
def bench_prod(n, **kwargs):                                           
    with topic.get_producer(**kwargs) as prod:
        for _ in xrange(n): prod.produce(None, partition_key="yow")
```
 * the futures implementation, rebased at 470daf211bc9f32e66f5047c96ed4d7175cb5462 ran in **7.3s** with
```python
def bench_prod(n, **kwargs):                                           
    with top.get_producer(**kwargs) as prod:
        concurrent.futures.wait([
            prod.produce(None, partition_key="yow") for _ in xrange(n)])
```
 * the queue implementation at f85e7434526251948cb5a04e83866c0931617ce4 ran in **4.5s** with
```python
def bench_prod(n, **kwargs):                                           
    with top.get_producer(**kwargs) as prod:                                   
        for _ in xrange(n):
            prod.produce(None, partition_key="yow") 
        for _ in xrange(n):
            msg, exc = prod.get_delivery_report()
            assert exc is None
```
Of course master is at a considerable advantage, not doing any reporting, and I think the overhead of the reports-queue implementation is quite reasonable.

pykafka/producer.py
```diff
@@ -242,33 +239,54 @@ def produce(self, message, partition_key=None):
...
-            leader_id = self._topic.partitions[partition_id].leader.id
+            leader_id = self._topic.partitions[message.partition_id].leader.id
             if leader_id in self._owned_brokers:
-                self._owned_brokers[leader_id].enqueue([(kv, partition_id, attempts)])
+                self._owned_brokers[leader_id].enqueue(message)
                 success = True
             else:
                 success = False
 
+    def _put_delivery_report(self, msg, exc=None):
```
emmett9001: I think this function should be factored out, since it's only one line and doesn't make that line significantly more readable than it already is.

pykafka/producer.py
```diff
@@ -493,3 +505,9 @@ def _wait_for_slot_available(self):
             else:
                 raise ProducerQueueFullError("Queue full for broker %d",
                                              self.broker.id)
+
+
+class _DeliveryReportQueue(threading.local):
```
emmett9001: Could factor out this class with

```python
self._delivery_reports = threading.local()
self._delivery_reports.queue = Queue()
```

setup.py
```diff
@@ -28,6 +28,7 @@ def get_version():
                          version_file.read()).group('version')
 
 install_requires = [
+    'futures',
```
emmett9001: Needs to be removed

.travis.yml
```diff
@@ -23,7 +23,7 @@ notifications:
 
 
 install:
-    - pip install codecov kazoo tox testinstances
+    - pip install codecov kazoo tox testinstances futures
```
emmett9001: Needs to be removed

README.rst
```diff
@@ -58,10 +58,44 @@ producing messages.
 
 .. sourcecode:: python
 
-    >>> with topic.get_producer() as producer:
+    >>> with topic.get_sync_producer() as producer:
     ...     for i in range(4):
     ...         producer.produce('test message ' + i ** 2)
 
+The example above would produce to kafka synchronously, that is, the call only
```
emmett9001: Just marking stuff that's become outdated so we don't forget it. This too.

emmett9001:
@yungchin this looks great, I'm excited to get it into master. My main thought in response to your design considerations is that I'd prefer to avoid adding switchable modes to this system too eagerly. I think we should figure out which mode is the most likely to be generally useful and present it as the only option. If after that we find that users would like tweaks that support different use cases, we can keep these mode suggestions in mind. (This doesn't include enabling/disabling delivery reporting entirely - I agree that's necessary).

I think we should stick with reporting all deliveries, not just errors. The original impetus for this feature was a user who needed confirmation that each message had been produced, and I think it's fine to risk providing slightly more information than some use cases will require. Making this switchable seems like needless machinery for too small a benefit.

README.rst
yungchin: Thanks that's great - was just thinking I needed to make me a list!

pykafka/producer.py
yungchin: It might grow to a two-liner when we check if reporting is enabled or not, but yeah, it could still be quickly factored out. I had mostly been thinking of what it might all look like if we ended up with a sort of mixin :)

yungchin:
Thanks @emmett9001, cool, that sounds like the best plan really - if nobody comes asking for other modes, we'll have avoided a bunch of code that nobody ends up using. The mode that's here now, with thread-local queues, is probably the most flexible one and would be the one I'd be using.

yungchin:
@emmett9001 I've updated the docs, and delivery-reporting is now disabled with default settings. When it's disabled, the mini-benchmark posted previously runs .1s faster than current master - presumably that's a little `Message.__slots__` boost.

If we wanted to have a "global queue" option later, it would be a very small addition: we could just drop in basically the same class as `_DeliveryReportQueue`, but not inheriting from `threading.local`.

Also for later reference: I ran a very quick benchmark with a "report errors only" adaptation, and that runs as fast as the no-reporting option (as you'd expect, if no delivery errors occur). So basically that's 10e5 messages in 3.7s for errors-only, and 4.1/4.5s for report-all (the 4.5s figure includes the calls to `get_delivery_report()`).

Issue #346, opened 2015-11-11
Allow grace period when BalancedConsumer detects failed ZK connection


yungchin:
Currently, we immediately bail out when our zookeeper connection is pending or dead. On a laggy network that's annoying, because it means the user sees the BalancedConsumer die all the time. (Reported by @amontalenti.)

amontalenti:
The way this looks:

```
Traceback (most recent call last):
  File "pykafka_consumer.py", line 91, in <module>
    tup = spout.next_tuple()
  File "pykafka_consumer.py", line 77, in next_tuple
    return self.emit_next()
  File "pykafka_consumer.py", line 59, in emit_next
    msg = self.consumer.consume()
  File "/home/am/.pyenv/versions/ist/lib/python2.7/site-packages/pykafka/balancedconsumer.py", line 715, in consume
    raise ZookeeperConnectionLost
pykafka.exceptions.ZookeeperConnectionLost
```

and, how the DEBUG logs look:

```
ARNING:kazoo.client:Connection dropped: outstanding heartbeat ping not received
WARNING:kazoo.client:Transition to CONNECTING
INFO:kazoo.client:Zookeeper connection lost
INFO:pykafka.balancedconsumer:zk_state_listener: SUSPENDED
INFO:kazoo.client:Connecting to z01.istresearch.com:2181
DEBUG:pykafka.balancedconsumer:Suspending internal consumer (tx220:4c915509-2717-48f8-be64-006867b83f47)
```

amontalenti:
BTW, I was able to reproduce this using `ufw` port blocking from my client machine.

So, I:

1. spun up a pykafka consumer (single balanced consumer on 1 node).
2. ran `ufw reject out 2181` on my machine, and then reset my network interface to make the firewall rule take effect, because it didn't effect pykafka's established connection
3. almost instantly, the pykafka consumer lost connection with Zooekeeper and raised the `ZookeeperConnectionLost` exception

This made me realize that this exception simply shouldn't be raised at all. It's basically monitoring the ZK connection for the slightest outage (even ephemeral network outage or packet loss), and raising an exception that breaks the consumer thread instantly. I looked into the commits that added it with @rduplain and I don't see a good reason this custom exception type should be raised. So, I'm going to kill that exception from the code in a PR altogether.

PR #348, opened 2015-11-11
Stop ZK network disconnects from killing consumer

amontalenti:
Addresses the issue raised in #346.

A Zookeeper connection loss shouldn't kill the main consumer thread, instead we should let Kazoo's session timeout happen if ZK is truly down for awhile.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -711,8 +711,6 @@ class BalancedConsumer():
             except ConsumerStoppedException:
                 if not self._running:
                     return
-                elif not self._zookeeper.connected:
-                    raise ZookeeperConnectionLost
                 continue
             if message:
                 self._last_message_time = time.time()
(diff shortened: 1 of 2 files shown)


yungchin:
Yes, fair enough, raising an exception this quickly was too heavy-handed.

Still I think the user might want to be able to distinguish between having lost the kazoo session and just having no new messages available. How about if instead of checking for `self._zookeeper.connected` (which lumps `KazooState.PENDING` and `KazooState.LOST` together), we would check `self._zookeeper.state == KazooState.LOST` (ie only a real session timeout), would that be alright?

yungchin:
@amontalenti as just noted on #273 I changed my mind here :)

Note however that the removal of just the exception doesn't change the fact that the internal consumer gets killed by our `_zk_state_listener` - with this pullreq a `consume()` loop would just do a busy loop, never getting any new messages until the zk connection is back. So if that's ok, I'll close this and open a new pullreq with a complete revert of #317.

PR #351, opened 2015-11-13
Revert #317

### LINKED ISSUE #273: BalancedConsumer.consume() should block/fail during rebalance or if zk connection is pending/dead

### LINKED ISSUE #346: Allow grace period when BalancedConsumer detects failed ZK connection

yungchin:
Resolves #346, closes #273, and replaces #348.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -30,12 +30,11 @@ import weakref
 
 from kazoo.client import KazooClient
 from kazoo.exceptions import NoNodeException, NodeExistsError
-from kazoo.protocol.states import KazooState
 from kazoo.recipe.watchers import ChildrenWatch
 
 from .common import OffsetType
 from .exceptions import (KafkaException, PartitionOwnedError,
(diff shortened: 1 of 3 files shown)


yungchin:
@amontalenti @emmett9001 this is a full revert of the changes in #317, which caused the consumer to be suspended if we couldn't speak to zookeeper. So there's nothing new and a lot of old here, but just to be sure I've slapped a "needs review" on it anyway.

PR #359, opened 2015-11-17
dead weakrefs

### LINKED ISSUE #345: BalancedConsumer watcher callback fails over weakref
As reported by @amontalenti and most likely introduced by #329:
```
ERROR:kazoo.recipe.watchers:unbound method wrapped() must be called with BalancedConsumer instance as first argument (got weakproxy instance instead)
  File "/home/am/.pyenv/versions/ist/lib/python2.7/site-packages/kazoo/recipe/watchers.py", line 333, in _get_children
    result = self._func(children)
TypeError: unbound method wrapped() must be called with BalancedConsumer instance as first argument (got weakproxy instance instead)
DEBUG:kazoo.client:Received response(xid=4): [u'1-1', u'1-0', u'3-2']
```

emmett9001:
This pull request improves the `ChildrenWatch` callbacks registered by `BalancedConsumer`, making them safe to use even after the `BalancedConsumer` that created them is no longer referenced by the `weakref.proxy` used in their creation. It uses `__repr__` as a cheap way to discover whether the object is still referenced, and simply passes if not. This is a way around the fact that `KazooClient` lacks a way to explicitly remove `ChildrenWatch` objects from zookeeper.

Fixes #345.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -18,7 +18,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 """
 __all__ = ["BalancedConsumer"]
-import functools
 import itertools
 import logging
 import socket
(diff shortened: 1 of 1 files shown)


pykafka/balancedconsumer.py
```diff
@@ -416,6 +415,17 @@ def _get_participants(self):
...
         return participants
 
+    def _build_watch_callback(self, fn, proxy):
+        """Return a function that's safe to use as a ChildrenWatch callback"""
+        def _callback(children):
+            # discover whether the referenced object still exists
+            try:
+                proxy.__repr__()
+            except ReferenceError:
+                return
```
yungchin: A `return False` here will prevent `ChildrenWatch` from calling again after this.

It might also be good to reference the github issue this relates to in the docstring, because it may not be obvious why we did this later. It's a neat solution though!

PR #358, opened 2015-11-16
Topic instantiation improvements

emmett9001:
This pull request makes `Topic.__init__` resilient to ephemeral `LeaderNotAvailable` errors caused by differences in Kafka's reporting on available brokers. It retries the creation of a new `Topic` three times, each time asking a randomly selected broker from the list of currently connected brokers.

Closes #338.

PR #360, opened 2015-11-17
update cluster in response to NotCoordinatorForConsumer

### LINKED ISSUE #353: Consider a full cluster update on NotCoordinatorForConsumer
This was reported by @ottomata in a comment at https://github.com/Parsely/pykafka/issues/189#issuecomment-155821386 but deserves its own ticket: if we get a `NotCoordinatorForConsumer` response from our current offset coordinator, I think it makes sense to assume that the cluster has changed and we should update our metadata. The update should prevent hitting the `SocketDisconnectedError` shown in the linked gist https://gist.github.com/ottomata/c3a6c47685cdd75bfc3f

emmett9001:
This pull request does a full cluster update on `NotCoordinatorForConsumer` in `SimpleConsumer`.

Fixes #353.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -246,7 +246,8 @@ class SimpleConsumer(object):
             )
 
         def _handle_NotCoordinatorForConsumer(parts):
-            self._discover_offset_manager()
+            log.info("Updating cluster in response to NotCoordinatorForConsumer")
+            self._update()
 
         def _handle_NotLeaderForPartition(parts):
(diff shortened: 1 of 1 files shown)


yungchin:
There's a chance that this might additionally need the `random.shuffle` when selecting a broker to fetch metadata (which is in #358) in order for the `update()` to solve anything, and here, as in that other pullreq, I cannot reproduce the situation to test it. But it seems to me that this is an improvement anyway, so I'll just merge.

PR #361, opened 2015-11-18
Set disconnected on socket errors from util.

rduplain:
As @yungchin pointed out in #341, we missed a case. Here it is.


emmett9001:
Looks fine to me. :+1:

yungchin:
Looks fine to me too.

While we're here... should we do the same disconnect thing around all uses of `self._socket`? (There's eg. a `sendall()` at the end of `request()` a few lines up.)

PR #363, opened 2015-11-18
held_offsets idempotency

### LINKED ISSUE #216: Proposal for changes in manual offset management

emmett9001:
This pull request makes `held_offsets()` return `-2` (`OffsetType.EARLIEST`) when the consumer holds `-1`. This fixes #216 by allowing the output from `held_offsets()` to be used directly as input to `reset_offsets` without producing unexpected results.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -283,7 +283,9 @@ class SimpleConsumer(object):
     @property
     def held_offsets(self):
         """Return a map from partition id to held offset for each partition"""
-        return {p.partition.id: p.last_offset_consumed
+        return {p.partition.id:
+                (OffsetType.EARLIEST if p.last_offset_consumed == -1
+                 else p.last_offset_consumed)
                 for p in itervalues(self._partitions_by_id)}
 
     def __del__(self):
--- a/tests/pykafka/test_simpleconsumer.py
+++ b/tests/pykafka/test_simpleconsumer.py
@@ -50,6 +50,22 @@ class TestSimpleConsumer(unittest2.TestCase):
             self.assertEquals(len(messages), self.total_msgs)
             self.assertTrue(None not in messages)
 
+    @staticmethod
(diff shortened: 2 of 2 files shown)


yungchin:
Thanks, this looks good. I suppose it's debatable whether it requires a version number bump (strictly speaking, yes, because there could be a theoretical user who has code that says `if offset == -1: do_something_totally_magical()`) but the practical reality is probably that it's fine to slip it in as a minor fix.

The tests were failing, which I thought I'd quickly fix, but of course the whole "convert -2s to something that makes sense" took me twenty iterations in the end :)

Issue #316, opened 2015-10-27
connect to kafka cluster using zookeeper (without knowing the actual kafka cluster)


timfeirg:
I use zookeeper to manage kafka cluster, but using `pykafka` one should always initiate a `pykafka.cluster.Cluster` instance using the kafka cluster hosts, **is there anyway to connect to kafka cluster via zookeeper cluster, so that the consumer doesn't care about the actual kafka hosts**?

typical java code to do the job:
```java
public static void main(String[] args) {
        Properties props = new Properties();
        props.put("zookeeper.connect", "xxxx:2181,xxxx:2181,xxxx:2181/kafka");
        props.put("zookeeper.session.timeout.ms", "100000");
        props.put("group.id", "test_kakfa_1234");
        props.put("serializer.class", "kafka.serializer.StringEncoder");
​
        ConsumerConfig config = new ConsumerConfig(props);
​
        ConsumerConnector consumer = kafka.consumer.Consumer.createJavaConsumerConnector(config);
​
        Map<String, Integer> map = new HashMap<>();
        map.put("app_action", 1);
​
        StringDecoder keyDecoder = new StringDecoder(new VerifiableProperties());
        StringDecoder valueDecoder = new StringDecoder(new VerifiableProperties());
​
        Map<String, List<KafkaStream<String, String>>> consumerMap =
                consumer.createMessageStreams(map, keyDecoder, valueDecoder);
        KafkaStream<String, String> stream = consumerMap.get("app_action").get(0);
        ConsumerIterator<String, String> it = stream.iterator();
        while (it.hasNext())
            System.out.println(it.next().message());
    }
```

emmett9001:
Hi @timfeirg, thanks for filing this. Pykafka doesn't support this at this time, unfortunately. It does seem like a worthwhile feature, since I don't think kafka is ditching zookeeper anytime soon.

emmett9001:
I've taken a shot at this in #364. Let me know if it satisfies your use case @timfeirg.

PR #364, opened 2015-11-19
zk connect strings in Cluster()

emmett9001:
This pull request implements #316 by falling back to `KazooClient` in `Cluster._get_metadata`. `KafkaClient` still accepts a list of brokers, but in cases where it can't connect to any of those brokers, it attempts to connect a `KazooClient` to the list of hosts and to get the list of brokers from that ZooKeeper instance.


yungchin:
@emmett9001 this is really neat and convenient. I was wondering if you see a good way to make the `zookeeper_connect` default arg in `BalancedConsumer` also get auto-filled based on this?

Issue #365, opened 2015-11-19
pykafka not compatible with gevent.monkey.patch_socket


fuellee:
after `gevent.monkey.patch_socket()` :

    Traceback (most recent call last):
    File "./lbs_info_all.py", line 66, in <module>
        client = KafkaClient(hosts=hosts)
    File "/home/lify/ENV/lib/python2.7/site-packages/pykafka/client.py", line 83, in __init__
        source_address=self._source_address
    File "/home/lify/ENV/lib/python2.7/site-packages/pykafka/cluster.py", line 173, in __init__
        self.update()
    File "/home/lify/ENV/lib/python2.7/site-packages/pykafka/cluster.py", line 315, in update
        metadata = self._get_metadata()
    File "/home/lify/ENV/lib/python2.7/site-packages/pykafka/cluster.py", line 226, in _get_metadata
        'Unable to connect to a broker to fetch metadata. See logs.')
    RuntimeError: Unable to connect to a broker to fetch metadata. See logs.

emmett9001:
Thanks @fuellee. Are you trying to run pykafka in a gevent-managed environment? If so I'm afraid we can't guarantee that it will work properly at the moment. We have a ticket open about supporting this in the future, so maybe this is a push to prioritize that feature. As far as my investigation shows, though, solving this seems to require patching pykafka's socket machinery to directly use `gevent.socket`. If you have any insight into this, please share.

emmett9001:
The pull request for gevent support is in the final testing phase and should hopefully make it into a release by the end of this week.

Issue #301, opened 2015-10-13
use_greenlets not Implemented?


blooshirt:
Setting `use_greenlets` to `True` in the `KafkaClient` constructor results in an error. It sets the `KafkaClient._handler` to `None`, which is eventually called for a `Queue()` [here](https://github.com/Parsely/pykafka/blob/master/pykafka/handlers.py#L102), resulting in an error. The docs state regarding `use_greenlets`, `"If True, use gevent instead of threading"`. What is the intention for this feature? Is it something that is yet to be implemented, or something that has been deprecated? Either way, it might be useful to update the docs to reflect the current state.

yungchin:
I believe this was carried over from the earlier samsa code base, and @kbourgoin has plans to reimplement it - he'll have a more complete answer for you.

emmett9001:
`use_greenlets` is currently one of the lower-priority features on our future roadmap. Its presence is a holdover from a previous idea that hasn't yet been implemented. I'll update the docs.

emmett9001:
I've [removed](https://github.com/Parsely/pykafka/commit/64df140a062ec0bb921eabf56180090d9b4a0b67) this option for now. I'll keep this issue open as a reminder to implement this at some point.

emmett9001:
#365 is related to this

emmett9001:
#368 is likely to be put into a release later this week, after which pykafka will support gevent for pure-python producers and consumers on cpython 2 and 3.

PR #368, opened 2015-11-24
use_greenlets

emmett9001:
This pull request adds switchable support for greenlet-based concurrency to KafkaClient.

Related to #301, #365.

### DIFF
--- a/pykafka/handlers.py
+++ b/pykafka/handlers.py
@@ -83,9 +83,8 @@ class ThreadingHandler(Handler):
     # turn off RLock's super annoying default logging
     RLock = functools.partial(threading.RLock, verbose=False)
 
-    def _sleep(instance, seconds=0):
+    def sleep(self, seconds=0):
         time.sleep(seconds)
-    sleep = _sleep
 
     def spawn(self, target, *args, **kwargs):
         t = threading.Thread(target=target, *args, **kwargs)
@@ -101,9 +100,8 @@ class GEventHandler(Handler):
     Lock = GERLock  # fixme
     RLock = GERLock
 
-    def _sleep(instance, seconds=0):
+    def sleep(self, seconds=0):
         gevent.sleep(seconds)
-    sleep = _sleep
 
     def spawn(self, target, *args, **kwargs):
         t = gevent.spawn(target, *args, **kwargs)
--- a/pykafka/handlers.py
+++ b/pykafka/handlers.py
@@ -21,9 +21,9 @@ __all__ = ["ResponseFuture", "Handler", "ThreadingHandler", "RequestHandler"]
 from collections import namedtuple
 import functools
 import gevent
-from gevent.queue import JoinableQueue
-from gevent.event import Event as GEEvent
-from gevent.lock import RLock as GERLock
+import gevent.event
+import gevent.lock
+import gevent.queue
 import logging
 import threading
 import time
@@ -95,10 +95,10 @@ class ThreadingHandler(Handler):
 
 class GEventHandler(Handler):
     """A handler that uses a greenlet to perform its work"""
-    Queue = JoinableQueue
-    Event = GEEvent
-    Lock = GERLock  # fixme
-    RLock = GERLock
+    Queue = gevent.queue.JoinableQueue
+    Event = gevent.event.Event
+    Lock = gevent.lock.RLock  # fixme
+    RLock = gevent.lock.RLock
 
     def sleep(self, seconds=0):
         gevent.sleep(seconds)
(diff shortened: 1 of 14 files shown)


pykafka/handlers.py
```diff
@@ -71,14 +76,17 @@ def spawn(self, target, *args, **kwargs):
...
+    """A handler that uses a :class:`threading.Thread` to perform its work"""
     Queue = Queue
     Event = threading.Event
     Lock = threading.Lock
     # turn off RLock's super annoying default logging
     RLock = functools.partial(threading.RLock, verbose=False)
 
+    def _sleep(instance, seconds=0):
+        time.sleep(seconds)
+    sleep = _sleep
```
kbourgoin: When private then not?

pykafka/handlers.py
```diff
@@ -86,6 +94,22 @@ def spawn(self, target, *args, **kwargs):
         return t
 
 
+class GEventHandler(Handler):
+    """A handler that uses a greenlet to perform its work"""
+    Queue = JoinableQueue
+    Event = GEEvent
+    Lock = GERLock  # fixme
```
kbourgoin: wha'ts to be fixed?

pykafka/handlers.py
```diff
@@ -86,6 +94,22 @@ def spawn(self, target, *args, **kwargs):
...
+    RLock = GERLock
```
kbourgoin: This is mostly a style thing, but if you're only using it one place, the extra characters to have this be `gevent.event.Event` will make it a lot clearer to read. Otherwise, you're using a work `GEEvent` which nobody will have seen before. Not a big thing, but it'll help readability, IMO.

kbourgoin:
This is actually really cool to see. The `Handler` idea was based on samsa with the intent of someday supporting gevent. It's nice to see work which @mwhooker started and I carried forward into PyKafka finally come to fruition. :)

pykafka/handlers.py
emmett9001: This is an `RLock` and it should be a regular `Lock`, but gevent doesn't seem to have one?

kbourgoin: Oh, huh. Does it function the same as the RLock? If so, I guess it doesn't matter much? Worth testing to be sure it's compatible for how we're using the RLock.

pykafka/producer.py
```diff
@@ -392,6 +393,7 @@ def _wait_all(self):
         log.info("Blocking until all messages are sent")
         while any(q.message_is_pending() for q in itervalues(self._owned_brokers)):
             time.sleep(.3)
+            self._cluster.handler.sleep(.3)
```
philip-sterne: Sleep called twice?

emmett9001: Yup, that's an error. This branch is still in testing, and it doesn't work fully yet.

philip-sterne: Well it's certainly a great feature that would make our lives much easier.  

Our team is going to be watching this branch with interest :smile:

ryananguiano:
Are there any tasks that you think you could use any help with? I need this feature ASAP in my production system and am willing to help with this.

emmett9001:
@ryananguiano Thanks for reaching out. Currently this branch seems to work well, but it introduces some concerning test failures. I've spent a while trying to figure out what's causing them, but I'm afraid my lack of experience with gevent is making it difficult. It could also be something not directly related to gevent that's causing the tests to fail. If you'd like to look into the test failures in particular, that would be helpful. It's the most important blocking issue at this point. Aside from that, if you have experience with gevent and can suggest any improvements to the changeset, that would be helpful too.

ryananguiano:
You're right. It seems to be working fine outside of testing. I will definitely take a look and see if I can find anything.

rduplain:
This fixes #365.

ryananguiano:
This failures seem to be confined to RdKafka. Since this is a C extension, I don't think that Gevent can safely use it in a concurrent manner. Maybe gevent should not be allowed to be used with rdkafka. This would allow this much needed feature to get out in the wild.

emmett9001:
@ryananguiano That's a reasonable idea. There is one test failure that's not related to rdkafka, but I'm re-running that test to see if it's ephemeral. I can look into merging this without rdkafka support.

emmett9001:
Tests on this branch are now passing. It totally ignores PyPy and rdkafka compatibility, but it works fine on python 2 and python 3. I'll let this sit for a few days in case anyone else would like to test it, but I'm planning on doing a 2.2.0 release by the end of the week that includes this changeset.

PR #373, opened 2015-11-25
Add consume_topic to kafka_tools

dan-blanchard:
This just adds a simple producer that will dump a few messages to a file or stdout.  Mostly useful for testing.

Along the way I noticed some Python 3 issues with kafka_tools, so I fixed those.

### DIFF
--- a/pykafka/cli/kafka_tools.py
+++ b/pykafka/cli/kafka_tools.py
@@ -7,7 +7,6 @@ import sys
 import time
 
 import tabulate
-from kazoo.client import KazooClient
 
 import pykafka
 from pykafka.common import OffsetType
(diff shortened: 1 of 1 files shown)


dan-blanchard:
It works now!

pykafka/cli/kafka_tools.py
```diff
@@ -284,8 +367,11 @@ def _get_arg_parser():
 def main():
     parser = _get_arg_parser()
     args = parser.parse_args()
-    client = pykafka.KafkaClient(hosts=args.host)
-    args.func(client, args)
+    if args.command:
```
dan-blanchard: This makes it so if you forget a command, it prints the full help.

PR #374, opened 2015-11-27
[WIP] kafka 0.9 compatibility fixes

yungchin:
This branch will just put in the minimal changes to allow pykafka to run against a kafka 0.9 cluster, without however supporting any of the new features yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,21 @@ addons:
 matrix:
   include:
     - python: 2.7
-      env: TOXENV=py27 BROKERS=localhost:9092,localhost:9093,localhost:9094 ZOOKEEPER=localhost:2181 KAFKA_BIN=/home/travis/kafka-bin
+      env: TOXENV=py27 BROKERS=localhost:9092,localhost:9093,localhost:9094 ZOOKEEPER=localhost:2181 KAFKA_BIN=/home/travis/kafka-bin KAFKA_VERSION=0.8.2.1
     - python: 3.4
-      env: TOXENV=py34 BROKERS=localhost:9092,localhost:9093,localhost:9094 ZOOKEEPER=localhost:2181 KAFKA_BIN=/home/travis/kafka-bin
+      env: TOXENV=py34 BROKERS=localhost:9092,localhost:9093,localhost:9094 ZOOKEEPER=localhost:2181 KAFKA_BIN=/home/travis/kafka-bin KAFKA_VERSION=0.8.2.1
     - python: 3.5
-      env: TOXENV=py35 BROKERS=localhost:9092,localhost:9093,localhost:9094 ZOOKEEPER=localhost:2181 KAFKA_BIN=/home/travis/kafka-bin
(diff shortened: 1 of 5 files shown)


yungchin:
So it looks like most of the tests pass with just a one-char addition in `SimpleConsumer`. However, our cluster-management code breaks in the test tear-downs: apparently topic deletion works through a different script on kafka 0.9....

yungchin:
Remaining todo's here:

  - [ ] change [this test cluster setup code][0] so that we can hold multiple kafka versions (currently the test setup will silently run whatever version is found, even if you specify another version explicitly)
  - [ ] add kafka 0.9 to the travis-ci matrix

[0]: https://github.com/Parsely/pykafka/blob/00d4ad424303b3ed0a6f1c2f3eadde8a01ac602b/pykafka/test/kafka_instance.py#L167-L170

yungchin:
Ugh - I just tried this, saw `test_consume()` hang locally, updated the test to get it to fail without hanging forever, and couldn't get it to hang again - of course not...

So now I pushed it to see it hang on travis-ci, but it looks like the rewritten version is way too impatient for that environment, failing all over the place. I'll try again later.

yungchin:
Ok, I've now caught that test hanging locally, but it rarely happens so it's hard to investigate. The issue is that there are fewer messages to consume than we produce during the test setup. The consumer doesn't seem to raise any exceptions. I wonder if something changed about the (kafka-provided) producer script that the test harness calls, that means some of our messages go missing, without errors about it.

yungchin:
@emmett9001 would it make sense if I cherry-pick c756b2f to master, so people can start experimenting with pykafka + kafka 0.9, while we leave this pullreq to sort out the test suite?

emmett9001:
@yungchin seems fine to me. If someone's cloning the repo from github, they can also just grab this branch for their testing. I suppose the benefit of putting it in master is that it indicates some degree of stability (though I don't think it necessarily should).

yungchin:
(Blegh - sorry for all the commit noise here, this thing is just a pain because it only reliably shows up on travis-ci. I guess I've turned this pullreq into an angry testing ground, not something particularly mergeable now.)

@emmett9001 I'm convinced the issue is in the test harness only: when we call the 0.9 `kafka-console-producer.sh` with the `--old-producer` flag (which makes the script work the way it did under 0.8) as in a5fa927, it passes `test_consume` just fine on 0.9 (but breaks on kafka 0.8 because that flag didn't exist there).

All this would go away if we just use our own `Producer` during the test setup. I know that's not ideal, because it means breaking `Producer` would break consumer tests too, but heh, we do the opposite in the producer tests already :)

emmett9001:
@yungchin I've updated the test to use our producer. I agree it's not perfect, but I'd argue it's actually better than using the test harness for the time being, since we don't have any tests for the harness itself. With these passing tests, it looks like this branch is fine to merge (if you care to rebase away some of your inbetween commits, feel free).

yungchin:
Thanks, yes I've rebased away all my failed patches to the console-producer, and added one more cleanup commit. Will merge.

PR #377, opened 2015-12-04
enforce bytes consumer group argument

### LINKED ISSUE #376: Python3 topic and group names must be bytes objects
In Python3, string startswith does not support bytes object arguments and bytes startswith does not support string object arguments.

Because of this, topic name strings given as a key to client.topics result in a traceback:

```
File "... /python3.4/site-packages/pykafka/cluster.py", line 141, in _should_exclude_topic
    return topic_name.startswith(b"__")
TypeError: startswith first arg must be str or a tuple of str, not bytes
```

After passing the topic name as a bytes object instead, that issue stopped. Then other strangeness happened when attempting to get a balanced consumer from a topic with a specified consumer_group. When the consumer group was specified as a string object keyword argument to get_balanced_consumer, a NoPartitionsForConsumerException occurred. This was fixed by passing the consumer group as a bytes object as well.

emmett9001:
Fixes #376 by improving the error message given when unicode is supplied to a consumer constructor.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -182,6 +182,8 @@ class BalancedConsumer(object):
         :type use_rdkafka: bool
         """
         self._cluster = cluster
+        if not isinstance(consumer_group, bytes):
+            raise TypeError("consumer_group must be a bytes object")
         self._consumer_group = consumer_group
         self._topic = topic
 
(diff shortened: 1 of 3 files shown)


yungchin:
The type checking here is a bit too strict: it doesn't allow `NoneType`. Other than that it's a definite improvement.

I'm actually puzzled though why this bug shows up as a `NoPartitionsForConsumerException` (as reported in #376), I mean, why doesn't this make `start()` fail altogether?

PR #381, opened 2015-12-06
rdkafka: fix reference leak in Consumer_consume()

### LINKED ISSUE #379: rdkafka memory issue
First wonderful project, i've watching from the sidelines and really excited about pushing for first class python tools in the kafka ecosystem. 

I have started testing the new rdkafka support in 2.1.0 and i do see a throughput improvement (yeah!), but there appears to be a memory leak. My attempts to implement memory profiling have come up with very little because the tooling around C extensions is limited and my knowledge is even less. 

I'm going to dig deeper next week but wanted to put a note here to see if anyone else run into this. 

I am deploying these workers as docker containers, so i am not ruling out some issue with my container setup, but below is a chart of memory.rss for the same container. The flat lines are configured `use_rdkafka=False`, the ominous slanted line only difference is `use_rdkafka=True`. These are configured from envvars so the underlying images are the same.

screen shot 2015-12-05 at 12 02 20 pm

Sorry for the vagueness of this, i work on getting some better threads to pull on next week.

My setup:
pykafka 2.1.0
rdkafka 0.8.6

yungchin:
This was introduced with 5d7abfa2 - there's a Py_XDECREF in place, but
the erroneously declared block-scoped variable is not the one that gets
DECREF'd.  Resolves #379.

### DIFF
--- a/pykafka/rdkafka/_rd_kafkamodule.c
+++ b/pykafka/rdkafka/_rd_kafkamodule.c
@@ -965,7 +965,7 @@ Consumer_consume(RdkHandle *self, PyObject *args)
 #else
         const char *format = "{s:s#,s:s#,s:l,s:L}";
 #endif
-        PyObject *kwargs = Py_BuildValue(
+        kwargs = Py_BuildValue(
             format,
             "value", rkmessage->payload, rkmessage->len,
             "partition_key", rkmessage->key, rkmessage->key_len,


yungchin:
@emmett9001 this is a pretty nasty one - it leaks the whole init-dict for the consumed `Message`. Could it go in a point release fairly soon? I ought to be adding a test case to catch this sort of thing, but the truth is I don't know a way to write that test.

emmett9001:
@yungchin Nicely done, this really reminds me how long it's been since I've written C code. I'll look toward cutting a bugfix release this week.

yungchin:
Great, thanks! Just to add a confessional: I think when we decided to follow PEP 7, I proposed to deviate from it on one point: strict C89-compliance, which I felt was ugly because it means you must declare all vars at the top of a block (which makes things less readable, I felt, because you don't see the variable declarations in the context in which they're first used). But.... that would have most likely prevented this bug. So much for ignoring the wisdom of the BDFL!

Issue #369, opened 2015-11-24
BalancedConsumer raises an exception with not much details


tigran-a:
Hi,

I've a consumption loop 

```
#!/usr/bin/env python

import pykafka
from consumeconfig import *
import time

import logging
log= logging.getLogger(__file__)

console_handler = logging.StreamHandler()
console_handler.setLevel(logging.DEBUG)
formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
console_handler.setFormatter(formatter)
logging.getLogger("pykafka").addHandler(console_handler)

log.addHandler(console_handler)
GROUP = "GroupB"
client = pykafka.KafkaClient(hosts=KAFKA_HOSTS)
topic = client.topics[TOPIC]

consumer  = topic.get_balanced_consumer(consumer_group=GROUP, zookeeper_connect=ZOOKEEPER_HOSTS, auto_commit_enable=True)

need_reconn  = False
cntr = 0
while True:
        try:
                m = consumer.consume()
                if m is not None:
                        print ("%d | %s."%(m.offset,m.value))
                        print ("\t\t\t" + str( consumer.held_offsets) )
                        cntr+=1
                if cntr % 100 == 0:
                        consumer.commit_offsets()
        except Exception, e:
                log.error(10*"*"+"|" + str(e) + "|"+10*"*")
```

which works fine... for some time. After an exception is raised and I can't see any details of it. 
Here is a log: 
```
No handlers could be found for logger "kazoo.client"
2015-11-24 13:30:15,931 - pykafka.balancedconsumer - WARNING - Internal partition registry doesn't match ZooKeeper!
2015-11-24 19:01:30,732 - ./consume-many.py - ERROR - **********||**********
2015-11-24 19:01:30,733 - ./consume-many.py - ERROR - **********||**********
2015-11-24 19:01:30,733 - ./consume-many.py - ERROR - **********||**********
2015-11-24 19:01:30,733 - ./consume-many.py - ERROR - **********||**********
2015-11-24 19:01:30,733 - ./consume-many.py - ERROR - **********||**********
2015-11-24 19:01:30,733 - ./consume-many.py - ERROR - **********||**********
2015-11-24 19:01:30,733 - ./consume-many.py - ERROR - **********||**********
2015-11-24 19:01:30,733 - ./consume-many.py - ERROR - **********||**********
2015-11-24 19:01:30,733 - ./consume-many.py - ERROR - **********||**********
[...]

PR #382, opened 2015-12-07
remove checker worker from balancedconsumer

emmett9001:
This pull request removes the zookeeper checker thread from the `BalancedConsumer`. As discussed in #369, I'm pretty sure it doesn't actually do anything.


yungchin:
Cool. I've added one commit to toss out the helper you had already replaced in the tests.

PR #387, opened 2015-12-11
early unicode warnings

### LINKED ISSUE #386: warn early against unicode arguments
pykafka doesn't support unicode, which is fine. If client code supplies unicode in the wrong place, the error will manifest itself as something like this:
```
  File "bolts/common/kafka.py", line 58, in process_batch
    producer.produce(msg, partition_key=partition_key)
  File "/home/emmett/virtualenvs/ct-casterisk/local/lib/python2.7/site-packages/pykafka/producer.py", line 269, in produce
  File "/home/emmett/virtualenvs/ct-casterisk/local/lib/python2.7/site-packages/pykafka/producer.py", line 169, in _raise_worker_exceptions
    raise ex
error: argument for 's' must be a string
```
A similar error also causes metadata fetching to fail when a unicode topic name is supplied to `TopicDict.__getitem__` (via subscript access on `KafkaClient.topics`).

I think it's worth catching these cases immediately instead of allowing them to raise inscrutable exceptions on worker threads.

emmett9001:
This pull request makes unicode-related errors clearer to the end user in `TopicDict` and `produce()`. It fixes #386.

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -245,8 +245,11 @@ class Producer(object):
             message to
         :type partition_key: bytes
         """
+        if not (isinstance(partition_key, bytes) or partition_key is None):
+            raise TypeError("Producer.produce accepts a bytes object as partition_key, "
+                            "but it got '%s'", type(partition_key))
         if not (isinstance(message, bytes) or message is None):
-            raise TypeError("Producer.produce accepts a bytes object, but it "
+            raise TypeError("Producer.produce accepts a bytes object as message, but it "
                             "got '%s'", type(message))
         if not self._running:
             raise ProducerStoppedException()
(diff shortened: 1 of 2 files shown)


pykafka/producer.py
```diff
@@ -245,8 +245,11 @@ def produce(self, message, partition_key=None):
             message to
         :type partition_key: bytes
         """
+        if not (isinstance(partition_key, bytes) or partition_key is None):
+            raise TypeError("Producer.produce accepts a bytes object as partition_key, "
+                            "but it got '%s'", type(partition_key))
         if not (isinstance(message, bytes) or message is None):
-            raise TypeError("Producer.produce accepts a bytes object, but it "
+            raise TypeError("Producer.produce accepts a bytes object as message, but it "
```
yungchin: I'm wondering if it's useful to check the opposite for these: `if isinstance(message, unicode)` - so that people could in principle use anything that behaves as `bytes`?

PR #392, opened 2015-12-14
NoPartitionsForConsumer fix

### LINKED ISSUE #391: inaccurate NoPartitionsForConsumerException
Just spotted on Travis: https://travis-ci.org/Parsely/pykafka/jobs/96370900#L687

I believe this happens when a `BalancedConsumer` is in the middle of rebalancing, and the `consume()` call hits it just when it stopped its internal consumer - that is, it's more than a test quirk.

My idea for fixing this would be that we raise `NoPartitionsForConsumer` from the worker thread that ran `_rebalance()`, instead of from `consume()` (which wasn't possible when this was originally added, because we didn't have the re-raising logic in place then).

However, @emmett9001 I wonder whether we really care about keeping this exception at all? It seems to cover only the special case where we added more consumer instances than there are partitions - which should be something the user is aware of doing, and maybe they would want to actually be able to do this (to keep a "hot spare" that listens in)?

emmett9001:
The `NoPartitionsForConsumer` exception is usually raised when it shouldn't be. This pull request fixes #391 by having the consumer return `None` instead of raising this exception.

### DIFF
--- a/tests/pykafka/test_balancedconsumer.py
+++ b/tests/pykafka/test_balancedconsumer.py
@@ -8,7 +8,7 @@ from kazoo.client import KazooClient
 
 from pykafka import KafkaClient
 from pykafka.balancedconsumer import BalancedConsumer, OffsetType
-from pykafka.exceptions import NoPartitionsForConsumerException, ConsumerStoppedException
+from pykafka.exceptions import ConsumerStoppedException
 from pykafka.test.utils import get_cluster, stop_cluster
 from pykafka.utils.compat import range, iterkeys, iteritems
 
@@ -264,8 +264,8 @@ class BalancedConsumerIntegrationTests(unittest2.TestCase):
         consumer._decide_partitions = lambda p: set()
         consumer.start()
         self.assertFalse(consumer._running)
-        with self.assertRaises(NoPartitionsForConsumerException):
-            consumer.consume()
+        res = consumer.consume()
+        self.assertEqual(res, None)
 
     def test_zk_conn_lost(self):
         """Check we restore zookeeper nodes correctly after connection loss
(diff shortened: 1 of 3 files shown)


tests/pykafka/test_balancedconsumer.py
```diff
@@ -264,8 +264,8 @@ def test_no_partitions(self):
         consumer._decide_partitions = lambda p: set()
         consumer.start()
         self.assertFalse(consumer._running)
-        with self.assertRaises(NoPartitionsForConsumerException):
-            consumer.consume()
+        res = consumer.consume()
+        self.assertEqual(res, None)
```
yungchin: This test seems to have a contradiction now: in L266 we demand that the consumer has in fact stopped, but in that case L268 should see an exception?

yungchin:
Ok, I was thinking about this a bit longer, and I agree this pullreq takes the right approach by removing the exception. If you have n partitions, and you add n+1 consumers, one of them is going to be partitionless (and you can't predict which one because they are sorted by their randomly picked IDs) and there shouldn't be any problems as a result of that.

But then if that's not going to raise exceptions, I think that means the consumer shouldn't `stop()` (so basically that means all the `should_stop` logic would disappear from `_rebalance()`). Revisiting #313 I now think the main issue there was just that a partitionless consumer was "unstoppable", not that it should stop automatically.

So I think we would be ok if we're "watering down" `test_no_partitions()`: if you have a consumer that doesn't hold any partitions, the test should just check that `consume()` returns `None` and that a call to `stop()` won't fail.

emmett9001:
@yungchin I see your point, and I've adjusted the changeset here to reflect this logic. Again, feel free to merge if you like.

yungchin:
@emmett9001 I started some small tweaks thinking I'd merge it straight afterwards, but ended up changing more things, so now it's probably best if you could give it another quick pass if that's ok. The additional changes are both for edge cases: 1) if the consumer held zk nodes previously, it should release them, and 2) if the consumer is somehow not running, it shouldn't return `None`.

emmett9001:
Looks good, the remaining `break` statement was an oversight on my part. Thanks!

PR #397, opened 2015-12-18
message enqueue filtering

### LINKED ISSUE #389: Q: BalancedConsummer's commit_offsets() does not prevent from message double reception?
Hi, 

I've a balanced consumer that reads a flow from kafka.
At some point, e.g. every 1 second (for test purpose) I execute the code like:
```python
...
     consumer.consume() # note: non-blocking call, timeout 100ms
...
            if (time()-last_ts) >= 1:  # every second, for the example
                self.consumer.commit_offsets()
                self.consumer.stop() #?
                reconnected += 1
                log.warning("Recreating the consumer for the %d st/nd/th time", reconnected)
                self.connect(self.connect_retries)
                last_ts=time()
...
```
(i had to do it because of the failures when having long-running kafka consumer script) 

the self.connect looks like:
```python
        self.connect_retries = connect_retries
        ntries = 0
        self.kafka_client = None
        log.info("Connecting to kafka...")
        while (self.kafka_client is None): 
            try:
                self.kafka_client = pykafka.KafkaClient(hosts=self.kafka_hosts)
            except Exception, e: 
                t, v, tb = sys.exc_info()
                ntries +=1
                if (ntries > self.connect_retries) and (self.connect_retries != -1):
                    raise t, v, tb
                log.exception("Could not connect to kafka: %s. Retrying in 1 second. Try %d/%d.", str(e), ntries, self.connect_retries)
                sleep(1)

        log.info("Looking for a topic %s...", self.topic_name)
        while self.topic_name not in self.kafka_client.topics: 
            log.error("Could not find topic at the kafka brokers. Updating info and retrying in 1 second")
            sleep(1)
            self.kafka_client.update_cluster()
        self.topic = self.kafka_client.topics[self.topic_name]

        log.info("Creating a balanced consumer...")
        self.consumer  = self.topic.get_balanced_consumer(
               consumer_group=self.consumer_group, 
               zookeeper_connect=self.zoo_hosts, 
               consumer_timeout_ms=timeout_ms,
               auto_commit_enable=True)
```
Each consumed message in a test message set had unique id; and what i do with them then received, is inserting in a database. When I have this reconnection code, I got duplication primary key errors, meaning the consumer processed some messages twice (or more)

Could you please clarify if it is what is expected to happen? 

Note: if no reconnection is done during consumption, no duplication primary keys raised, confirming that the IDs of the test dataset are unique... 

Thanks

emmett9001:
This pull request ensures that the `SimpleConsumer` doesn't re-enqueue the last message it consumed. It fixes #389.

It's not entirely clear to me at the moment if there's some other logic error in the consumer initialization offset logic, but this at least appears to be a real bug and does actually fix #389.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -816,7 +816,7 @@ class OwnedPartition(object):
         :type messages: Iterable of :class:`pykafka.common.Message`
         """
         for message in messages:
-            if message.offset < self.last_offset_consumed:
+            if message.offset <= self.last_offset_consumed:
                 log.debug("Skipping enqueue for offset (%s) "
                           "less than last_offset_consumed (%s)",
                           message.offset, self.last_offset_consumed)
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -816,10 +816,10 @@ class OwnedPartition(object):
         :type messages: Iterable of :class:`pykafka.common.Message`
         """
         for message in messages:
-            if message.offset <= self.last_offset_consumed:
+            if message.offset != self.next_offset:
                 log.debug("Skipping enqueue for offset (%s) "
-                          "less than last_offset_consumed (%s)",
-                          message.offset, self.last_offset_consumed)
+                          "not equal to next_offset (%s)",
+                          message.offset, self.next_offset)
                 continue
             message.partition = self.partition
             if message.partition_id != self.partition.id:
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -720,7 +720,7 @@ class OwnedPartition(object):
         self.partition = partition
         self._messages = Queue()
         self._messages_arrived = semaphore
-        self.last_offset_consumed = 0
+        self.last_offset_consumed = -1
         self.next_offset = 0
         self.fetch_lock = threading.RLock()
 
--- a/tests/pykafka/test_simpleconsumer.py
+++ b/tests/pykafka/test_simpleconsumer.py
@@ -201,13 +201,15 @@ class TestSimpleConsumer(unittest2.TestCase):
 
 class TestOwnedPartition(unittest2.TestCase):
     def test_partition_saves_offset(self):
+        offset = 20
         msgval = "test"
         partition = mock.MagicMock()
         op = OwnedPartition(partition)
+        op.next_offset = offset
 
         message = mock.Mock()
         message.value = msgval
-        message.offset = 20
+        message.offset = offset
 
         op.enqueue_messages([message])
         self.assertEqual(op.message_count, 1)


yungchin:
@emmett9001 great catch, that was a super-puzzling issue. I think this is clearly a correct fix, and so I couldn't really figure out why the test suite disagreed. Turns out we had another bug that was masked by this one - I've added the fix on this branch.

Also: I went back to the kafka protocol docs, because I couldn't recall why a fetch response would contain earlier messages than requested. The relevant bit seems to be

> However, with compressed messages, it's possible for the returned messages to have offsets smaller than the starting offset. The number of such messages is typically small and the caller is responsible for filtering out those messages.

That is, the bug in #389 should manifest itself only if the messages have been compressed at produce time. We should probably try to turn the quoted requirement into a test case?

emmett9001:
@yungchin I wondered the same thing, and in my testing discovered that we actually are getting messages from before `last_offset_consumed` even when messages are uncompressed. Whether this points to a deeper issue in pykafka or kafka itself i'm not sure.

yungchin:
Ah ok, thanks. This might just be how kafka works - if consumers must be filtering anyway, they might've decided to also allow this optimisation for uncompressed messages. I guess it is much the same as the partial messages at the end of a fetch response: the server is optimised to just send you a raw segment it lifted off its disk, and there may be bits before and after that need trimming.

emmett9001:
It certainly wouldn't be the first time that document didn't reflect the actual state of the kafka codebase.

yungchin:
@emmett9001 hmmm I wonder if there's still an issue though: what if you `fetch()` twice, and the user didn't consume fast enough to reach the end of the partition between the two fetches? I think that would still allow a double-enqueue?

pykafka/simpleconsumer.py
```diff
@@ -816,7 +816,7 @@ def enqueue_messages(self, messages):
         :type messages: Iterable of :class:`pykafka.common.Message`
         """
         for message in messages:
-            if message.offset < self.last_offset_consumed:
+            if message.offset <= self.last_offset_consumed:
```
yungchin: How about if this would read `if message.offset < self.next_offset:`?

emmett9001:
@yungchin If we accept that fetches can get messages before the specified offset, then I believe you're right. I guess that means we also need to keep track of the highest offset enqueued and not allow messages with lower offsets to be re-enqueued.

pykafka/simpleconsumer.py
yungchin: (..... or even stricter: `message.offset` ought to always equal `self.next_offset` I suppose)

emmett9001: I just came to the same conclusion!

yungchin:
Looks all good to me now!

PR #403, opened 2015-12-23
return Message instances from produce

### LINKED ISSUE #398: Cannot wait until a certain message is delivered
Hi. I want to wait until a certain message is guaranteed to be delivered. Theoretically, I could do this by spinning `get_delivery_report` in a loop, and reading out the reports that come back. Since `produce` produces an internal `Message` which it doesn't return, I can't get anything to identify the resulting message. The best I can get is a binary compare on the `message` value, which could be incorrect if I send the same message multiple times.

If the producer returned the internal `Message` from `produce`, this would be OK. I could then wait until a certain message has been successfully delivered.

For now, I suppose I will call the private, internal `_wait_all` API.

emmett9001:
This pull request fixes #398 by returning the `Message` instance added to the `Producer`'s internal queue from a call to `produce()`. I can't see anything that this breaks, but I may be overlooking something. The benefit of this change is that it allows unambiguous matching of produced messages to those popped from the delivery report queue.

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -244,6 +244,8 @@ class Producer(object):
         :param partition_key: The key to use when deciding which partition to send this
             message to
         :type partition_key: bytes
+        :return: The :class:`pykafka.protocol.Message` instance that was
+            added to the internal message queue
         """
         if not (isinstance(partition_key, bytes) or partition_key is None):
             raise TypeError("Producer.produce accepts a bytes object as partition_key, "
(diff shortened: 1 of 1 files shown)


emmett9001:
Putting this on hold for the time being. It's pretty innocuous except for the fact that it exposes the `Message` interface, which we're not completely certain is ready to be exposed in a public API.

emmett9001:
I think it might be ok to merge this now. Any opinions @yungchin @kbourgoin?

PR #411, opened 2016-01-07
Capture and surface the error thrown by post_rebalance_callback

jofusa:
If the post_rebalance_callback throws an error, it was not getting reported. This pr wraps it call in a try catch and binds the exception to _worker_exception.

the `try... except Exception:` catchall makes me feel dirty, but i don't know what else to do. Any suggestions appreciated.


emmett9001:
I think the `except Exception` is ok in this case, especially because clients can put whatever they want in the callback.

PR #412, opened 2016-01-08
get_offset_manager return failure

### LINKED ISSUE #402: AttributeError: 'NoneType' object has no attribute 'id'
when the code ran for a while, it raised below exception:
```
No handlers could be found for logger "pykafka.simpleconsumer"
  File "test.py", line 8, in <module>
    for message in balanced_consumer:
  File "/usr/lib/python2.7/site-packages/pykafka/balancedconsumer.py", line 707, in __iter__
    message = self.consume(block=True)
  File "/usr/lib/python2.7/site-packages/pykafka/balancedconsumer.py", line 693, in consume
    message = self._consumer.consume(block=block)
  File "/usr/lib/python2.7/site-packages/pykafka/simpleconsumer.py", line 371, in consume
  File "/usr/lib/python2.7/site-packages/pykafka/simpleconsumer.py", line 203, in _raise_worker_exceptions
    raise ex
AttributeError: 'NoneType' object has no attribute 'id'
Exception AttributeError: "'NoneType' object has no attribute 'id'" in <bound method BalancedConsumer.__del__ of <pykafka.balancedconsumer.BalancedConsumer at 0x28ec8d0 (consumer_group=test2)>> ignored
```

my code is:
```python
from pykafka import KafkaClient
client = KafkaClient(hosts="XXXXX")
topic = client.topics['topic']
balanced_consumer = topic.get_balanced_consumer(
    consumer_group='test2',
    auto_commit_enable=True,
    num_consumer_fetchers=5,
    reset_offset_on_start=False,
    fetch_message_max_bytes=1048576*10,
    zookeeper_connect='XXXXX')
for message in balanced_consumer:
    if message is not None:
        print "offset: " + str(message.offset)
    else: 
        pass
```

emmett9001:
This pull request fixes #402 by removing a case in which `get_offset_manager` could return `None`. Specifically, this case occurs when `SocketDisconnectedError` is encountered on the last retry of the loop.

### DIFF
--- a/pykafka/cluster.py
+++ b/pykafka/cluster.py
@@ -364,6 +364,8 @@ class Cluster(object):
                 log.error("Socket disconnected during offset manager "
                           "discovery. This can happen when using PyKafka "
                           "with a Kafka version lower than 0.8.2.")
+                if i == MAX_RETRIES - 1:
+                    raise
                 self.update()
             else:
                 coordinator = self.brokers.get(res.coordinator_id, None)


yungchin:
That fix seems good to me, even regardless if it explains #402 (but I'd also wager that it does).

One other thing I noticed, which may be related: we pick a `broker` outside the loop. If the reason we hit `SocketDisconnectedError` is that this broker is in some confused state (eg as per KAFKA-1387), `update()` might not fix that, and we'd be better off trying to ask a different broker on the next iteration?

Issue #390, opened 2015-12-11
Group Membership API protocol implementation


emmett9001:
Support the group membership protocol introduced in 0.9 in the protocol implementation.

https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-GroupMembershipAPI

emmett9001:
I've made a lot of progress on #416. `ManagedBalancedConsumer` (bad name) now minimally extends `BalancedConsumer` and implements 0.9's rebalance semantics. It still requires polishing, a lot of testing, and proper error handling on `JoinGroupRequest` and `SyncGroupRequest`. @rduplain @yungchin It's now far enough along that it's not unreasonable to take a look at what I've made and provide some thoughts on the form it's taking.

PR #416, opened 2016-01-12
 Group Membership API

emmett9001:
This pull request implements support for the new [consumer group management protocol](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-GroupMembershipAPI) in Kafka 0.9 as the `ManagedBalancedConsumer` class, new request/response support in `protocol.py`, and a new interface on `Topic`.

Changes to the public API are minimal: several new keyword arguments on `BalancedConsumer` and `SimpleConsumer`, as well as on `Topic.get_balanced_consumer(managed=True)`.

Other notable changes:
* Monkeypatches around a bug in py.test related to `skipif`
* Adds functionality to `Broker` to allow multiple `RequestHandlers` on a single instance
* Refactors internals of the `BalancedConsumer` rebalance process for extensibility
* Uses `gevent.socket` when appropriate

Here's the [kafka-python implementation](https://github.com/dpkp/kafka-python/blob/master/kafka/consumer/group.py) for reference.

Related to #390


pykafka/broker.py
```diff
@@ -348,3 +348,25 @@ def fetch_consumer_group_offsets(self, consumer_group, preqs):
             self.connect_offsets_channel()
         req = OffsetFetchRequest(consumer_group, partition_requests=preqs)
         return self._offsets_channel_req_handler.request(req).get(OffsetFetchResponse)
+
+    ##########################
+    #  Group Membership API  #
+    ##########################
```
rduplain: Consider renaming these methods to match the names of the requests in [the docs](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-GroupMembershipAPI):

* `join_group`
* `leave_group`
* `sync_group`
* `heartbeat`

pykafka/managedbalancedconsumer.py
```diff
@@ -0,0 +1,368 @@
...
+        :type post_rebalance_callback: function
+        :param use_rdkafka: Use librdkafka-backed consumer if available
+        :type use_rdkafka: bool
+        :param compacted_topic: Set to read from a compacted topic. Forces
+            consumer to use less stringent message ordering logic because compacted
+            topics do not provide offsets in stict incrementing order.
+        :type compacted_topic: bool
+        :param heartbeat_interval_ms: The amount of time in milliseconds to wait between
+            heartbeat requests
+        :type heartbeat_interval_ms: int
```
rduplain: This appears to be copied over from the base class. Ideally, we don't repeat details. There may be decent tools out there for this, but in a pinch, you could grab `__doc__` from the super and merge in the local docstring. This could e.g. be done in a decorator.

pykafka/managedbalancedconsumer.py
```diff
@@ -0,0 +1,368 @@
...
+        if res.error_code in (IllegalGeneration.ERROR_CODE,
+                              RebalanceInProgress.ERROR_CODE,
+                              UnknownMemberId.ERROR_CODE):
+            pass
+        elif res.error_code in (GroupCoordinatorNotAvailable.ERROR_CODE,
+                                NotCoordinatorForGroup.ERROR_CODE):
+            self._group_coordinator = self._cluster.get_group_coordinator(
+                self._consumer_group)
+        elif res.error_code == GroupAuthorizationFailed.ERROR_CODE:
+            raise GroupAuthorizationFailed()
```
rduplain: Could we get this error code logic into some sort of response manager or wrapper? There will certainly be common error cases across all responses.

rduplain: Looking further into the diff, I see the response classes now. This logic could be folded into those classes, with a base response class that deals with common errors.

tests/pykafka/test_protocol.py
```diff
@@ -256,5 +256,95 @@ def test_offset_fetch_response(self):
...
 
+class TestGroupMembershipAPI(unittest2.TestCase):
+    maxDiff = None
+
+    def test_consumer_group_protocol_metadata(self):
+        meta = protocol.ConsumerGroupProtocolMetadata()
+        msg = meta.get_bytes()
+        self.assertEqual(
+            msg,
+            bytearray(b'\x00\x00\x00\x00\x00\x01\x00\ndummytopic\x00\x00\x00\x0ctestuserdata')
```
rduplain: To make this more readable to humans, you could consider stuffing data into the docstring or another string which is converted into the bytearray, i.e. write a utility to convert this:

```python
"""
00 00 00 00 00 01 00
dummytopic
00 00 00 0c
testuserdata
```

```python
bytearray(b'\x00\x00\x00\x00\x00\x01\x00\ndummytopic\x00\x00\x00\x0ctestuserdata')
```

In other words, the utility trims whitespace and converts `\d+` to bytes. This kind of utility pays off quickly for writing tests when developing protocol drives.

rduplain:
There's a lot of good stuff here.

My primary architectural comment is to build more into your request-response abstractions to let them do more work for you. Ideally, each step up in abstraction reduces the number of cases that you have to know. Exceptions can bubble up transparently (I'll comment in the diff where this makes sense).

I was glad to see the `ManagedBalancedConsumer` be a subclass of `BalancedConsumer`.

pykafka/managedbalancedconsumer.py
rduplain: If the logic is folded into a utility or the response class, then `raise GroupAuthorizationFailed()` just happens in the call stack and you wouldn't worry about it here.

pykafka/managedbalancedconsumer.py
```diff
@@ -0,0 +1,368 @@
...
+        """Start the heartbeat worker"""
+        self = weakref.proxy(self)
+
+        def fetcher():
+            while True:
+                try:
+                    if not self._running:
+                        break
+                    self._send_heartbeat()
+                    self._cluster.handler.sleep(self._heartbeat_interval_ms / 1000)
```
yungchin: This might need a `from __future__ import division`.

pykafka/managedbalancedconsumer.py
```diff
@@ -0,0 +1,368 @@
...
+                except ReferenceError:
+                    break
+                except Exception:
+                    # surface all exceptions to the main thread
+                    self._worker_exception = sys.exc_info()
+                    break
```
yungchin: Do we care enough that if `_send_heartbeat()` fails, we'd want to abort the consumer altogether? I'm not really sure what my answer would be, but IIUC it's a bit like the decision we made earlier for `BalancedConsumer`: "if you're not certain if you're still a group member, just keep consuming until it's clear you're not"?

pykafka/managedbalancedconsumer.py
```diff
@@ -0,0 +1,368 @@
...
+                group_assignments = [
+                    MemberAssignment([
+                        (self._topic.name,
+                            self._decide_partitions(iterkeys(members),
+                                                    consumer_id=member_id))
+                    ], member_id=member_id) for member_id in members]
+
+                assignment = self._sync_group(group_assignments)
+                my_partitions = [p for p in itervalues(self._topic.partitions)
+                                 if p.id in assignment[0][1]]
```
yungchin: Just a nitpick: I think this could be written `[self._topic.partitions[pid] for pid in assignment[0][1]`.

pykafka/simpleconsumer.py
```diff
@@ -157,6 +164,8 @@ def __init__(self,
         self._auto_start = auto_start
         self._reset_offset_on_start = reset_offset_on_start
         self._is_compacted_topic = compacted_topic
+        self._generation_id = generation_id
+        self._consumer_id = consumer_id
```
yungchin: Does this need a default string if `None`?

yungchin:
Good stuff. I only had a few minor comments on the diff, and certainly nothing on structure. I do think @rduplain's suggestion for error handling abstractions sounds good, as all the conditionals are challenging to navigate right now.

As for testing: it should be possible to make a class that inherits all its tests from `TestBalancedConsumer`, which would practically cover all expected behaviour for the new class as well, if that's a good idea?

pykafka/managedbalancedconsumer.py
emmett9001: I agree that we should address the duplication of docstrings across the project, but I think it's outside the scope of this pull request. It already happens in a few [other places](https://github.com/Parsely/pykafka/blob/79e8938f55484e4c892744ec203e98998c2f7f2d/pykafka/simpleconsumer.py#L79), and I'm just following that convention for now. I've opened an [issue](https://github.com/Parsely/pykafka/issues/473) to remind us to do this deduplication in a separate pull request after this is in.

pykafka/managedbalancedconsumer.py
emmett9001: I agree, an error handling cleanup is something the whole codebase could potentially benefit from. To do it right will require comparing all of the error handling code across pykafka so we know how to best abstract it. This will add a lot of extra code to this pull request, so I think it's out of scope. Here's an [issue](https://github.com/Parsely/pykafka/issues/474) to track our progress on that.

I'd really like master to remain consistent in how it handles errors, and right now there is very little error handling happening in `protocol.py`. That said, there is definitely some refactoring I can do here to deduplicate some of the error code logic in this class.

tests/pykafka/test_protocol.py
emmett9001: Sounds like a great idea. Again, the tests in this pull request are consistent with the rest of the existing protocol tests. I think overhauling the way we do protocol testing is outside the scope of this pull request. I've opened an [issue](https://github.com/Parsely/pykafka/issues/475) about this as well.

pykafka/managedbalancedconsumer.py
emmett9001: The way that the 0.9 group membership semantics work, if a consumer stops sending heartbeats, it it by definition out of the group. Thus I do think it makes sense to fail immediately when that case occurs.

pykafka/managedbalancedconsumer.py
rduplain: I'm okay with that, for docstring duplication issues.

pykafka/managedbalancedconsumer.py
rduplain: I can see why you might see it as out-of-scope -- we could use some abstractions that may be far reaching. However, I'm proposing changes that are completely isolated to code added by this pull request. My suggestion is to move these if-statements into the response classes that you've added.

tests/pykafka/test_protocol.py
rduplain: For this one as well, I'm proposing what is a simple testing utility that will improve the tests being added by this pull request. It's a judgement call, but I'll advocate that we look to improve utilities on every opportunity instead of pushing them into some future.

rduplain: Since bytes can be painful in Python, we could forego the idea of the utility and just use indenting to improve readability:

```python
xs = bytearray(
    b'\x00\x00\x00\x00\x00\x01\x00\n'
    b'dummytopic'
    b'\x00\x00\x00\x0c'
    b'testuserdata')
```

I split lines based on the data contained. It will make the most sense to split lines based on what makes sense in the protocol (i.e. number of bytes or on a sentinel byte).

emmett9001: That's a great way to easily make these tests a lot more readable.

emmett9001:
Update: I've tracked down one of the biggest remaining test failures. Unfortunately it looks like it will require a minor overhaul to this diff. The issue appears to be that because of synchronous network requests like [this](https://github.com/Parsely/pykafka/blob/9bc6d8c1545321c96c002eef3da8f08759351a30/pykafka/broker.py#L365), starting multiple `ManagedBalancedConsumer`s in the same process doesn't work properly. This was not an issue in the `SimpleConsumer` because there is no situation in which it matters to one `SimpleConsumer` that its request completes before another `SimpleConsumer` makes a particular request. This was also not an issue in `BalancedConsumer` because all synchronous communication was moderated by ZooKeeper, allowing consumer instances not to care about the particular sequencing of their requests relative to those of others.

The difference is that the semantics for `ManagedBalancedConsumer` specify that all consumers in a group must send their `JoinGroupRequest`s before the session timeout. When more than one `ManagedBalancedConsumer` is run in the same process, currently the following sequence of events will occur:

* The first consumer starts normally using synchronous requests
* The second consumer sends its `JoinGroupRequest` and blocks for the session timeout
* The first consumer's heartbeat thread sends a heartbeat request and blocks
* The second consumer's `JoinGroupRequest` completes after waiting the full timeout without any additional joins
* The second consumer does as much work as it can before making another network request; namely, it makes it all the way to its `SyncGroupRequest`
* The first consumer's heartbeat thread takes this opportunity to unblock and register that a rebalance has occurred.
* The first consumer continues with its normal rebalance procedure until it encounters `UnknownMemberId`, which happens because the join group phase completed before its heartbeat request was allowed to return.

I'm not sure yet how I'll go about fixing this, though I'm pretty confident it will require a serious overhaul. I'm hoping that I'll be able to leverage the fact that the requests already return futures.

emmett9001:
Looks like my above assessment was off the mark (predictably). After digging in further, it appears that the real issue is that consumers in the same process share a single `RequestHandler` instance. Since `RequestHandler` queues all requests before making them serially, this renders the request interleaving necessary for the group join phase impossible. My [most recent commit](https://github.com/Parsely/pykafka/commit/45b2349af13fe0111b55bacb3d2192eaece6511d) attempts to address this issue by maintaining `RequestHandler` instances unique to each consumer.

emmett9001:
Current gevent test failures are probably directly related to the fact that normal `socket`-based requests will [block serially](https://twitter.com/ossmkitty/status/705889991449120768) under gevent. Fixing this will be a lot of work - I'm considering the merits of leaving it for a later pull request and simply disabling the group membership API under gevent for the time being.

emmett9001:
List of existing bugs that didn't become apparent until this changeset:

* The py.test subclass skipping issue [patched here](https://github.com/Parsely/pykafka/blob/c5f602c487ae702fe6c3ff6b9bd5e99ea167ac22/tests/pykafka/__init__.py#L4)
* The lack of [switchable socket implementations](https://github.com/Parsely/pykafka/blob/7ecb5fb4bd82bdd7f589c44d56fd9dc793b4063a/pykafka/handlers.py#L85) for gevent
* The incorrect [caching](https://github.com/Parsely/pykafka/commit/cf2ff7aa69898ec2fb341522d9ec32a265224e9c) of kafka binaries, which caused us to not test against 0.9 for a brief period
* The [sharing](https://github.com/Parsely/pykafka/commit/45b2349af13fe0111b55bacb3d2192eaece6511d) of `RequestHandler` instances by multiple consumers/producers in the same process

Now that these are all figured out, this pull request is *finally* ready for a final review. @yungchin @rduplain @amontalenti @kbourgoin please feel free to make any comments or concerns about this diff known. I'd love to have a better name than `ManagedBalancedConsumer`, but I can't actually think of a great one.

rduplain:
Looks good to me. @yungchin?

As for the name, I don't have an immediate recommendation, but I can offer an observation. `BalancedConsumer` would likely be a base class if we designed pykafka today, with subclasses each for the implementation using zookeeper and the membership APIs (i.e. something like `ZookeeperBalancedConsumer` and `KafkaBalancedConsumer` would be the boring names).

yungchin:
@emmett9001 a question about the unique request handlers: iiuc you only need this for the case where multiple consumers of a single `KafkaClient` join the same group, is that right? So I'm wondering if you could get away with just bluntly disallowing that instead? That would avoid all the connection-id complexity.

My thinking is that in the typical use case, you'd always run each consumer group member from a different box (or, at least, from a different process). But maybe there's a use case I'm not thinking of.

(The test cases would become a bit more complex, because consumer_a and consumer_b would have to come from different `KafkaClient` instances.)

Other than that, no real nags from me. Big kudos for all the test suite work, especially also discovering and fixing the subclassing issues!!

emmett9001:
@yungchin If you assume that multiple consumers in the same group in the same process is not a real use case, then it's just a tradeoff between test complexity and library complexity. It might be better to just disallow it entirely, but I'll have to think through it.

emmett9001:
I think I'm going to keep the multiple connections logic for now. I've seen multiple consumers per process used in various issue reports in the past, and disallowing it can cause an unnecessary hurdle when debugging certain issues.

We have 0.9 group support now!

PR #421, opened 2016-01-19
RLock verbosity

### LINKED ISSUE #407: Cannot create producer on Python 3.5
I'm trying to create a simple producer in Python 3.5 and come across the following error message:
```
  File "/app/.tox/py35/lib/python3.5/site-packages/pykafka/topic.py", line 88, in get_producer
    return Cls(self._cluster, self, **kwargs)
  File "/app/.tox/py35/lib/python3.5/site-packages/pykafka/producer.py", line 152, in __init__
    self.start()
  File "/app/.tox/py35/lib/python3.5/site-packages/pykafka/producer.py", line 189, in start
    self._setup_owned_brokers()
  File "/app/.tox/py35/lib/python3.5/site-packages/pykafka/producer.py", line 228, in _setup_owned_brokers
    self, partition.leader)
  File "/app/.tox/py35/lib/python3.5/site-packages/pykafka/producer.py", line 426, in __init__
    self.lock = self.producer._cluster.handler.RLock()
  File "/usr/lib/python3.5/threading.py", line 82, in RLock
    return _PyRLock(*args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'verbose'
```

I've tracked the problem down to the following line in `pykafka/handlers.py`:
```python
    # turn off RLock's super annoying default logging
    RLock = functools.partial(threading.RLock, verbose=False)
```

So, what this seems to be doing is instantiating a new RLock, which is fine. But if we take a look at python 3.5's `threading.py` we see the following (I removed unnecessary info):

```python
try:
    _CRLock = _thread.RLock
except AttributeError:
    _CRLock = None
 
   ...

def RLock(*args, **kwargs):
    if _CRLock is None:
        return _PyRLock(*args, **kwargs)
    return _CRLock(*args, **kwargs)

  ...

class _RLock:
    def __init__(self):
       ...

   ...

_PyRLock = _RLock
```

As you can see, if we don't have a CRLock, we end up using a PyRLock, which does not take any arguments. This causes the producer to fail being created since it gets an unexpected keyword argument.

emmett9001:
This pull request fixes #407 by turning off RLock verbose logging in a way that doesn't break 3.5 compatibility.

### DIFF
--- a/pykafka/handlers.py
+++ b/pykafka/handlers.py
@@ -19,7 +19,6 @@ limitations under the License.
 __all__ = ["ResponseFuture", "Handler", "ThreadingHandler", "RequestHandler"]
 
 from collections import namedtuple
-import functools
 import logging
 import threading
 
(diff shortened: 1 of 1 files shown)


timmartin19:
Thank you!

PR #426, opened 2016-01-25
log exception info on balanced consumer start failure

kalotay:
This helps finding out the cause for balanced consumer start failures.


yungchin:
Hey Dan, long time no speak!
Thanks for this, that's a useful tweak.

PR #431, opened 2016-01-26
Fixed linger_ms to match documentation.

pdex:
`self.flush_ready.wait(None)` will actually linger forever until the `flush_ready` condition is notified. The documentation for linger_ms specifies that a linger_ms of 0 will not linger.


emmett9001:
@pdex Thanks for this fix. I think the existing error is [my fault](https://github.com/Parsely/pykafka/commit/eb0f0fcd12aaaa58a241afdc1ba8382440c2370f) after another bugfix. That fix did correct the issue I was after, but I got the meaning of `None` wrong.

PR #441, opened 2016-02-04
Cache pip in Travis

thedrow:


yungchin:
Thanks very much for this. We'd been meaning to look into the caching options for a while. One concern we had was that it might potentially cause oddities when the cache is stale - so one question @thedrow: do you know if clearing the travis cache is quickly done?

Also, provided that @emmett9001 is also happy with caching, I'd have another feature req for this PR - could we also cache the `KAFKA_BIN` path (where testinstances puts the kafka and zk binaries)?

Many thanks!

thedrow:
You can easily clear the cache from the settings tab or the command line if something goes wrong which is almost never, at least with pip.

thedrow:
I added a cache to $HOME/kafka-bin. Not sure if it will speed things up since the cache is just S3.

thedrow:
We can also cache the compilation of librdkafka. I'll try that as well.

emmett9001:
I just ran into trouble with the kafka binary caching in #416. Since a single Travis build needs to test multiple versions of kafka, I'll remove the caching on that directory for now.

thedrow:
The problem is with Travis. They don't namespace cache buckets according to environment variables correctly. It's a known issue and they are working on a fix.

PR #442, opened 2016-02-04
Test in Travis with the latest Kafka 0.8.x release

thedrow:


emmett9001:
@thedrow What does this get us that testing with 0.8.2.1 doesn't? I'd prefer to test the oldest possible compatible release if we're only going to test one of the 0.8.2 series.

thedrow:
Usually people prefer to test with the latest stable release.
Feel free to close this PR if it's not required.

emmett9001:
@rduplain Do you have an opinion on this? I don't feel too strongly one way or the other, but I'd like to avoid testing both bugfix versions.

thedrow:
@emmett9001 There was a serious snappy bug fixed in 0.8.2.2.
We should recommend upgrading anyway.

thedrow:
:+1:

PR #445, opened 2016-02-05
Test with librdkafka 0.9.x as well

thedrow:


thedrow:
It's either a performance regression or a bug on our part because the build just hangs for 0.9.x.

emmett9001:
@thedrow Are you usiing librdkafka 0.9.x successfully in production? I wouldn't be surprised if this were some bug in our tests, though I'm not sure.

thedrow:
No I'm not using Kafka at all yet. Just felt like contributing to something useful.

yungchin:
Thanks @thedrow - not sure what this is yet, but it fails consistently across python versions, so I should be able to tell you more soon.

yungchin:
PS: @emmett9001 once we do get this working with librdkafka 0.9 (which we're upgrading to anyway for TLS support), would you be happy if we just bump the version dependency, and drop 0.8.6 from testing?

emmett9001:
@yungchin Based on the precedent set in #442, I'd be fine with that.

thedrow:
Hi guys, 0.9 is considered the testing version and 0.8 as the stable version.
I don't think dropping support is required just yet. Maybe when 0.9 hits the stable mark.

yungchin:
Quick update here: I can reproduce the failure in `test_reset_offsets()`, and additionally see a failure in `test_consumer_lag()`. I'm git-bisecting against librdkafka to see what's changed and if we can adapt to it, but it's a bit of a slow process. To be continued.

yungchin:
Another update: the `test_consumer_lag()` failure seems to be due to my test cluster not being completely clean between test runs (that's why that one doesn't show up on Travis), so it's just a test artefact.

I've tried to narrow down what changes in librdkafka could explain the `test_reset_offsets()` failure and filed the issue at https://github.com/edenhill/librdkafka/issues/559

yungchin:
The test issue was fixed in librdkafka. All tests pass now.

I'm not sure what the best way forward would be with the Travis matrix. I don't think we really need to test all possible combinations of dependencies. On the other hand, as @thedrow said, 0.9 is not considered stable yet, so it makes sense to keep 0.8.6 here, yet we need 0.9 for the SSL branch...

emmett9001:
@thedrow Please merge master into your branch if you'd like this pull request to be merged.

thedrow:
Done.

thedrow:
Is this stale? Should we close it?

emmett9001:
Yes, this has been obsoleted by the merge of #477

PR #447, opened 2016-02-05
Disallow compiling the C extension when running with PyPy

thedrow:
Also added missing PyPi trove classifiers for CPython and PyPy.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ limitations under the License.
 import re
 import sys
 import os
+import platform
 
 from distutils.command.build_ext import build_ext
 from distutils.errors import CCompilerError, DistutilsExecError, DistutilsPlatformError
@@ -24,6 +25,11 @@ from setuptools import setup, find_packages
 from setuptools.command.test import test as TestCommand
 from setuptools.extension import Extension
 
+if sys.version_info < (2, 7):
+    raise Exception('pykafka requires Python 2.7 or higher.')
+
+python_implementation = platform.python_implementation()
+cpython = python_implementation == 'CPython'
 
 # Get version without importing, which avoids dependency issues
 def get_version():
@@ -162,6 +168,8 @@ def run_setup(with_rdkafka=True):
     )
 
 try:
+    if not cpython:
+        raise ve_build_ext.BuildFailed("librdkafka is not supported under %s" % python_implementation)
     run_setup()
 except ve_build_ext.BuildFailed as exc:
     print(15 * "-")
--- a/setup.py
+++ b/setup.py
@@ -72,6 +72,10 @@ class ve_build_ext(build_ext):
         def __init__(self):
             self.cause = sys.exc_info()[1]  # work around py 2/3 different syntax
 
+    class UnsupportedPlatform(Exception):
+        def __init__(self, python_implementation):
+            super(ve_build_ext.UnsupportedPlatform, self).__init__("librdkafka is not supported under %s" % python_implementation)
+
     def run(self):
         try:
             build_ext.run(self)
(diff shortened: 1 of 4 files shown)


thedrow:
Sorry about the failure. I can rebase if needed.

emmett9001:
@thedrow Looks like the PyPy failures are caused by attempting to import rdkafka-related test files that shouldn't even be imported. An easy way to fix this might be to wrap the [`_rd_kafka` imports](https://github.com/Parsely/pykafka/blob/2.2.0/pykafka/rdkafka/producer.py#L8) in `try..except`. It would probably be better to not run the tests under PyPy altogether, but I'm not sure how easy Travis will make that for us.

setup.py
```diff
@@ -24,6 +25,11 @@
 from setuptools.command.test import test as TestCommand
 from setuptools.extension import Extension
 
+if sys.version_info < (2, 7):
+    raise Exception('pykafka requires Python 2.7 or higher.')
+
+python_implementation = platform.python_implementation()
+cpython = python_implementation == 'CPython'
```
emmett9001: A better variable name here would be `is_cpython`

emmett9001:
Aside from my comment and the failing tests, this looks great. Let me know if you manage to make the PyPy tests pass. Thanks!

thedrow:
So the tests hang again. See https://travis-ci.org/Parsely/pykafka/jobs/107402230#L496

thedrow:
@emmett9001 Can you check the tests again? This has nothing to do with PyPy.

setup.py
```diff
@@ -162,6 +174,8 @@ def run_setup(with_rdkafka=True):
     )
 
 try:
+    if not is_cpython:
+        raise ve_build_ext.UnsupportedPlatform(python_implementation)
```
emmett9001: Would this `raise..except` pair make more sense as a simple boolean check on `is_cpython`? It seems like that would make the logic clearer.

tests/pykafka/rdkafka/test_producer.py
```diff
@@ -3,11 +3,16 @@
...
+try:
+    from pykafka.rdkafka import _rd_kafka
+    RDKAFKA = True
+except ImportError:
+    RDKAFKA = False # C extension not built
 
 @pytest.mark.skipif(platform.python_implementation() == "PyPy",
                     reason="We pass PyObject pointers as msg_opaques for "
                            "delivery callbacks, which is unsafe on PyPy.")
+@pytest.mark.skipif(not RDKAFKA, reason="C extension for librdkafka not built.")
```
emmett9001: Remove trailing whitespace

tests/pykafka/rdkafka/test_rd_kafka_consumer.py
```diff
@@ -1,12 +1,20 @@
...
 from pykafka.exceptions import RdKafkaStoppedException, RdKafkaException
-from pykafka.rdkafka import _rd_kafka
+try:
+    from pykafka.rdkafka import _rd_kafka
+    RDKAFKA = True
+except ImportError:
+    RDKAFKA = False # C extension not built
 from pykafka.test.utils import get_cluster, stop_cluster
 from pykafka.utils.compat import get_bytes
+import pytest
```
emmett9001: This import should be above, in the same group as `unittest2`

tests/pykafka/rdkafka/test_rd_kafka_consumer.py
```diff
@@ -1,12 +1,20 @@
...
 
-
+@pytest.mark.skipif(platform.python_implementation() == "PyPy",
+                    reason="librdkafka is not supported with PyPy.")
```
emmett9001: I actually don't think we need to skip this test under PyPy - it's passing in master.

emmett9001:
@thedrow Yup, the tests are looking better now. I left a few minor comments.

emmett9001:
@thedrow After the comments I've left are addressed, this will be mergeable.

tests/pykafka/rdkafka/test_rd_kafka_consumer.py
thedrow: isort says otherwise. I can change that though.

emmett9001:
Looks like this pull request is still using the naive method for test inheritance. See the docstring [here](https://github.com/Parsely/pykafka/blob/bb8363392f665b4e322f5e7bb7e540df31bea3cd/tests/pykafka/__init__.py#L4) for the proper way to do test inheritance, as well as an example [here](https://github.com/Parsely/pykafka/blob/bb8363392f665b4e322f5e7bb7e540df31bea3cd/tests/pykafka/test_balancedconsumer.py#L482) and [here](https://github.com/Parsely/pykafka/blob/bb8363392f665b4e322f5e7bb7e540df31bea3cd/tests/pykafka/test_balancedconsumer.py#L385).

emmett9001:
@thedrow If you'd like this pull request to get merged, please address the above comment about test inheritance. After that's done, this should be mergeable, but it won't be until then.

emmett9001:
Closed in favor of #715

PR #450, opened 2016-02-05
sync producer worker hang

### LINKED ISSUE #449: synchronous producer hangs when kafka is unreachable
_This issue was brought to my attention by Michael Kowalchik in the [mailing list](https://groups.google.com/forum/#!topic/pykafka-user/iyZlHj_NTDU)_

**Steps to reproduce**
- Run a synchronous producer pointing at a kafka cluster with a single node
- Kill the single kafka node while the producer is running
- Observe that the producer's main thread hangs and must be killed externally

The expected result is that the producer's main thread should raise a `SocketDisconnectedError` from `produce()`.

The following sequence of events causes this bug. When a broker is unavailable, `SocketDisconnectedError` is raised and caught. If there are no brokers available, [this call to _update()](https://github.com/Parsely/pykafka/blob/2.1.2/pykafka/producer.py#L366) raises an [exception](https://github.com/Parsely/pykafka/blob/2.1.2/pykafka/cluster.py#L286). This in turn [kills the worker thread](https://github.com/Parsely/pykafka/blob/2.1.2/pykafka/producer.py#L443), meaning that code after the `_update` call, including pushing the original `SocketDisconnectedError` onto the delivery reports queue, does not execute. On the main thread, `produce()` is still waiting for the [delivery report](https://github.com/Parsely/pykafka/blob/2.1.2/pykafka/producer.py#L268), which will never come because the worker has died.

Possible solutions include ignoring certain exception types from the `_update()` call, or catching any exception that happens between `to_retry` being populated and the delivery report being enqueued. I'd lean toward simply catching and warning about [this specific exception](https://github.com/Parsely/pykafka/blob/2.1.2/pykafka/cluster.py#L286), maybe giving it a more specific name in the process.

emmett9001:
This pull request fixes #449 by keeping the producer worker thread alive long enough to fill the delivery report queue for the synchronous check in `produce()`.

### DIFF
--- a/pykafka/cluster.py
+++ b/pykafka/cluster.py
@@ -29,6 +29,7 @@ from kazoo.client import KazooClient
 from .broker import Broker
 from .exceptions import (ERROR_CODES,
                          ConsumerCoordinatorNotAvailable,
+                         NoBrokersAvailableError,
                          SocketDisconnectedError,
                          LeaderNotAvailable)
 from .protocol import ConsumerMetadataRequest, ConsumerMetadataResponse
@@ -271,7 +272,7 @@ class Cluster(object):
                     return metadata
 
         # Couldn't connect anywhere. Raise an error.
-        raise RuntimeError(
+        raise NoBrokersAvailableError(
             'Unable to connect to a broker to fetch metadata. See logs.')
 
     def _get_brokers_from_zookeeper(self, zk_connect):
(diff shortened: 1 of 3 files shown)


yungchin:
Also and/or alternatively, would it be good to call `get_delivery_report()` with some timeout value, so that it couldn't hang the produce call? So something like:

```python
while True:
    try:
        report = self.get_delivery_report(timeout=1)
        # do stuff with report
        break
   except Empty:
        continue
```
(I haven't tested that.)

emmett9001:
@yungchin I added your idea to this changeset. Let me know what you think.

rduplain:
Code looks good to me.

Issue #428, opened 2016-01-25
KafkaClient timeout retries


emmett9001:
I've seen some users, including client code at Parsely, doing things like this to get around possible `socket.timeout` errors from `KafkaClient`.

```python
def _get_kafka_client(self, kafka_hosts, retries=2):
    fails = 0
    while True:
        try:
            return KafkaClient(hosts=kafka_hosts, socket_timeout_ms=30 * 1000)
        except socket.timeout:
            if fails == retries:
                raise
            fails += 1
            log.warning('Connection to kafka failed. Retry #%d', fails)
```

Does this logic have a place in pykafka? My thought is that we might want to add similar logic to `Cluster._request_metadata`.

PR #453, opened 2016-02-06
unify retry logic in Cluster

emmett9001:
This pull request addresses #428 by adding some rudimentary connection retrying logic to the metadata fetching section of broker discover. It also unifies the timeouts used between the zookeeper and kafka broker discovery methods and uses a single member attribute to keep track of the retry limit for each component of `Cluster`.

### DIFF
--- a/pykafka/cluster.py
+++ b/pykafka/cluster.py
@@ -61,13 +61,13 @@ class TopicDict(dict):
             return topic_ref()
         else:
             # Topic exists, but needs to be instantiated locally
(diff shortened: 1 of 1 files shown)


emmett9001:
@yungchin Any thoughts on this?

yungchin:
I see nothing but improvements here. Half-wondering if these retry loops could do with a `sleep()` somewhere, but that's really a question rather than a suggestion.

PR #454, opened 2016-02-06
GroupLoadInProgress exception

### LINKED ISSUE #396: Offset resets while Kafka cluster unavailable.
Hi,
I had a question about the intended behavior of a consumer and the offset while a Kafka cluster is unavailable. 
This scenario happened when I restarted a `kafka` cluster while a consumer was still running. 

Here is the simplified  consumer code (running on pykafka 2.0.4)

```
    consumer = grid_topic.get_balanced_consumer(
        consumer_group="...",
        zookeeper_connect="..",
        auto_commit_enable=False
    )

    while True:
         message = message.consume()
         process(message)
         consumer.commit()
```
Here are the logs during when the kafka cluster was restarting.
```
Dec 17 17:06:01 <ip> docker[2033]: INFO:pykafka.cluster:Attempting to discover offset manager for consumer group 'dumont'
Dec 17 17:06:01 <ip> docker[2033]: INFO:pykafka.cluster:Found coordinator broker with id 1
Dec 17 17:06:01 <ip> docker[2033]: ERROR:pykafka.simpleconsumer:Error fetching offsets for topic grid (errors: {<class 'pykafka.exceptions.OffsetsLoadInProgress'>: [0]})
Dec 17 17:06:01 <ip> docker[2033]: ERROR:pykafka.simpleconsumer:Error fetching offsets for topic grid (errors: {<class 'pykafka.exceptions.OffsetsLoadInProgress'>: [0]})
Dec 17 17:06:02 <ip> docker[2033]: INFO:pykafka.handlers:RequestHandler worker: exiting cleanly
Dec 17 17:06:02 <ip> docker[2033]: INFO:pykafka.handlers:RequestHandler worker: exiting cleanly
Dec 17 17:06:02 <ip> docker[2033]: ERROR:pykafka.simpleconsumer:Error fetching offsets for topic grid (errors: {<class 'pykafka.exceptions.OffsetsLoadInProgress'>: [0]})
Dec 17 17:06:04 <ip> docker[2033]: ERROR:pykafka.simpleconsumer:Error fetching offsets for topic grid (errors: {<class 'pykafka.exceptions.OffsetsLoadInProgress'>: [0]})
Dec 17 17:06:07 <ip> docker[2033]: ERROR:pykafka.simpleconsumer:Error fetching offsets for topic grid (errors: {<class 'pykafka.exceptions.OffsetsLoadInProgress'>: [0]})
Dec 17 17:06:11 <ip> docker[2033]: INFO:pykafka.simpleconsumer:Starting 1 fetcher threads
Dec 17 17:06:11 <ip> docker[2033]: INFO:pykafka.balancedconsumer:Rebalancing Complete.
Dec 17 17:06:11 <ip> docker[2033]: INFO:pykafka.simpleconsumer:Resetting offsets in response to OffsetOutOfRangeError
Dec 17 17:06:11 <ip> docker[2033]: INFO:pykafka.simpleconsumer:Resetting offsets for 1 partitions
Dec 17 17:06:11 <ip> docker[2033]: INFO:pykafka.simpleconsumer:Flushed queue for partition 0
```

This is my current understanding of what is happening behind the scenes, please point out if I've made a mistake! 

This behavior from the logs matches the [code](https://github.com/Parsely/pykafka/blob/e3fb2e95ff29d9bfcb080991ad6dcbf140a71ac9/pykafka/simpleconsumer.py#L494).

It retries to fetch the offset ([5 times, backing off 1000ms * try number each time](https://github.com/Parsely/pykafka/blob/e3fb2e95ff29d9bfcb080991ad6dcbf140a71ac9/pykafka/simpleconsumer.py#L64)) before continuing. 

**After it continues, it assumes the offset cannot be found, and therefore resets it according to the `auto_offset_reset`.**

This behavior doesn't seem ideal (at least in our application). It seems to be able to detect the `OffsetsLoadInProgress` error, would it make sense to raise an Exception if that is the last  error encountered on the retry as the offset should be available afterward? 

It appears the retry does already do something with [`NotCoordinatorForConsumer` and `OffsetsLoadInProgress`]((https://github.com/Parsely/pykafka/blob/e3fb2e95ff29d9bfcb080991ad6dcbf140a71ac9/pykafka/simpleconsumer.py#L518)) exceptions, but I'm not sure what that is.

Happy to provide more details if it's helpful!

emmett9001:
This pull request fixes #396 by raising an exception when `GroupLoadInProgress` is returned from an offset fetch after the maximum number of retries. I can't think of any cases in which it's actually beneficial for the consumer not to find its previously stored offsets if `reset_offsets_on_start=False`, which means I'm probably forgetting something. This change is up for discussion.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -530,6 +530,9 @@ class SimpleConsumer(object):
             to_retry.extend(parts_by_error.get(NotCoordinatorForConsumer.ERROR_CODE, []))
             reqs = [p.build_offset_fetch_request() for p, _ in to_retry]
 
+        if parts_by_error.get(GroupLoadInProgress.ERROR_CODE, []):
+            raise GroupLoadInProgress()
+
     def reset_offsets(self, partition_offsets=None):
         """Reset offsets for the specified partitions
 
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -524,14 +524,11 @@ class SimpleConsumer(object):
 
             self._cluster.handler.sleep(i * (self._offsets_channel_backoff_ms / 1000))
 
-            # retry only specific error responses
-            to_retry = []
-            to_retry.extend(parts_by_error.get(GroupLoadInProgress.ERROR_CODE, []))
-            to_retry.extend(parts_by_error.get(NotCoordinatorForConsumer.ERROR_CODE, []))
+            to_retry = [pair for err in itervalues(parts_by_error) for pair in err]
             reqs = [p.build_offset_fetch_request() for p, _ in to_retry]
 
-        if parts_by_error.get(GroupLoadInProgress.ERROR_CODE, []):
-            raise GroupLoadInProgress()
+        if len(parts_by_error) > 1:
+            raise KafkaException(parts_by_error)
 
     def reset_offsets(self, partition_offsets=None):
         """Reset offsets for the specified partitions


emmett9001:
@cameronmaske @yungchin @rduplain Do you have any thoughts on this ticket as it relates to #396?

yungchin:
I think it makes total sense that `fetch_offsets` should error out whenever any partition offsets could not be retrieved/reset. So from that point of view, I'd say we wouldn't even need to filter for `GroupLoadInProgress` here - we could just raise whenever there are any errored partitions left, regardless the error code?

emmett9001:
@yungchin I agree. It also turns out that the two exceptions being filtered for here are the only two that this request type can legally return in `0.8.2`. I committed a simplification that raises a generic exception when any error is left after retries.

Issue #457, opened 2016-02-08
Produced messages getting lost when broker is down


jonsource:
Hello, I have the following issue with pykafka, in this setting:

Running 3 kafka brokers and 1 zookeeper in docker on one machine in docker
zookeeper 172.17.0.2:2181
kafka[1-3] 172.17.0.[3-5]:9092
all default configs except kafka `config/server.properties` has 
`leader.imbalance.check.interval.seconds = 30` added, otherwise the topic leader rebalancing interval is too long - 5 minutes.
_(actually I noticed first in a real high availability cluster with 3 kafka brokers and 3 zookeepers, but that's not important I think - I reduced the problem into a local setting)_
using pykafka latest commit 0891c81 or 2.2.0, python 2.7.6, kafka 0.9.0.0, zookeeper 3.4.6, docker 1.9.1, running on Ubuntu 14.04

I send a few messages to be produced, then stop one of the kafka containers (simulating a failure) and send a few more - _(source added in gist)_. Messages for the broker that has stopped get lost.

The strange thing is pykafka reconnects all-right after a few time-outs, starts the proper worker threads, but doesn't send the messages. But during the time-out pykafka attempts to resend the messages, it seems it forgets them when the reconnect occurs _(logfile in the gist)_.
I reduced the problem to a gist here: https://gist.github.com/jonsource/4f5ec5b345f75f21b16b
with a log of a typical run, and testPykafka.py source at the bottom.

One more strange thing I noticed - the messages get lost when pausing kafka1 or kafka2. When kafka3 is paused, they get sent OK after the reconnect.

Best regards.

emmett9001:
Thanks @jonsource. I'll have to dig into this one, but at a glance it does look like a real bug.

cscheffler:
I have found similar behaviour using pykafka 2.2.0, but even when there is not a full Kafka broker failure. (Not tested using the latest master branch, so not sure whether this has been resolved there. I'll try to confirm today.)

Particularly, using

 * a 3-broker Kafka cluster,
 * a topic with 120 partitions and replication factor 2.

Start a pykafka producer that sends a message containing a counter (i.e. first message is "0", second is "1", etc.) at a rate of about 100 per second.

While the producer is running:

 * Stop kafka broker 1 – leader changes on 40 partitions, pykafka notices and after a while starts producing to the new leader.
 * Start kafka broker 1 – nothing changes
[...]

PR #459, opened 2016-02-11
Added message [not]delivered debug logs

jonsource:
Regarding the somewhat strange behaviour when a kafka broker disconnects as described here #457 I added message delivered and not delivered debug messages


pykafka/producer.py
```diff
@@ -387,9 +387,13 @@ def mark_as_delivered(message_batch):
                 for msg in mset.messages:
                     if (non_recoverable or msg.produce_attempt >= self._max_retries):
                         self._delivery_reports.put(msg, exc)
+                        log.debug("Message not delivered!! %r" % exc)
                     else:
                         msg.produce_attempt += 1
                         self._produce(msg)
+        else:
```
emmett9001: It would probably make more sense to count the number of responses without errors (at line 351) and log that number after the loop. The current way you're doing it will only log if there were no errored messages.

PR #460, opened 2016-02-11
[WIP] Compacted topic 'Skipping enqueue for offset' error

jofusa:
This is an incomplete PR. but i wanted to get a conversation started around this. 

I am reading from a compacted topic. Kafka compaction guarantees that the offset are increasing but not strictly incrementing. Two issues:

1. kafka reports earliest available offset as 0. So the next_offset is set to 0. the first message is a some offset much higher and all messages are discarded by `if message.offset != self.next_offset:`

Here is the output from the admin tools:

```
bin/kafka-run-class.sh kafka.tools.GetOffsetShell --broker-list
kafka01:9092 --topic compactedtopic --time -2
compactedtopic:0:0
compactedtopic:1:0
```
and actually pulling from that topic/partition (using kafkacat):
```
kafkacat -b kafka01 -C -t compactedtopic -p 0 -f 'Topic[%p], offset: %o,
key: %k\n' -c 3
Topic[0], offset: 151866, key: key1
Topic[0], offset: 151867, key: key2
Topic[0], offset: 151869, key: key3
```
When i pull from this topic using pykafka i get:
```
DEBUG:pykafka.simpleconsumer:Skipping enqueue for offset (151866) not equal to next_offset (0)
DEBUG:pykafka.simpleconsumer:Skipping enqueue for offset (151867) not equal to next_offset (0)
DEBUG:pykafka.simpleconsumer:Skipping enqueue for offset (151869) not equal to next_offset (0)
```
What complicates this even more then just the starting offset. A compacted log will not increment offsets. Note the jump from 151867 -> 151869.

This PR is a hack to deal with starting offset problem, however the non incrementing nature of compacted logs will require much more significant changes to the enque_messages method.

Any thoughts on how to tackle this?

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -213,7 +213,7 @@ class BalancedConsumer(object):
         self._running = False
         self._worker_exception = None
         self._worker_trace_logged = False
-        self._compacted_topic = compacted_topic
+        self._is_compacted_topic = compacted_topic
 
         if not rdkafka and use_rdkafka:
             raise ImportError("use_rdkafka requires rdkafka to be installed")
@@ -378,7 +378,7 @@ class BalancedConsumer(object):
             auto_offset_reset=self._auto_offset_reset,
             reset_offset_on_start=reset_offset_on_start,
             auto_start=start,
-            compacted_topic=self._compacted_topic
+            compacted_topic=self._is_compacted_topic
         )
 
     def _decide_partitions(self, participants):
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -90,7 +90,8 @@ class BalancedConsumer(object):
                  auto_start=True,
                  reset_offset_on_start=False,
                  post_rebalance_callback=None,
-                 use_rdkafka=False):
+                 use_rdkafka=False,
+                 compacted_topic=False):
         """Create a BalancedConsumer instance
 
         :param topic: The topic this consumer should consume
@@ -181,6 +182,10 @@ class BalancedConsumer(object):
         :type post_rebalance_callback: function
         :param use_rdkafka: Use librdkafka-backed consumer if available
         :type use_rdkafka: bool
+        :param compacted_topic: Set to read from a compacted topic. Forces
+            consumer to use less stringent ordering logic when because compacted
+            topics do not provide offsets in stict incrementing order.
+        :type compacted_topic: bool
         """
         self._cluster = cluster
         if not isinstance(consumer_group, bytes):
@@ -208,6 +213,7 @@ class BalancedConsumer(object):
         self._running = False
         self._worker_exception = None
         self._worker_trace_logged = False
+        self._compacted_topic = compacted_topic
 
         if not rdkafka and use_rdkafka:
             raise ImportError("use_rdkafka requires rdkafka to be installed")
@@ -371,7 +377,8 @@ class BalancedConsumer(object):
             offsets_commit_max_retries=self._offsets_commit_max_retries,
             auto_offset_reset=self._auto_offset_reset,
             reset_offset_on_start=reset_offset_on_start,
-            auto_start=start
+            auto_start=start,
+            compacted_topic=self._compacted_topic
         )
 
     def _decide_partitions(self, participants):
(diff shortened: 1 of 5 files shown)


jofusa:
I have changed the next_offset rejection logic to only enforce that the current message offset is larger then next_offset. 

I can't wrap my head around the horrible race conditions this opens us up to. 

Another option is to add an arg to the consumer specify if the topic is compacted and only then do we use the weaker ordinal check.

emmett9001:
@jofusa Thanks so much for this pull request, this is a great way to start the conversation around this issue.

Funny enough, you've written something very close to the [old logic](https://github.com/Parsely/pykafka/commit/4dde5d4cfe5b35cbb6c98aa47b3ba8a4fe62e36a) for this function prior to [another fix](https://github.com/Parsely/pykafka/pull/397) for exactly the type of race condition you're concerned about. It's clear that we weren't taking compacted topics into account when we wrote this. The idea of relaxing the condition slightly for compacted topics does seem like a good way to go, though of course it would be better to fix whatever race condition exists rather than band-aid it with this check. That said, I do prefer the band-aid given how difficult the race conditions can be to discover without detailed error reports.

pykafka/simpleconsumer.py
```diff
@@ -828,7 +829,18 @@ def enqueue_messages(self, messages):
...
-            if message.offset != self.next_offset:
+            # HAck to deal with compacted topic not returning earliest offset
+            # correctly
+            if not self._consumed_first_msg and message.offset != self.next_offset:
+                log.info("Detected compacted topic. "
+                          "Setting earliest offest to (%s) "
+                          "not equal to next_offset (%s)",
+                          message.offset, self.next_offset)
+                self.next_offset = message.offset
+                self._consumed_first_msg = True
```
emmett9001: Can you help me understand why this variable is necessary? Since compacted topics can have sparse offsets, won't it always be necessary to set `next_offset` here?

jofusa: I originally put the `_consumed_first_msg` in there because kafka was always reporting 0 as the earliest offset. (and i didn't notice the non-incrementing nature of the offsets at first).

I guess as long as the consumer will accept a message that has a larger offset this is not needed.

emmett9001: Thanks, I agree. @yungchin and @rduplain should chime in, but I think that a keyword argument on `SimpleConsumer` indicating whether the topic is compacted and being used to switch between `!=` and `<` is a fine stopgap to merge.

tests/pykafka/test_simpleconsumer.py
```diff
@@ -228,16 +228,28 @@ def test_partition_saves_offset(self):
         self.assertNotEqual(ret_message, None)
         self.assertEqual(ret_message.value, msgval)
 
-    def test_partition_rejects_old_message(self):
+    def test_partition_rejects_old_message_after_initial(self):
```
emmett9001: If we're going with a kwarg that indicates that a topic is compacted, I think a test should be added here, not altered.

jofusa:
I don't think this flag needs to be added to rdkafka. Take a look at the new param. I'll add an integration test at the consumer level once you like the shape this is taking.

pykafka/balancedconsumer.py
```diff
@@ -208,6 +213,7 @@ def __init__(self,
         self._running = False
         self._worker_exception = None
         self._worker_trace_logged = False
+        self._compacted_topic = compacted_topic
```
emmett9001: Boolean attribute should be named as a condition - ie `self._is_compacted_topic`

pykafka/simpleconsumer.py
```diff
@@ -828,11 +841,18 @@ def enqueue_messages(self, messages):
...
         """
         for message in messages:
-            if message.offset != self.next_offset:
+            # enforce ordering of messages
+            if self._compacted_topic and message.offset < self.next_offset:
                 log.debug("Skipping enqueue for offset (%s) "
                           "not equal to next_offset (%s)",
                           message.offset, self.next_offset)
                 continue
+            elif not self._compacted_topic and message.offset != self.next_offset:
```
emmett9001: I'd lean toward combining these two conditions since they have the same body.

jofusa:
can i get another look at this. will rebase after we are good.

emmett9001:
Sure @jofusa. I can take a look at it early next week.

emmett9001:
This change will go in with a minor version update since it adds a new kwarg.

PR #462, opened 2016-02-12
added debug notification when messages not delivered and when cluster…

jonsource:
Regarding my problems here https://github.com/Parsely/pykafka/issues/457 I cleaned the delivery debug messages as you suggested and added a further one when updating the cluster.

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -328,6 +328,7 @@ class Producer(object):
             required_acks=self._required_acks,
             timeout=self._ack_timeout_ms
         )
+        req.delivered = 0
         for msg in message_batch:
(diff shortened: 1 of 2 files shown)


pykafka/producer.py
```diff
@@ -349,6 +350,7 @@ def mark_as_delivered(message_batch):
                 for partition, presponse in iteritems(partitions):
                     if presponse.err == 0:
                         mark_as_delivered(req.msets[topic][partition].messages)
+                        delivered += len(req.msets[topic][partition].messages)
```
emmett9001: I think this counter increment should go inside `mark_as_delivered`, to keep the logic dependent on `req.msets[topic][partition].messages` in one place.

Issue #352, opened 2015-11-13
Producer finaliser may fail to flush queue, hitting ReferenceError


yungchin:
I just ran into a bug on the rdkafka branch, where queue-flushing fails due to the use of weakref (as @emmett9001 had warned about). We don't have test coverage for it, and I only noticed it by coincidence, spotting an unimportant-looking debugging log message.

In theory at least, the same bug should exist in `pykafka.Producer`, as of merging #329. This is not a regression, because before #329 we would never run the finaliser, so we would also never wait to flush the queue at program exit, but it needs fixing of course.

The `SimpleConsumer` and `BalancedConsumer` (which were also patched in #329) do not suffer from this problem: the only state they carry that needs flushing are the offsets, and we explicitly commit offsets on the thread that calls `stop()`, so we really don't care for weakref exceptions on the worker threads.

yungchin:
@rogaha I'm tagging you here just to notify that there's probably still an issue with producer queue flushing.

yungchin:
After adding a `del prod` to `test_async_produce()` (for reasons unrelated to this issue), here's a test run that hangs: https://travis-ci.org/Parsely/pykafka/jobs/90941409 (presumably in `prod._wait_all()` although we can't see that). We need a more explicit test case for this of course, but I think it confirms the suspicion that there's still a bug here.

rogaha:
Thanks for spotting this @yungchin! Do you have an idea of how to approach it? We ended up using the upstream version ``2.0.3`` to avoid maintaining a fork and adding this right after the ``Producer`` instantiation: ``atexit.register(lambda p: p.stop() if p._running else None, producer)``.

yungchin:
@rogaha if you're using `atexit`, you won't be affected by this particular problem, because when the atexit handlers run the producer object is still reachable and so there are no invalidated weakrefs yet. That is, if you call `stop()` via `atexit` it will successfully see the queue to flush through.

About your code snippet: `p._running` is not really part of the public interface, so it might at some point break for you. Are there issues if you just call `p.stop()`? It should be safe to call that unqualified, otherwise I'd say we have a bug in `stop()`. Thanks!

yungchin:
See also #422 which confirms this isn't just theoretical.

matthewatabet:
Do we know when this might be resolved?

[...]

PR #463, opened 2016-02-12
gracefuly stop producer via joining OwnedBroker.queue_reader threads

jonsource:
I've found a solution to my problem (https://github.com/Parsely/pykafka/issues/457) with the producer exiting early on producer.stop() and forgetting to send messages that originally targeted a kafka broker which went down.

Simply by naming the threads, filtering them and joining them in the producer.stop() method I was able to achieve a graceful shut-down.

While at it I also added names to all the threads I've found.

### DIFF
--- a/pykafka/handlers.py
+++ b/pykafka/handlers.py
@@ -81,7 +81,7 @@ class ThreadingHandler(Handler):
     Event = threading.Event
     Lock = threading.Lock
     Semaphore = Semaphore
-    _maxCount = 0
+    _workers_spawned = 0
 
     def sleep(self, seconds=0):
         time.sleep(seconds)
@@ -97,11 +97,11 @@ class ThreadingHandler(Handler):
 
     def spawn(self, target, *args, **kwargs):
         if 'name' in kwargs:
-            kwargs['name'] = "{}: {}".format(ThreadingHandler._maxCount, kwargs['name'])
+            kwargs['name'] = "{}: {}".format(ThreadingHandler._workers_spawned, kwargs['name'])
         t = threading.Thread(target=target, *args, **kwargs)
         t.daemon = True
         t.start()
-        ThreadingHandler._maxCount += 1
+        ThreadingHandler._workers_spawned += 1
         return t
 
 
--- a/pykafka/handlers.py
+++ b/pykafka/handlers.py
@@ -81,6 +81,7 @@ class ThreadingHandler(Handler):
     Event = threading.Event
     Lock = threading.Lock
     Semaphore = Semaphore
+    _maxCount = 0
 
     def sleep(self, seconds=0):
         time.sleep(seconds)
@@ -95,9 +96,12 @@ class ThreadingHandler(Handler):
             return threading.RLock(*args[1:], **kwargs)
 
     def spawn(self, target, *args, **kwargs):
+        if 'name' in kwargs:
+            kwargs['name'] = "{}: {}".format(ThreadingHandler._maxCount, kwargs['name'])
         t = threading.Thread(target=target, *args, **kwargs)
         t.daemon = True
         t.start()
+        ThreadingHandler._maxCount += 1
         return t
 
 
@@ -113,6 +117,9 @@ class GEventHandler(Handler):
         gevent.sleep(seconds)
 
     def spawn(self, target, *args, **kwargs):
+        # Greenlets don't support naming
+        if 'name' in kwargs:
+            kwargs.pop('name')
         t = gevent.spawn(target, *args, **kwargs)
         return t
 
--- a/pykafka/handlers.py
+++ b/pykafka/handlers.py
@@ -193,4 +193,6 @@ class RequestHandler(object):
                     shared.requests.task_done()
             log.info("RequestHandler worker: exiting cleanly")
 
-        return self.handler.spawn(worker)
+        return self.handler.spawn(worker, name="pykafka.RequestHandler.worker for {}:{}".format(
+                                            self.shared.connection.host,
+                                            self.shared.connection.port))
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -237,12 +237,7 @@ class Producer(object):
             for owned_broker in self._owned_brokers.values():
                 owned_broker.stop()
 
-    def get_queue_readers(self):
-        if not self._owned_brokers:
-            return []
-        return [owned_broker._queue_reader_worker for owned_broker in self._owned_brokers.values() if owned_broker.running]
-
-    def stop(self, wait=True):
+    def stop(self):
         """Mark the producer as stopped, and wait until all messages to be sent
 
         The join here works because new queue readers are spawned during the execution of the old ones i.e:
@@ -250,12 +245,14 @@ class Producer(object):
         NotLeaderForPartition.ERROR_CODE it updates the cluster, sets up owned brokers thus starting new queue reader
         threads. Because of this when we call self.get_queue_readers, we get the newly created queue readers, and so try
         to stop and join them etc ... until they all stop without encountering problems in producer._send_request
-
-        :param wait: whether we should wait until all messages to be sent or not
-        :type wait: bool
         """
-        while wait and self._running:
-            queue_readers = self.get_queue_readers()
+        def get_queue_readers():
+            if not self._owned_brokers:
+                return []
+            return [owned_broker._queue_reader_worker for owned_broker in self._owned_brokers.values() if owned_broker.running]
+
+        while self._running:
+            queue_readers = get_queue_readers()
             self._stop_owned_brokers()
             if len(queue_readers) == 0:
                 self._running = False
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -237,25 +237,29 @@ class Producer(object):
             for owned_broker in self._owned_brokers.values():
                 owned_broker.stop()
 
+    def get_queue_readers(self):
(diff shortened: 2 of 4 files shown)


pykafka/producer.py
```diff
@@ -231,14 +231,32 @@ def _setup_owned_brokers(self):
...
+    def stop(self, wait=True):
+        """Mark the producer as stopped, and wait until all messages to be sent
+
+        :param wait: whether we should wait until all messages to be sent or not
+        :type wait: bool
+        """
+        def get_queue_readers():
+            return [qr for qr in threading.enumerate() if qr.name.find('pykafka.OwnedBroker.queue_reader') == 0]
+
+        while wait and self._running:
```
emmett9001: I don't understand this logic - where is `self._running` set to `False` in this loop?

emmett9001:
Thank you for this contribution. I'm having a hard time figuring out exactly what problem it's solving. Each queue reader thread's `running` flag is used as an external interface to stop that thread, so I'm not clear on the need to call `join()`. Also, message retries happen by re-enqueueing the failed messages, so this logic seems to still allow messages to be dropped if they're re-enqueued after the queue reader is stopped.

pykafka/producer.py
jonsource: The loop is quit by `break` on line 256 once all queue_reader threads are joined. The self._running property indicates the producer is stopped and not waiting for any threads to be joined.

jonsource:
Thank you for your time. The problem I'm trying to solve with the joins is this:

In my tests with the master branch, when `producer.stop()` is called - the queue reader threads `running` flags are set to `False`, and because the threads are daemonic (set in ThreadingHandler.spawn) the execution doesn't wait for them to finish - as soon as the MainThread completes, the program stops.

With my graceful_stop when `producer.stop()` is called - the queue reader threads are flagged to be stopped, and by `join` we wait for their execution, then try checking for existing queue readers once more.
After a queue reader in its call to `producer._send_request` encounters either a `SocketDisconnectedError` or `NotLeaderForPartition.ERROR_CODE` it updates the cluster, sets up owned brokers thus starting new queue reader threads. Because of this when we filter existing queue reader threads on line 215, we see the newly created queue readers, and so try to stop them and join them etc ... until they all stop without encountering problems in `producer._send_request`

So, it seems to me, the main thread will wait for the last attempt at resend to finish, and exit afterwards.

If I could help by, for example, providing a docker suite to illustrate the problem, I'll gladly do it. This is really important to me.

pykafka/producer.py
```diff
@@ -231,14 +231,32 @@ def _setup_owned_brokers(self):
...
                 owned_broker.stop()
 
+    def stop(self, wait=True):
+        """Mark the producer as stopped, and wait until all messages to be sent
+
+        :param wait: whether we should wait until all messages to be sent or not
+        :type wait: bool
+        """
+        def get_queue_readers():
+            return [qr for qr in threading.enumerate() if qr.name.find('pykafka.OwnedBroker.queue_reader') == 0]
```
emmett9001: This logic will break when `use_greenlets=True`. Two possible solutions occur to me. You could save references to the objects returned by `_cluster.handler.spawn` and pass them into a `self._cluster.handler.join_worker(t)` function, implementing both a gevent and a threaded version of that function. You could also write a `self._cluster.handler.get_workers_by_substring()` and call it here. The threaded version would use `threading.enumerate()`, and the greenlet version would use the gevent equivalent.

Also, I think that the identifying string for these worker threads should be deduplicated by putting it into a class-level variable.

pykafka/producer.py
```diff
@@ -231,14 +231,32 @@ def _setup_owned_brokers(self):
...
+
+        while wait and self._running:
+            self._stop_owned_brokers()
+            readers = get_queue_readers()
+            if len(readers):
```
emmett9001: I think this is clearer as
```python
if len(readers) == 0:
    break
for reader in readers:
    reader.join()
```

pykafka/producer.py
```diff
@@ -231,14 +231,32 @@ def _setup_owned_brokers(self):
...
         if self._owned_brokers is not None:
             for owned_broker in self._owned_brokers.values():
                 owned_broker.stop()
 
+    def stop(self, wait=True):
+        """Mark the producer as stopped, and wait until all messages to be sent
+
+        :param wait: whether we should wait until all messages to be sent or not
+        :type wait: bool
+        """
```
emmett9001: This docstring would benefit from a detailed explanation of why the looped `join()` calls work here, similar to your comment on this pull request.

pykafka/producer.py
```diff
@@ -448,7 +466,8 @@ def queue_reader():
             log.info("Worker exited for broker %s:%s", self.broker.host,
                      self.broker.port)
         log.info("Starting new produce worker for broker %s", broker.id)
-        self.producer._cluster.handler.spawn(queue_reader)
+        self.producer._cluster.handler.spawn(queue_reader,
+                                             name="pykafka.OwnedBroker.queue_reader for broker %d" % self.broker.id)
```
emmett9001: Does `name` work in gevent? That might have to be wrapped as well.

emmett9001:
@jonsource Thank you for the very clear explanation of what's going on here. It helps me understand why this is useful, and I was actually able to cross-reference it with another longstanding issue (#352). I'm not certain they're the same issue, but this at least sheds light on the situation.

I've left a bunch of comments on your changeset, many related to how this will work with gevent. If you'd prefer not to touch the complexity added by that aspect of the project, maybe you can reopen this pull request against a branch other than master? (I'm not sure if github allows that). If so, I can tackle the gevent-related complexity after merging into a branch.

pykafka/producer.py
```diff
@@ -231,14 +231,38 @@ def _setup_owned_brokers(self):
...
-    def stop(self):
-        """Mark the producer as stopped"""
-        self._running = False
+    def _stop_owned_brokers(self):
         self._wait_all()
         if self._owned_brokers is not None:
             for owned_broker in self._owned_brokers.values():
                 owned_broker.stop()
 
+    def get_queue_readers(self):
```
emmett9001: I'm actually in favor of keeping this function inside `stop()` for now since it's only used there.

pykafka/handlers.py
```diff
@@ -81,6 +81,7 @@ class ThreadingHandler(Handler):
     Event = threading.Event
     Lock = threading.Lock
     Semaphore = Semaphore
+    _maxCount = 0
```
emmett9001: This could use a clearer name, maybe `workersSpawned`

pykafka/producer.py
```diff
@@ -231,14 +231,38 @@ def _setup_owned_brokers(self):
...
+        """
+        while wait and self._running:
+            queue_readers = self.get_queue_readers()
+            self._stop_owned_brokers()
+            if len(queue_readers) == 0:
+                self._running = False
+            else:
+                for queue_reader in queue_readers:
+                    queue_reader.join()
+
```
emmett9001: If `wait` is `False`, we'll never set `self._running = False`, which seems wrong. Maybe you could put another `self._running = False` after the end of the loop.

pykafka/producer.py
```diff
@@ -231,14 +231,38 @@ def _setup_owned_brokers(self):
...
         if self._owned_brokers is not None:
             for owned_broker in self._owned_brokers.values():
                 owned_broker.stop()
 
+    def get_queue_readers(self):
+        if not self._owned_brokers:
+            return []
+        return [owned_broker._queue_reader_worker for owned_broker in self._owned_brokers.values() if owned_broker.running]
+
+    def stop(self, wait=True):
```
emmett9001: I'm not sure this extra kwarg is necessary. The current implementation of `stop()` assumes that there is no reason we'd want to avoid waiting, and I think it's fine to keep it that way for now. Additionally, keeping the function signature the same will make this change easier to get into a release since it won't need to be rolled into a minor version update.

emmett9001:
Thanks for your continued effort on this pull request @jonsource. I've highlighted a few issues in comments, but they're pretty minor. There's also a failing test `tests/pykafka/test_producer.py::ProducerIntegrationTests::test_async_produce_lingers`, which is failing because this change can cause multiple calls to `_wait_all`. I think it's appropriate to change that test to check that the time waited is *at least* `linger_ms` instead of exactly that value, since that's effectively what this change does.

Unless you have a strong opinion about the extra kwarg on `stop()`, I think this is ready to merge. If you don't feel like fixing the small comments I've made, I can just address them post-merge. Thanks again for your help!

pykafka/producer.py
emmett9001: Alternately, removing the `wait` kwarg from this function solves this issue pretty nicely.

jonsource:
I'm not sure why the checks failed, but it seems to me more like a random build error than enything else. Also I encounter difficulties while running tox locally, some tests fail all the time (tests/pykafka/test_balancedconsumer.py::BalancedConsumerIntegrationTests::test_zk_conn_lost, tests/pykafka/test_balancedconsumer.py::BalancedConsumerGEventIntegrationTests::test_zk_conn_lost - but are OK on TravisCI) some tests sometimes go well, sometimes fail.
Is there some specification of the ZK-Kafka cluster setting that should be provided for the tests to work OK ?

emmett9001:
@jonsource I hate to admit this, but the tests have been so inconsistent for me locally that I've basically given up running the entire suite locally. I usually make a change on a branch and then run the tests on Travis. Ephemeral test failures are a big problem for us, and something we'll eventually need to fix before it gets worse. The continued bugfixes are helpful toward that goal.

PR #477, opened 2016-02-29
ssl support

yungchin:
This leverages the standard library's `ssl` module to wrap our `BrokerConnection` sockets, enabling TLS connections to Kafka. It's a work-in-progress:

 - [x] auto-generate certificates for use in test suite
 - [x] enable `KafkaInstance` SSL ports for testing (where possible)
 - [x] build successfully with librdkafka 0.9.x (see #445)
 - [x] insert necessary config for rdkafka-backed classes
 - [x] wrap `BrokerConnection` sockets
 - [x] decide a non-ugly `KafkaClient` init-signature that can support both the non-rdkafka and rdkafka-backed classes (this is the most difficult step in all)
 - [x] add connection examples to documentation
 - [x] add (suitably scoped) tests


emmett9001:
I'm not well-versed in the history of Python's SSL support, but I do think we need to continue supporting all versions of python 2.7. If there's a choice that can allow this changeset to retain all 2.7 support, I'm in favor of making the interface slightly uglier to that end.

yungchin:
@emmett9001 @rduplain if you could have another look at this that'd be great. I've implemented the config object in the end, which also has a fallback for older 2.7-series interpreters. If anyone has very specific needs, the `SslConfig` docstring also explains how to supplant it (through duck-typing).

One thing I haven't had a chance to think about yet, is whether this config interface is flexible enough to also support future SASL functionality (alternatively, that would just have `KafkaClient` gain yet another init parameter when the time comes...).

yungchin:
PS: all the test failures on the recent commits on this branch are, I believe, incidental and unrelated to work on this branch. I'll try and get out a branch with fixes for one or two tests that seem to keep randomly tripping up.

pykafka/connection.py
```diff
@@ -18,7 +18,9 @@
 limitations under the License.
 """
 __all__ = ["BrokerConnection"]
+from functools import partial
 import logging
+import ssl
```
emmett9001: Mind keeping imports alpha-sorted?

pykafka/cluster.py
```diff
@@ -234,10 +238,14 @@ def _request_metadata(self, broker_connects, topics):
...
                                     source_host=self._source_host,
-                                    source_port=self._source_port)
+                                    source_port=self._source_port,
+                                    ssl_config=self._ssl_config)
                     response = broker.request_metadata(topics)
                     if response is not None:
                         return response
+                # FIXME: handle ssl errors separately somehow, because if we
+                # don't, _get_metadata() might attempt to treat it as if it
+                # were a zookeeper host, which gets messy
```
emmett9001: Curious - what happens on the fallback that gets so messy? If it was bad enough it would be a case for permanently [removing](https://github.com/Parsely/pykafka/pull/452) that fallback (which, by the way, I'm not actually certain warrants a major version bump, all things considered).

pykafka/connection.py
```diff
@@ -41,12 +116,14 @@ def __init__(self,
...
-                 source_port=0):
+                 source_port=0,
+                 ssl_config=None):
         """Initialize a socket connection to Kafka.
 
         :param host: The host to which to connect
         :type host: str
-        :param port: The port on the host to which to connect
+        :param port: The port on the host to which to connect.  Assumed to be
+            an ssl-endpoint if (and only if) `ssl_config` is also provided
```
emmett9001: This point might bear repeating in the `KafkaClient` docstring

pykafka/connection.py
```diff
@@ -29,6 +31,79 @@
...
+        config = SslConfig(cafile='/your/ca/file')
+        config._wrap_socket = config._legacy_wrap_socket()
+        client = KafkaClient('localhost:<ssl-port>', ssl_config=config)
+
+    Alternatively, completely supplanting this class with your own is also
+    simple: if you are not going to be using the `pykafka.rdkafka` classes,
+    only a method `wrap_socket()` is expected (so you can eg. simply pass in
+    a plain `ssl.SSLContext` instance instead).  The `pykafka.rdkafka`
+    classes require four further attributes: `cafile`, `certfile`, `keyfile`,
+    and `password` (for details, see init docstring)
```
emmett9001: I don't see the init docstring this references anywhere

pykafka/rdkafka/helpers.py
```diff
@@ -0,0 +1,31 @@
...
+            "ssl.cipher.suites": ciphers,
+            "ssl.ca.location": conf.cafile,
+            "ssl.certificate.location": conf.certfile,
+            "ssl.key.location": conf.keyfile,
+            "ssl.key.password": conf.password}
+        # Some of the values above may not have been configured:
+        return {k: v for k, v in rdk_conf.items() if v is not None}
+
+
+def get_ciphers_from_sockets(brokers):
```
emmett9001: We should make this function private, since it's not getting used outside of this file.

tests/pykafka/test_ssl.py
```diff
@@ -0,0 +1,64 @@
...
+    def setUpClass(cls):
+        cls.kafka = get_cluster()
+        if cls.kafka.brokers_ssl is None:
+            pytest.skip("Test-cluster doesn't advertise ssl ports.")
+
+    @classmethod
+    def tearDownClass(cls):
+        stop_cluster(cls.kafka)
+
+    def roundtrip_test(self, client):
```
emmett9001: Why does this use a naming convention that keeps it from being automatically run in travis?

pykafka/test/kafka_instance.py
```diff
@@ -336,6 +378,133 @@ def produce_messages(self, topic_name, messages):
...
+               '-subj', '/CN=ca.example.com',
+               '-days', self.validity_days,
+               '-out', self.root_cert,
+               '-keyout', self.root_key,
+               '-passout', 'env:ROOT_PASS']
+        return subprocess.check_call(cmd, env=dict(ROOT_PASS=self.root_pass))
+
+    def _gen_broker_truststore(self):
+        """Generate truststore that trusts self.root_cert"""
+        cmd = ['keytool', '-importcert', '-alias', 'root',
```
emmett9001: I'm not very well-versed in the details of SSL, so just making sure I understand here. These functions are necessary because the SSL dance requires the host machine to contain some specific files in its registry of trusted SSL entities?

emmett9001:
Looks awesome, thanks @yungchin. I've left a bunch of comments, but it's nothing major. I love that you can use anything that implements `wrap_socket` as an SSL config - that's a smart design choice. Also, it's clear that the `CertManager` is the most arcane aspect of this diff, so thanks very much for figuring that out. Once the comments I've made above are addressed, I'm excited to merge this.

Note to self, this change will require a minor version bump.

pykafka/connection.py
yungchin: Hahaha, uhm, I suppose if I were to confess that I thought I'd done that, that'd be kind of embarassing, right? So, no confessions :)

pykafka/connection.py
yungchin: I've got something like this in the `hosts` description there.

pykafka/test/kafka_instance.py
yungchin: Yes, that's right: this one tells the broker to trust the root cert that we've used to sign the client's cert with.in L497 below. That is, it will accept connections from any client as long as the client cert was signed by the trusted cert.

pykafka/cluster.py
yungchin: You can see it when you connect to an SSL port while not providing an `ssl_config` to `KafkaClient`: the connection keeps getting closed (because the broker is expecting some sort of SSL handshake and not getting it), and kazoo stubbornly keeps trying to reconnect - it doesn't figure out that the other end is simply not a Zookeeper machine.

yungchin:
Thanks for picking this up so promptly @emmett9001 - I think I've addressed everything except the fixme (which I'd forgotten about, thanks!). I'm not super-clear how to deal with that one quite yet. The connection errors aren't easy to identify as ssl-related...

pykafka/cluster.py
emmett9001: That makes me strongly want to remove the zookeeper fallback. I don't think we actually need to increment the major version - what do you think?

yungchin: I was thinking about this a bit, and strictly speaking, it is not a regression caused by this pull request - in the sense that if you use current master and try to connect to an ssl port (which obviously won't work) you'd already get stuck in the same way. Also, it's mostly just a nuisance, in that if the user gets the config wrong, the feedback is unhelpful.

So how about I remove these "fixme" lines here, but cross-reference them on #452 as an additional motivation for it, and we could still keep #452 on hold for the next version bump?

emmett9001: Sounds fine to me @yungchin, though I'll be looking for opportunities to get #452 in before 3.0

yungchin:
Ugh. I added another two commits to sort out the test configuration for Travis, and was hoping that was that, but the last test run surfaced a librdkafka error (which I can't reproduce, so I don't dare restart that test - it will probably disappear forever): https://travis-ci.org/Parsely/pykafka/jobs/127758160#L821
```
python3.5: rd.h:275: rd_refcnt_sub0: Assertion `!*"refcnt sub-zero"' failed.
```
Will investigate.

PR #479, opened 2016-03-03
Expose BalancedConsumer.topic as a property

erickt:
This matches the api of SimpleConsumer


emmett9001:
Thanks for the contribution - this is an oversight I'm happy to correct.

PR #480, opened 2016-03-04
compressed message batch sizing

jofusa:
Just wanted some code to start a discussion.

I have been struggling to deal with some `kafka.common.MessageSizeTooLargeException` I have found in my kafka logs. My messages should be well below the max message size. I am most interested in surfacing these errors in my client. My first attempt to build a producer exception background thread failed because the delivery_report queue is thread local. Also note the sync producer is too slow to process the source topic in any reasonable amount of time.

So my first question is why is the delivery_report queue is thread local. I am sure there is a good reason, but just wondering.

Second, what are thoughts on the the approach in this pr. I think if would make sense to only trigger the callback on a the list of `non_recoverable` exceptions. Really i am using this code to try to report the failed message to help diagnose the underlying cause.

```py
        def callback(msgs, exc):
            # send full message to external store or log.
            raise exc

        producer = topic.get_producer(
                delivery_reports=True,
                producer_exception_callback=callback
        )
```

The general issue is currently there is a warning level log, but no easy way to rectify.

Third, because the producer bundles the messages into a messageset. Does kafka reject the entire messageset because it is too large even though the underlying messages are small enough?

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -66,8 +66,10 @@ class Producer(object):
                  min_queued_messages=70000,
                  linger_ms=5 * 1000,
                  block_on_queue_full=True,
+                 max_request_size=1048576,
                  sync=False,
-                 delivery_reports=False):
+                 delivery_reports=False,
+                 auto_start=True):
         """Instantiate a new AsyncProducer
 
         :param cluster: The cluster to which to connect
@@ -119,6 +121,13 @@ class Producer(object):
             indicates we should block until space is available in the queue.
             If False, we should throw an error immediately.
         :type block_on_queue_full: bool
+        :param max_request_size:
+            The maximum size of a request in bytes. This is also effectively a
+            cap on the maximum record size. Note that the server has its own
+            cap on record size which may be different from this. This setting
+            will limit the number of record batches the producer will send in a
+            single request to avoid sending huge requests.
+        :type max_request_size: int
         :param sync: Whether calls to `produce` should wait for the message to
             send before returning.  If `True`, an exception will be raised from
             `produce()` if delivery to kafka failed.
@@ -131,6 +140,10 @@ class Producer(object):
             or an `Exception` in case of failed delivery to kafka.
             This setting is ignored when `sync=True`.
         :type delivery_reports: bool
+        :param auto_start: Whether the producer should begin communicating
+            with kafka after __init__ is complete. If false, communication
+            can be started with `start()`.
+        :type auto_start: bool
         """
         self._cluster = cluster
         self._topic = topic
@@ -144,6 +157,7 @@ class Producer(object):
         self._min_queued_messages = max(1, min_queued_messages)
         self._linger_ms = linger_ms
         self._block_on_queue_full = block_on_queue_full
+        self._max_request_size = max_request_size
         self._synchronous = sync
         self._worker_exception = None
         self._worker_trace_logged = False
@@ -151,10 +165,11 @@ class Producer(object):
         self._delivery_reports = (_DeliveryReportQueue(self._cluster.handler)
                                   if delivery_reports or self._synchronous
                                   else _DeliveryReportNone())
-
+        self._auto_start = auto_start
         self._running = False
         self._update_lock = self._cluster.handler.Lock()
-        self.start()
+        if self._auto_start:
+            self.start()
 
     def __del__(self):
         log.debug("Finalising {}".format(self))
@@ -470,8 +485,12 @@ class OwnedBroker(object):
     :type messages_pending: int
     :ivar producer: The producer to which this OwnedBroker instance belongs
     :type producer: :class:`pykafka.producer.AsyncProducer`
+    :param auto_start: Whether the OwnedBroker should start flushing all
+    waiting messages and send to with kafka after __init__ is complete. If
+    false, communication can be started with `start()`.
+    :type auto_start: bool
     """
-    def __init__(self, producer, broker):
+    def __init__(self, producer, broker, auto_start=True):
         self.producer = weakref.proxy(producer)
         self.broker = broker
         self.lock = self.producer._cluster.handler.RLock()
@@ -480,11 +499,16 @@ class OwnedBroker(object):
         self.queue = deque()
         self.messages_pending = 0
         self.running = True
+        self._auto_start = auto_start
 
+        if self._auto_start:
+            self.start()
+
+    def start(self):
         def queue_reader():
             while self.running:
                 try:
-                    batch = self.flush(self.producer._linger_ms)
+                    batch = self.flush(self.producer._linger_ms, self.producer._max_request_size)
                     if batch:
                         self.producer._send_request(batch, self)
                 except Exception:
@@ -493,7 +517,7 @@ class OwnedBroker(object):
                     break
             log.info("Worker exited for broker %s:%s", self.broker.host,
                      self.broker.port)
-        log.info("Starting new produce worker for broker %s", broker.id)
+        log.info("Starting new produce worker for broker %s", self.broker.id)
         name = "pykafka.OwnedBroker.queue_reader for broker {}".format(self.broker.id)
         self._queue_reader_worker = self.producer._cluster.handler.spawn(
             queue_reader, name=name)
@@ -527,12 +551,15 @@ class OwnedBroker(object):
                 if not self.flush_ready.is_set():
                     self.flush_ready.set()
 
-    def flush(self, linger_ms, release_pending=False):
+    def flush(self, linger_ms, max_request_size=1048576, release_pending=False):
         """Pop messages from the end of the queue
 
         :param linger_ms: How long (in milliseconds) to wait for the queue
             to contain messages before flushing
         :type linger_ms: int
+        :param max_request_size: The max size should each batch of messages
+            should be in bytes
+        :type max_request_size: int
         :param release_pending: Whether to decrement the messages_pending
             counter when the queue is flushed. True means that the messages
             popped from the queue will be discarded unless re-enqueued
@@ -541,7 +568,25 @@ class OwnedBroker(object):
         """
         self._wait_for_flush_ready(linger_ms)
         with self.lock:
-            batch = [self.queue.pop() for _ in range(len(self.queue))]
+            batch = []
+            batch_size_in_bytes = 0
+            while len(self.queue) > 0:
+                # TODO: Should we optimistically pop and renenqueue vs. peak
+                peaked_message = self.queue[-1]
+
+                if len(peaked_message.value) > max_request_size:
+                    raise MessageSizeTooLarge("Message size larger then max_request_size: %d",
+                                              max_request_size)
+
+                # test if adding the message would go over the
+                # max_request_size. if it would, break out of loop
+                if batch_size_in_bytes + len(peaked_message.value) > max_request_size:
+                    break
+
+                message = self.queue.pop()
+                batch_size_in_bytes += len(message.value)
+                batch.append(message)
+
             if release_pending:
                 self.increment_messages_pending(-1 * len(batch))
             if not self.slot_available.is_set():
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -243,10 +243,9 @@ class Producer(object):
                 # broker queue
                 while True:
                     batch = owned_broker.flush(self._linger_ms, self._max_request_size)
-                    if batch:
-                        queued_messages.extend(batch)
-                    else:
+                    if not batch:
                         break
+                    queued_messages.extend(batch)
 
         self._owned_brokers = {}
         for partition in self._topic.partitions.values():
@@ -493,8 +492,8 @@ class OwnedBroker(object):
     :ivar producer: The producer to which this OwnedBroker instance belongs
     :type producer: :class:`pykafka.producer.AsyncProducer`
     :param auto_start: Whether the OwnedBroker should start flushing all
-    waiting messages and send to with kafka after __init__ is complete. If
-    false, communication can be started with `start()`.
+        waiting messages and send to kafka after __init__ is complete. If
+        false, communication can be started with `start()`.
     :type auto_start: bool
     """
     def __init__(self, producer, broker, auto_start=True):
@@ -578,11 +577,10 @@ class OwnedBroker(object):
             batch = []
             batch_size_in_bytes = 0
             while len(self.queue) > 0:
-                # TODO: Should we optimistically pop and renenqueue vs. peak
-                peaked_message = self.queue[-1]
+                peeked_message = self.queue[-1]
 
-                if peaked_message and peaked_message.value is not None:
-                    if len(peaked_message) > max_request_size:
+                if peeked_message and peeked_message.value is not None:
+                    if len(peeked_message) > max_request_size:
                         exc = MessageSizeTooLarge(
                             "Message size larger then max_request_size: {}".format(max_request_size)
                         )
@@ -590,14 +588,14 @@ class OwnedBroker(object):
                         # bind the MessageSizeTooLarge error the delivery
                         # report and remove it from the producer queue
                         message = self.queue.pop()
-                        peaked_message.delivery_report_q.put((message, exc))
+                        peeked_message.delivery_report_q.put((message, exc))
                         # remove from pending message count
                         self.increment_messages_pending(-1)
                         continue
 
                     # test if adding the message would go over the
                     # max_request_size. if it would, break out of loop
-                    elif batch_size_in_bytes + len(peaked_message) > max_request_size:
+                    elif batch_size_in_bytes + len(peeked_message) > max_request_size:
                         log.debug("max_request_size reached. producing batch")
                         # if we did not fully empty the queue. reset the
                         # flush_ready so we send another batch immediately
@@ -605,8 +603,7 @@ class OwnedBroker(object):
                         break
 
                 message = self.queue.pop()
-                if message.value is not None:
-                    batch_size_in_bytes += len(message)
+                batch_size_in_bytes += len(message)
                 batch.append(message)
 
             if release_pending:
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -66,7 +66,7 @@ class Producer(object):
                  min_queued_messages=70000,
                  linger_ms=5 * 1000,
                  block_on_queue_full=True,
-                 max_request_size=1048576,
+                 max_request_size=1000012,
                  sync=False,
                  delivery_reports=False,
                  auto_start=True):
@@ -551,7 +551,7 @@ class OwnedBroker(object):
                 if not self.flush_ready.is_set():
                     self.flush_ready.set()
 
-    def flush(self, linger_ms, max_request_size=1048576, release_pending=False):
+    def flush(self, linger_ms, max_request_size, release_pending=False):
         """Pop messages from the end of the queue
 
         :param linger_ms: How long (in milliseconds) to wait for the queue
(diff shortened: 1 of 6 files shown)


jofusa:
regrading point 3. after some testing, kafka seems to be doing the reasonable thing. even if the sum of the messages in a set are larger then the max message size, the producer works just fine.

emmett9001:
Here's @yungchin's [initial explanation](https://github.com/Parsely/pykafka/pull/344#issuecomment-157762655) of why the delivery report queues are thread local:

> The main reason I started out with a thread-local queue here is that it nicely supports a use model that the futures interface also supported well: if you have something like a multi-threaded web server, with all threads sharing a single producer, but each thread checking its own delivery failures (or each running with sync=True), then you don't want them to de-queue each others reports. On the other hand, I suppose there may be use models where you'd produce() from multiple threads but have a single thread somewhere handling all delivery failures.

It looks like this may be a [case](https://github.com/Parsely/pykafka/pull/344#issuecomment-157850921) for rethinking this design.

Have you considered creating an exception monitoring thread per producing thread and  putting the caught exceptions onto a queue?

The [kafka documentation](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ErrorCodes) implies that the total size of a batch has no bearing on whether `MessageSizeTooLarge` is returned:

> The server has a configurable maximum message size to avoid unbounded memory allocation. This error is thrown if the client attempt to produce a message larger than this maximum.

Have you tried adjusting the `max.message.bytes` field in your Kafka configuration?

jofusa:
Thanks for the response. yung's approach makes complete sense.

My threading skillz are nascent at best so i don't actually follow how you would create a background thread that would have access to the thread local report queue.

I think i have actually found the underlying issue for my issue. compression causes the issue. i will add some tests that expose the bug then we can talk about how to squash it.

jofusa:
I have added two tests that exposes the issue. when you use a compression the produce fails. I suspect that the compression payload is being interpreted by kafka as one message. need to dig into this.

tests/pykafka/test_producer.py
```diff
@@ -177,6 +179,54 @@ def test_null_payloads(self):
         prod.produce(b"")  # empty string should be distinguished from None
         self.assertEqual(b"", self.consumer.consume().value)
 
+    def test_async_produce_compression_large_message(self):
+        large_payload = b''.join([uuid4().bytes for i in range(50000)])
+        assert sys.getsizeof(large_payload) / 1024 / 1024 < 1.0
```
jofusa: default message size for kafka is 1 meg

jofusa:
just catching up on the compression protocol: 

From the docs:

Kafka supports compressing messages for additional efficiency, however this is more complex than just compressing a raw message. Because individual messages may not have sufficient redundancy to enable good compression ratios, compressed messages must be sent in special batches (although you may use a batch of one if you truly wish to compress a message on its own). The messages to be sent are wrapped (uncompressed) in a MessageSet structure, which is then compressed and stored in the Value field of a single "Message" with the appropriate compression codec set. The receiving system parses the actual MessageSet from the decompressed value. The outer MessageSet should contain only one compressed "Message" (see KAFKA-1718 for details).
Kafka currently supports two compression codecs with the following codec numbers:


makes sense you would want to batch messages to give the compression algo a chance to actually do something. But it feels like an overloaded protocol, one `Message` made up of many `messages`, meh.

My first reaction is to limit the size of messagesets sent to the a compressed message set to 1 megabyte(kafka's default). Allow users to override. We batch the messages by the sum of the uncompressed size, this wouldn't maximize the compression and bandwidth but would keep the producer code simpler. Trying to guess the size of the compressed message a priori seems nasty. 

The other approach is we could be optimistic and keep the current approach, pass in as many messages all possible and then check the compressed size, throw and error if it's too big and then pass in a smaller set. This would be better for smaller messages as long as the total compressed Message is small enough, but would create an overhead for larger messages.

A mix of these two approaches i think would be the fastest but the most complex. Basically the producer would check the size of some enqueued messages and switch between the defensive and optimistic approach based on size.

Happy to help code up a PR, but would love some discussion before we start coding.

jofusa:
also could help to check what the java producer is doing.

jofusa:
And yet another option is to allow the user to set the message_set batch size manually. Seems cumbersome and leaves a pretty sharp edge for users to poke their eye out on.

jofusa:
here is the relevant code from the new java consumer:

https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/internals/RecordAccumulator.java#L340

emmett9001:
@jofusa I think your first suggestion is reasonable - batching by the sum of the uncompressed message sizes. I also believe that it's possible to get around this issue on current master by setting both `max.message.bytes` in Kafka and `min_queued_messages` on the pykafka producer.

If you'd like to put together a pull request that batches by the sum of uncompressed message sizes, I'd say go for it. You can keep it in this pull request, but you should remove the error callback code. I think there are better ways to solve that particular problem, and it's not directly related to the main thrust of this pull request anymore.

jofusa:
I will put together a PR and remove the exception callback.

I agree that this can be controlled by `min_queued_messages` but this in non obvious to the end user. I think there is a strong case based on ease of use for the client to keep this from happening. As a user, when i have base messages that are under the default 1mb, i should never hit a message too large error without having to understand the internals of kafka's compressed message protocol. 

Thinking of copying the java param `max_request_size`. The only remaining question is should this setting apply to all produce requests or just compressed requests? My initial reading of the code leads me to believe that applying this to all requests would be easier to implement but is that desirable?

emmett9001:
I'd say it should apply to all requests for consistency's sake

jofusa:
I have figured out my rdkafka test failures. calling _wait_all() on a rdkafka consumer causes it to hang.

Here is a review of the changes in the PR:

- Added auto_start and start methods to Producer and OwnedBroker to match consumer api and help with testing
- add a `max_request_size` param to the Producer init. defaults to 1000012 bytes. the default message size of kafka
- change OwnedBroker.fetch logic to build a batch of messages up to the size of `max_request_size` and not just drain the entire queue.
- Check if any individual message is larger then `max_request_size`. If it is mark it as failed in the `delivery_report_queue`

emmett9001:
Thanks @jofusa. This diff looks promising, but I will have to go over it in detail before I have a well-formed opinion about it. I'll aim to do that this week.

jofusa:
Can I get a bump on this. We have had to disable all compression for our producers. Eager to see this move forward.

emmett9001:
I can't see right now (on a phone), but please ensure that all tests are
passing in Travis. You can re-run certain test jobs in Travis if there are
ephemeral failures. Once all tests are passing in this branch I'll dig in.

I apologize if tests are passing now - they're not visible in the mobile UI.

On Friday, March 18, 2016, John Dennison <notifications@github.com> wrote:

> Can I get a bump on this. We have had to disable all compression for our
> producers. Eager to see this move forward.
>
> —
> You are receiving this because you were assigned.
> Reply to this email directly or view it on GitHub
> <https://github.com/Parsely/pykafka/pull/480#issuecomment-198597620>
>


-- 
Emmett Butler | Software Engineer
<http://www.parsely.com/?utm_source=Signature&utm_medium=emmett&utm_campaign=Signature>

pykafka/producer.py
```diff
@@ -540,7 +575,40 @@ def flush(self, linger_ms, release_pending=False):
         """
         self._wait_for_flush_ready(linger_ms)
         with self.lock:
-            batch = [self.queue.pop() for _ in range(len(self.queue))]
+            batch = []
+            batch_size_in_bytes = 0
+            while len(self.queue) > 0:
+                # TODO: Should we optimistically pop and renenqueue vs. peak
```
yungchin: I suppose that's safe as long as the lock is held, but the peeking approach shouldn't be any slower and so for me, I'd prefer it the way you do it now - a bit easier to reason about.

pykafka/producer.py
```diff
@@ -540,7 +575,40 @@ def flush(self, linger_ms, release_pending=False):
...
+                            "Message size larger then max_request_size: {}".format(max_request_size)
+                        )
+                        log.warning(exc)
+                        # bind the MessageSizeTooLarge error the delivery
+                        # report and remove it from the producer queue
+                        message = self.queue.pop()
+                        peaked_message.delivery_report_q.put((message, exc))
+                        # remove from pending message count
+                        self.increment_messages_pending(-1)
+                        break
```
yungchin: or `continue`?

pykafka/producer.py
```diff
@@ -540,7 +575,40 @@ def flush(self, linger_ms, release_pending=False):
...
         with self.lock:
-            batch = [self.queue.pop() for _ in range(len(self.queue))]
+            batch = []
+            batch_size_in_bytes = 0
+            while len(self.queue) > 0:
+                # TODO: Should we optimistically pop and renenqueue vs. peak
+                peaked_message = self.queue[-1]
+
+                if peaked_message and peaked_message.value is not None:
+                    if len(peaked_message.value) > max_request_size:
```
yungchin: You're better off using `len(peeked_message)` here - that also takes into account the length of the key and other overhead.

pykafka/producer.py
```diff
@@ -540,7 +575,40 @@ def flush(self, linger_ms, release_pending=False):
...
+                    elif batch_size_in_bytes + len(peaked_message.value) > max_request_size:
+                        log.debug("max_request_size reached. producing batch")
+                        # if we did not fully empty the queue. reset the
+                        # flush_ready so we send another batch immediately
+                        self.flush_ready.set()
+                        break
+
+                message = self.queue.pop()
+                if message.value is not None:
+                    batch_size_in_bytes += len(message.value)
```
yungchin: Same as above here: you want `len(message)`

pykafka/producer.py
```diff
@@ -540,7 +575,40 @@ def flush(self, linger_ms, release_pending=False):
...
+                        # report and remove it from the producer queue
+                        message = self.queue.pop()
+                        peaked_message.delivery_report_q.put((message, exc))
+                        # remove from pending message count
+                        self.increment_messages_pending(-1)
+                        break
+
+                    # test if adding the message would go over the
+                    # max_request_size. if it would, break out of loop
+                    elif batch_size_in_bytes + len(peaked_message.value) > max_request_size:
```
yungchin: Same `len` thing here, too.

tests/pykafka/test_producer.py
```diff
@@ -177,6 +180,145 @@ def test_null_payloads(self):
...
+        assert message.value == large_payload
+
+        for i in range(10):
+            prod.produce(large_payload)
+
+        reports = []
+        def ensure_all_messages_produced():
+            report = prod.get_delivery_report()
+            reports.append(report)
+            assert len(reports) == 10
```
yungchin: For me,  these assertions are a bit unintuitive - the function appends 1 report and then asserts there should be 10. I get how it works, it's mostly personal preference, so let's see what @emmett9001 thinks :)

One other downside of how `retry()` works is that the wait between the 10 calls is probably unnecessary for this particular callee, most of the time, and in that sense it slows the test down.

yungchin:
I'm not sure why the pypy tests fail (they also seem to be a pretty hard fail, the whole thing exits). I find it hard to see how the code changes here could be at fault for that?

yungchin:
@jofusa only catching up with earlier conversation up this thread now... About the thread-local delivery reports: the way that's been written, all it takes to get a global queue instead is overriding the report class (I think the pullreq would take maybe ten lines plus tests). We said at the time that we'd postpone implementing that option until someone actually had a need for it, so if you do, let's open a ticket for discussion.

tests/pykafka/test_producer.py
jofusa: The use of retry as the loop controller is an abuse for sure.

Let me take another stab at this assertion. I can change it to be more optimistic, try to get 10 delivery reports the first time and then fall back on the retry. My first stab would block on the delivery report but a test failure would hang on the assertion and freeze the entire test run.

pykafka/producer.py
```diff
@@ -222,9 +238,16 @@ def _setup_owned_brokers(self):
...
                 owned_broker.stop()
-                batch = owned_broker.flush(self._linger_ms)
-                if batch:
-                    queued_messages.extend(batch)
+
+                # loop becuase flush is not garentee to empty owned
+                # broker queue
+                while True:
+                    batch = owned_broker.flush(self._linger_ms, self._max_request_size)
+                    if batch:
```
emmett9001: Could lose a line without losing any clarity by rewriting this as
```python
if not batch:
    break
# do stuff
```

pykafka/producer.py
```diff
@@ -469,8 +492,12 @@ class OwnedBroker(object):
     :type messages_pending: int
     :ivar producer: The producer to which this OwnedBroker instance belongs
     :type producer: :class:`pykafka.producer.AsyncProducer`
+    :param auto_start: Whether the OwnedBroker should start flushing all
+    waiting messages and send to with kafka after __init__ is complete. If
```
emmett9001: Typo? `to with`

pykafka/producer.py
emmett9001: :+1: to @yungchin

pykafka/producer.py
```diff
@@ -540,7 +575,40 @@ def flush(self, linger_ms, release_pending=False):
         """
         self._wait_for_flush_ready(linger_ms)
         with self.lock:
-            batch = [self.queue.pop() for _ in range(len(self.queue))]
+            batch = []
+            batch_size_in_bytes = 0
+            while len(self.queue) > 0:
+                # TODO: Should we optimistically pop and renenqueue vs. peak
+                peaked_message = self.queue[-1]
```
emmett9001: This might seem pedantic, but I think it's really important for readability. I think you mean "peek" as in "to look at the last item in the queue" rather than "peak" meaning the top of something. Spelling the variable name properly will make it clearer what this code is trying to do.

pykafka/producer.py
```diff
@@ -540,7 +575,40 @@ def flush(self, linger_ms, release_pending=False):
...
+                    # max_request_size. if it would, break out of loop
+                    elif batch_size_in_bytes + len(peaked_message) > max_request_size:
+                        log.debug("max_request_size reached. producing batch")
+                        # if we did not fully empty the queue. reset the
+                        # flush_ready so we send another batch immediately
+                        self.flush_ready.set()
+                        break
+
+                message = self.queue.pop()
+                if message.value is not None:
```
emmett9001: I'm not sure this condition is necessary. `Message.__len__` takes the same condition into account, so I think it should be fine to remove this.

tests/pykafka/test_producer.py
```diff
@@ -5,12 +5,15 @@
 import time
 import unittest2
 from uuid import uuid4
+import sys
```
emmett9001: Unused import?

tests/pykafka/test_producer.py
```diff
@@ -177,6 +180,145 @@ def test_null_payloads(self):
...
+            large_payload,
+            partition_id=0,
+            delivery_report_q=delivery_report_queue
+        )
+
+
+        owned_broker.enqueue(msg)
+
+        max_request_size = 1000
+        assert max_request_size < len(msg.value)
```
emmett9001: It couldn't hurt here to check `len(msg)` to help validate that it's working.

tests/pykafka/test_producer.py
emmett9001: I think the current method is ok with an added comment explaining what's going on, but if you have something cleaner in mind I'd love to see it @jofusa.

emmett9001:
@jofusa Thanks for the pull request! I gave it a detailed look, and I think this is a welcome and necessary addition. I commented on a bunch of minor things, none of which should be too tough to fix. If the tests were passing, I'd say this would be ready for a merge after all comments are addressed, but we still need to figure out what's up with the PyPy tests.

I'm aiming for a new minor version release today, but I have no problem with incrementing the minor version again soon after this is merged.

tests/pykafka/test_producer.py
jofusa: I am trying come other approaches but they are not cleaner and they don't really work.

```
        reports = []
        def ensure_all_messages_produced():
            report = prod.get_delivery_report()
            reports.append(report)
            for i in range(10):
                try:
                    report = prod.get_delivery_report(block=False)
                    reports.append(report)
                except Empty:
                    pass
             assert len(reports) == 10
         retry(ensure_all_messages_produced)
```

emmett9001: In that case I'm fine with an explanatory comment here.

emmett9001:
I've discovered that the pypy failures stop happening if compression isn't used on the producer. Still not sure why that is.

jofusa:
Thanks for digging in. I was in the process of install pypy...

emmett9001:
After more investigation it's still not clear whether or not the compression issues are unique to my own system or if they're also causing problems on travis.

jofusa:
got it running on my vagrant.

```
(pypy) ➜  pykafka git:(jdennison/producer-callback) ✗ PYTHONPATH=. py.test tests/pykafka/test_producer.py -x
========================================================================================================================================================================== test session starts ===========================================================================================================================================================================
platform linux2 -- Python 2.7.10[pypy-5.0.1-final], pytest-2.9.1, py-1.4.31, pluggy-0.3.1 -- /home/vagrant/repos/pykafka/pypy/bin/pypy
cachedir: .cache
rootdir: /home/vagrant/repos/pykafka, inifile: setup.cfg
plugins: cov-2.2.1
collected 30 items

tests/pykafka/test_producer.py::ProducerIntegrationTests::test_async_produce PASSED
tests/pykafka/test_producer.py::ProducerIntegrationTests::test_async_produce_compression_large_message [1]    30723 segmentation fault (core dumped)  PYTHONPATH=. py.test tests/pykafka/test_producer.py -x
```

fun.

jofusa:
at least this conforms to the travis failure.

emmett9001:
You can use the `-s` flag to have it print its output to you as it runs. Let us know if you make any progress.

jofusa:
So this does not look like an issue related to this PR. Just python snappy pypy support.

If you could verify on your machine that help.

```
(pypy) ➜  pykafka git:(jdennison/producer-callback) ✗ cat s.py
from uuid import uuid4
import snappy

payload = b''.join([uuid4().bytes for i in range(10)])

c = snappy.compress(payload)
assert snappy.decompress(c) == payload
(pypy) ➜  pykafka git:(jdennison/producer-callback) ✗ python s.py
[1]    4587 segmentation fault (core dumped)  python s.py
(pypy) ➜  pykafka git:(jdennison/producer-callback) ✗ python
Python 2.7.10 (bbd45126bc69, Mar 18 2016, 21:35:08)
[PyPy 5.0.1 with GCC 4.8.4] on linux2
```

emmett9001:
@jofusa The confusing thing is that the pypy compression tests on master are still passing. It's not clear to me why they'd be able to pass there but fail for you.

jofusa:
Its a function of size. a small payload passes.

jofusa:
This passes
```
from uuid import uuid4
import snappy

payload = b''.join([uuid4().bytes for i in range(5)])

c = snappy.compress(payload)
assert snappy.decompress(c) == payload
```

emmett9001:
Ouch. I suppose we need to disable your failing test in PyPy and add large-payload tests in `tests/pykafka/utils/test_compression.py`. To disable to the failing test, you'll need the test-skipping monkeypatch from #416, which is very close to being ready to merge.

emmett9001:
You should be able to use the standard test skipping method to skip your compression test under PyPy. If there's a message size that the test can use that doesn't cause PyPy to explode, that's preferable. I can add some large-payload compression tests in a separate pull request.

jofusa:
i reduced the payload size for the compressed producer and added a comment. lets see if it passes

pykafka/producer.py
```diff
@@ -66,8 +66,10 @@ def __init__(self,
                  min_queued_messages=70000,
                  linger_ms=5 * 1000,
                  block_on_queue_full=True,
+                 max_request_size=1000012,
```
emmett9001: @jofusa Any particular reason this deviates from the JVM producer's default of `1048576`?

emmett9001:
@jofusa Looks great, thanks for dealing with the failing tests.

PR #486, opened 2016-03-08
Allow user to decide whether log undelivered messages as warnings or errors

jonsource:
I recently ran into an issue with undelivered messages - and noticed I log them only as info - this should be definitely bumped up a notch. On the other hand I can imagine a scenario where you are sending many thousands of messages and don't care much if a few go missing.

So I propose this change to the logging of undelivered messages - user can decide if it should be just a warning (the default) not to overwhelm the logs or an error to really stand out if something goes wrong.

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -67,8 +67,7 @@ class Producer(object):
                  linger_ms=5 * 1000,
                  block_on_queue_full=True,
                  sync=False,
-                 delivery_reports=False,
-                 log_undelivered_as_error=False):
+                 delivery_reports=False):
         """Instantiate a new AsyncProducer
 
         :param cluster: The cluster to which to connect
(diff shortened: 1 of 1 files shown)


emmett9001:
@jonsource I think it's fine to simply raise the level to `ERROR` in all cases. A message failing after retries is an error, and it makes sense to log it at that level.

jofusa:
+1 on the error level

jonsource:
On the other hand if you send many thousands of messages and encounter a server failure, your error log would be overwhelmed and possibly rendered useless. Maybe there should be a cap on the number of undelivered message errors?

jonsource:
I simplified the change - only logging undelivered messages as errors.

emmett9001:
@jonsource Looks good, thanks!

PR #522, opened 2016-04-11
Handle exceptions in intepreter shutdown in RequestHandler

benauthor:
Supersedes my weak attempt this AM at https://github.com/Parsely/pykafka/pull/521...

I am seeing assorted errors with some frequency originating in the handler function within RequestHandler._start_thread when using Pykafka in a web application that is run multithreaded under Gunicorn.

I see a recent commit d2a7e4c, which handles a TypeError caused by the messy nature of working with daemon threads. This is cool, but doesn't go quite far enough -- as the whole universe is getting deconstructed around the daemon thread, errors can originate in fairly hard-to-predict ways in this kind of context.

This patch uses an approach yanked from py3k's threading module -- just check to see if `sys` still exists before raising.

It's not a critical problem as the Python process is already going down, but causes a ton of log noise, which obscures real problems and is a bummer. Here are some of my errors from production:

```
Exception in thread 18: pykafka.RequestHandler.worker for broker-18.kafka.prclt.net:9092 (most likely raised during interpreter shutdown):
 File "/usr/lib/python2.7/threading.py", line 551, in __bootstrap_inner
 File "/usr/lib/python2.7/threading.py", line 504, in run
 File "/usr/local/lib/python2.7/dist-packages/pykafka/handlers.py", line 206, in worker
<type 'exceptions.AttributeError'>: 'NoneType' object has no attribute 'info'
```

```
Unhandled exception in thread started by <bound method Thread.__bootstrap of <Thread(1: pykafka.RequestHandler.worker for broker-1.kafka.prclt.net:9092, stopped daemon 140024375449344)>>
 File "/usr/lib/python2.7/threading.py", line 524, in __bootstrap
 self.__bootstrap_inner()
 File "/usr/lib/python2.7/threading.py", line 564, in __bootstrap_inner
 (self.name, _format_exc()))
 File "/usr/lib/python2.7/traceback.py", line 240, in format_exc
 etype, value, tb = sys.exc_info()
AttributeError: 'NoneType' object has no attribute 'exc_info'
```

### DIFF
--- a/pykafka/handlers.py
+++ b/pykafka/handlers.py
@@ -26,6 +26,7 @@ import gevent.queue
 import gevent.socket as gsocket
 import logging
 import socket as pysocket
+import sys
 import threading
 import time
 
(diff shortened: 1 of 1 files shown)


emmett9001:
Looks good to me based on the precedent [here](http://stackoverflow.com/a/18098884/735204). Thanks @benauthor!

benauthor:
💯

Issue #527, opened 2016-04-18
Allow more BalancedConsumers than partitions on a topic


jianbin-wei:
In case of one topic with one partition, if two balanced consumers are created within the same group for the topic, the later one would raise exception.  When the first balanced consumer is down, messages are not consumed anymore

However, in Java kafka client, the later kafka consumer would continue to consume the topic's messages.  The behavior is more resilient.

> 2016-04-18 06:25:09,420 ERROR           balancedconsumer        start:304       31791   Stopping consumer in response to error
Traceback (most recent call last):
  File "/service-venv/lib/python2.7/site-packages/pykafka/balancedconsumer.py", line 299, in start
    self._add_self()
  File "/service-venv/lib/python2.7/site-packages/pykafka/balancedconsumer.py", line 541, in _add_self
    raise KafkaException("Cannot add consumer: more consumers than partitions")

**PyKafka version**: 2.3.1
**Kafka version**: 0.8.2.1

emmett9001:
@jianbin-wei If you have a topic with a single partition, I'd suggest using a pair of `SimpleConsumer`s to consume it with some degree of resilience. A `BalancedConsumer` cannot be added to a full group, since doing so would violate the semantics Kafka defines for balanced consumption. The subset of messages that the N+1th consumer of an N-partition topic should consume is not defined by Kafka's balancing semantics - should it consume the whole stream? A random subset of partitions? A single partition? Since this behavior is undefined, pykafka chooses to disallow it. Pykafka still supports more than N consumers of a single topic if those consumers are `SimpleConsumer` instances.

jianbin-wei:
> A BalancedConsumer cannot be added to a full group, since doing so would violate the semantics Kafka defines for balanced consumption. The subset of messages that the N+1th consumer of an N-partition topic should consume is not defined by Kafka's balancing semantics - should it consume the whole stream? A random subset of partitions? A single partition?

From what I see the behavior of Kafka console consumer is good.  The N+1th consumer is idle as its assigned partition is none.  When one consumer dies, the N+1th consumer starts to consume (after rebalancing).

jianbin-wei:
thanks a lot.

jianbin-wei:
Actually this one is needed for #354 .  In that case, for one consumer to consume from multiple topics with different number of partitions, it would be better to idle around and consume if needed.

yungchin:
[...]

PR #554, opened 2016-05-24
Allow more consumers that partitions for a balanced consumer

vitalyli:
Addresses issue: https://github.com/Parsely/pykafka/issues/527


pykafka/balancedconsumer.py
```diff
@@ -537,13 +537,6 @@ def _set_watches(self):
     def _add_self(self):
         """Register this consumer in zookeeper.
 
-        This method ensures that the number of participants is at most the
-        number of partitions.
-        """
```
emmett9001: This endquote needs to be kept.

emmett9001:
Thanks @vitalyli, this is a good start. For this to be merged, the test will need to be runnable in Travis. That means it should probably avoid creating new threads wherever possible, since daemon threads can cause Travis to hang. Producing 10000 messages will probably make test runs take much longer than they need to. Additionally, this test only barely verifies that there are no crashes in the code you've written, but since there are components running on separate threads, crashes aren't guaranteed to be visible to Travis.

I suggest moving your test case to [`test_balancedconsumer.py`](https://github.com/Parsely/pykafka/blob/master/tests/pykafka/test_balancedconsumer.py), removing the use of threads, having it send only the minimum number of messages necessary to verify that consumers above the partition limit do indeed idle, and adding some `self.assertTrue` calls that ensure that extra consumers are able to start consuming if a space in the group opens up.

Broadly, this is the logic I'd think would be required for a test of this functionality (pseudocode):
```python
# produce some test messages to 2-partition topic

c1 = topic.get_balanced_consumer(group)
c2 = topic.get_balanced_consumer(group)
c3 = topic.get_balanced_consumer(group)

# verify that the first two consumers can consume correctly
m1 = c1.consume()
m2 = c2.consume()
self.assertNotEqual(m1, None)
self.assertNotEqual(m2, None)

# verify that the last consumer doesn't get any messages
m3 = c3.consume()
self.assertEqual(m3, None)

# verify that the last consumer can start consuming if some partitions become available
c2.stop()
m2 = c2.consume()
m3 = c3.consume()
self.assertEqual(m2, None)
self.assertNotEqual(m3, None)
```

This is a high-level view of what I think the test for this change needs to look like.

pykafka/balancedconsumer.py
vitalyli: oops; fixing

vitalyli:
@emmett9001 I'm not really familiar with travis and quite new to this project; literally spent one day banging my head against Kafka and found what works on my actual cluster. The test is by no means exhaustive. I'm in a process of setting up 10 or 20 dockers on ten machines; I will verify it there, but it will take a bit longer. If you guys want to take it over and make a better test for it, I'm ok closing this PR.

emmett9001:
@vitalyli Sounds fine, thanks for being so responsive. Of course, if you find that this change works in your own environment, you can certainly run a forked version of pykafka until this change makes it into a proper release. Even with the test being finished right now, it would probably take a few weeks to cut a new release. I'll prioritize this for the next time I do some work on open issues.

vitalyli:
Sounds good thanks!

PR #555, opened 2016-05-24
extra consumers

### LINKED ISSUE #527: Allow more BalancedConsumers than partitions on a topic

emmett9001:
This pull request fixes #527 by removing the upper limit on the number of consumers in a group. It also adds some tests to ensure that messages are not duplicated across a consumer group when there are more consumers than partitions.

### DIFF
--- a/tests/pykafka/test_balancedconsumer.py
+++ b/tests/pykafka/test_balancedconsumer.py
@@ -181,21 +181,26 @@ class BalancedConsumerIntegrationTests(unittest2.TestCase):
 
         def verify_extras(consumers, extras_count):
             messages = [c.consume() for c in consumers]
-            nones = [a for a in messages if a is None]
             successes = [a for a in messages if a is not None]
+            nones = [a for a in messages if a is None]
+            attempts = 0
+            while len(nones) != extras_count and attempts < 5:
+                messages = [c.consume() for c in consumers]
+                successes = [a for a in messages if a is not None]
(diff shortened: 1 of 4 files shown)


emmett9001:
The new test is still a bit unstable, I'll have to have another look.

tests/pykafka/test_balancedconsumer.py
```diff
@@ -170,6 +170,55 @@ def get_balanced_consumer(self, consumer_group, **kwargs):
             **kwargs
         )
 
+    def test_extra_consumer(self):
```
emmett9001: FYI @vitalyli this is what the test ended up looking like.

PR #569, opened 2016-06-10
zk consumer group bytes literal

### LINKED ISSUE #567: Python3 consumer groups
When using Python 3.5.1 with PyKafka 2.3.1, the consumer group names and topic names must be bytes (#376). Thanks for updating the API to note this.

However, inside zookeeper, the consumer groups are still stored with a literal 'b'.

```python
import pykafka
client = pykafka.KafkaClient(hosts='localhost:9092')

topic = client.topics[b'test_topic']
consumer = topic.get_balanced_consumer(consumer_group=b'test_group', zookeeper_connect='localhost:2181/kafka')
```

Using `zkCli.sh`,
```
[zk: localhost:2181(CONNECTED) 0] ls /kafka/consumers
[b'test_group']
[zk: localhost:2181(CONNECTED) 0] ls /kafka/brokers/topics
['test_topic']
```

So the topics are properly created (without the literal b), but the consumer groups don't appear to be.

This isn't a deal breaker as I don't share consumer groups across clients, but thought I should bring it to your attention. I believe the offending code is here:

https://github.com/Parsely/pykafka/blob/master/pykafka/balancedconsumer.py#L237-L241

If I've mis-used the API, please do let me know. Thanks for your hard work and a great library, nonetheless.

emmett9001:
This pull request fixes #567 by converting consumer group names to strings before storing them in zookeeper.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -257,7 +257,7 @@ class BalancedConsumer(object):
             module=self.__class__.__module__,
             name=self.__class__.__name__,
             id_=hex(id(self)),
-            group=self._consumer_group
+            group=get_string(self._consumer_group)
         )
 
     def _raise_worker_exceptions(self):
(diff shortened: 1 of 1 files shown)


echu:
lol. I just forked and made the exact same changes, but looks like you've beat me to the PR. Thanks.

FWIW, I also added `get_string` on line 260 (in `__repr__`).

PR #572, opened 2016-06-18
Bump PyPy version

thedrow:


thedrow:
Do we want to test with the old PyPy version as well?

emmett9001:
I think it's alright to only test with the latest PyPy version.

PR #573, opened 2016-06-19
Bump librdkafka version to 0.9.1

thedrow:


emmett9001:
Looks like the tests are failing, so we won't be able to merge this now.

edenhill:
librdkafka tests now uses [trivup](https://github.com/edenhill/trivup) for bringing up a test environment with Kafka, ZK, KerberosKDC, but this approach is optional and the default 'make' target in tests/ still runs without it.

emmett9001:
In that case, I wonder what we're doing to cause [this](https://travis-ci.org/Parsely/pykafka/jobs/138682730#L915) to happen in our Travis runs.

edenhill:
Looks like tox is picking up librdkafka's python based test scripts, it probably shouldn't.

emmett9001:
@thedrow please try changing [this line](https://github.com/thedrow/samsa/blob/a0e9066ae6a0c2be7ac80171d055fad682c24388/.travis.yml#L67) to `tox tests` and pushing to retrigger the travis build. This should stop tox from running non-pykafka tests.

emmett9001:
Looks like the rdkafka tests are now [failing](https://travis-ci.org/Parsely/pykafka/builds/139577046) against Kafka 0.8.2.

edenhill:
When using librdkafka with older broker versions you will need to set `broker.version.fallback=<yourbrokerversion>` which adjusts the protocol version(s) to use.

thedrow:
@edenhill Is this an environment variable?

edenhill:
@thedrow A librdkafka configuration property (https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md)

edenhill:
https://github.com/edenhill/librdkafka/wiki/Broker-version-compatibility

emmett9001:
Thanks for the tip @edenhill. @thedrow, it looks like this parameter needs to be passed in [here](https://github.com/Parsely/pykafka/blob/39bcf26f04f44c376c0f27074052504185dfa6f6/pykafka/rdkafka/simple_consumer.py#L186) and [here](https://github.com/Parsely/pykafka/blob/39bcf26f04f44c376c0f27074052504185dfa6f6/pykafka/rdkafka/producer.py#L101). For this to work, those classes will need to know the intended protocol version, which will require a new parameter, probably on `KafkaClient`. If you'd like to give this a shot, please feel free. Otherwise, I'll put this on my list and make the necessary changes when I can, after which I'll merge your branch into mine and try this out again.

emmett9001:
#584 has been merged, so this can now be closed too. Thanks for the help, @edenhill and @thedrow.

PR #577, opened 2016-07-10
Type hinting: hostnames are strings, not bytes

steffann:
These type hints say that hosts and zookeeper_hosts should be bytes, but they are actually strings.

### DIFF
--- a/pykafka/client.py
+++ b/pykafka/client.py
@@ -56,10 +56,10 @@ class KafkaClient(object):
         :param hosts: Comma-separated list of kafka hosts to which to connect.
             If `ssl_config` is specified, the ports specified here are assumed
             to be SSL ports
-        :type hosts: bytes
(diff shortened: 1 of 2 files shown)


emmett9001:
@steffann These parameters are labeled as `bytes` to clarify the difference between strings in python 2 and python 3. In python 2, you're right that these are strings - that's because python 2 strings are bytes. In python 3, supplying a string to these arguments may not work, because python 3 strings are not bytes. There is some internal logic related to unicode handling that requires that these parameters are bytes.

steffann:
Sorry, you misunderstood my issue. Providing bytes for those arguments actually raises exceptions: they *have* to be strings but the type hints in the comments are just wrong.

emmett9001:
Thanks @steffann, I checked this out and it appears you're correct.

PR #578, opened 2016-07-13
include consumer id and hostname in offset metadata

emmett9001:
This pull request adds a bit of useful information to the offset metadata field in offset commit requests. The consumer ID can be useful when debugging certain failure modes, as we've seen at Parsely. Comments welcome if there is any other information it might make sense to include here.

@dan-blanchard, if consumers were running this code, you'd be able to see consumer IDs and hostnames in our lag monitor.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -31,7 +31,7 @@ import weakref
 
 from .common import OffsetType
 from .utils.compat import (Queue, Empty, iteritems, itervalues,
-                           range, iterkeys, get_bytes)
+                           range, iterkeys, get_bytes, get_string)
 from .exceptions import (UnknownError, OffsetOutOfRangeError, UnknownTopicOrPartition,
                          OffsetMetadataTooLarge, GroupLoadInProgress,
                          NotCoordinatorForGroup, SocketDisconnectedError,
@@ -799,7 +799,7 @@ class OwnedPartition(object):
         self.fetch_lock = handler.RLock() if handler is not None else threading.RLock()
         # include consumer id in offset metadata for debugging
         self._offset_metadata = {
-            'consumer_id': self._consumer_id,
+            'consumer_id': get_string(self._consumer_id),
             'hostname': socket.gethostname()
         }
 
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -802,6 +802,8 @@ class OwnedPartition(object):
             'consumer_id': get_string(self._consumer_id),
             'hostname': socket.gethostname()
         }
+        # precalculate json to avoid expensive operation in loops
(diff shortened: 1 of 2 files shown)


luka1983:
Tnx! :+1: Btw, do you maybe know if there is a way to see number of currently active non-zk consumers in a group? Unfortunately not all of my consumers can be pykafka consumers. Tnx!

pykafka/simpleconsumer.py
```diff
@@ -777,12 +788,18 @@ def __init__(self, partition, handler=None, semaphore=None, compacted_topic=Fals
...
         self._messages = Queue()
         self._messages_arrived = semaphore
         self._is_compacted_topic = compacted_topic
         self.last_offset_consumed = -1
         self.next_offset = 0
         self.fetch_lock = handler.RLock() if handler is not None else threading.RLock()
+        # include consumer id in offset metadata for debugging
+        self._offset_metadata = {
+            'consumer_id': self._consumer_id,
+            'hostname': socket.gethostname()
```
dan-blanchard: I think we probably want to use `platform.node()` here instead of `socket.gethostname()`. I have seen it work more reliably (as in, it returns `"localhost"` on fewer systems). It's not guaranteed to be fully-qualified, if that's a problem, but neither is `socket.gethostname()`.

emmett9001: Sure, I'll make that change. `self._consumer_id` does already contain `socket.gethostname()` under normal circumstances, so now we'll have both!

emmett9001:
@luka1983 I'm not sure what you mean by "non-zk consumers". In general, querying zookeeper is how you'd get information about the number of consumer processes currently reading a topic. If the consumer you're using isn't talking to zookeeper, I don't believe there's a way to know about it directly from another process, at least under kafka 0.8.2

luka1983:
@emmett9001 I was talking about consumers that use new consumer API and do not rely on ZK for persistence. The protocol implements DescribeGroups API but this is available only from 0.9 and up. Since pykafka supports balanced consumers that use new consumer API I've thought that this admin api might be supported already, but as you said, with 0.8.2 you have to rely on ZK for this. Thanks for clarification!

emmett9001:
@luka1983 Thanks for clarifying. Pykafka doesn't currently support the admin API.

pykafka/simpleconsumer.py
```diff
@@ -849,7 +868,7 @@ def build_offset_commit_request(self):
             self.partition.id,
             self.last_offset_consumed + 1,
             int(time.time() * 1000),
-            b'pykafka'
+            get_bytes('{}'.format(json.dumps(self._offset_metadata)))
```
dan-blanchard: Ya know, if this function gets called a lot, it would probably be better to store the JSON string version of `self._offset_metadata` instead of dumping again every time, since that's super slow.

PR #585, opened 2016-07-27
nonbusy consumer fetch waiting

### LINKED ISSUE #581: Fetcher thread will cause high cpu usage when queues of partitions are full.
If queues of all owned partitions are full, no fetch requests will be generated. [This codes will be skipped](https://github.com/Parsely/pykafka/blob/39bcf26f04f44c376c0f27074052504185dfa6f6/pykafka/simpleconsumer.py#L730-L733). And finally, `fetch` does nothing. So the fetcher thread will run `fetch` 10,000 times per second.
I don't know if it is your expected behavior.

Steps:
1. create a consumer with queued_max_messages=1.
2. start the consumer without consuming.
3. send messages to make all queues full.
4. run top to monitor cpu usage.

emmett9001:
This pull request fixes #581 by using an `Event` to implement nonbusy waiting in the `SimpleConsumer` when all partition queues are full. This implementation could be improved by eventing per partition instead of using a single `Event`, but doing so would require a significant refactor. This version appears to work well and solves the issue raised by @fortime.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -774,11 +774,11 @@ class SimpleConsumer(object):
 
     def _wait_for_slot_available(self):
         """Block until at least one queue has less than `_queued_max_messages`"""
-        if all(op.message_count > self._queued_max_messages
+        if all(op.message_count >= self._queued_max_messages
                for op in itervalues(self._partitions)):
             for op in itervalues(self._partitions):
                 op.fetch_lock.acquire()
-            if all(op.message_count > self._queued_max_messages
+            if all(op.message_count >= self._queued_max_messages
                    for op in itervalues(self._partitions)):
                 self._slot_available.clear()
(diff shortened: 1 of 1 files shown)


pykafka/simpleconsumer.py
```diff
@@ -760,6 +772,19 @@ def unlock_partitions(parts):
                     success_handler=_handle_success)
                 unlock_partitions(iterkeys(partition_reqs))
 
+    def _wait_for_slot_available(self):
+        """Block until at least one queue has less than `_queued_max_messages`"""
+        if all(op.message_count > self._queued_max_messages
```
fortime: I think 
op.message_count > self._queued_max_messages
should be
op.message_count >= self._queued_max_messages

emmett9001: Good point.

PR #593, opened 2016-08-18
Adding GroupHashingPartitioner

lucrussell:
This is a specialized HashingPartitioner which can divide incoming messages between a configurable subset of the available partitions. For example, given a total of sixteen partitions, one could specify a group size of four to smooth the distribution of incoming messages across a group of four partitions, thus avoiding hot spots in one partition.


emmett9001:
Thanks for the contribution @lucrussell, looks great!

lucrussell:
Thanks!

Issue #498, opened 2016-03-11
Producer blocks when broker is unreachable


Jox4:
```python
logger.info('Sending events...')
for sip_event in self.events_list:
     try:
         self.sip_producer.produce(msg)
     except:
         while True:
             logger.error('Could not connect to Kafka')
             logger.error(traceback.format_exc())
             self.reconnect()
             try:
                 self.sip_producer.produce(msg)
                 break
             except:
                 pass
```
This code is blocked when server is disconected. Don't catch timeout or any exception. With 2.1.1 is running well.

**PyKafka version**: 2.2.1
**Kafka version**: 2.9.1

emmett9001:
Hi @Jox4, thanks for the ticket. Can you be more specific about the behavior you're seeing? How are you creating your `Producer` instance, and what keyword arguments are you using? What's the replication factor of your topic, how many brokers are you running, and how many of those brokers are going down during your test?

Jox4:
Client instance:
`self.client = StoneKafka(
            hosts="server_ip:server_port",
            socket_timeout_ms=10 * 1000,
            offsets_channel_socket_timeout_ms=8 * 1000,
        )`

It's a simply test of our application: 1 broker, 1 replication factor and 128 partitions (for the balanced consumers).


Thanks for your attention

emmett9001:
We've seen this same issue happen at Parsely, so we may be able to dig up some more server logs that will help us fix it.

emmett9001:
@Jox4 I'm unable to reliably replicate the hang you're mentioning, though I believe I've come across it once. Are you opposed to using delivery reports? This is the standard method that pykafka provides to verify the receipt of produced messages. Maybe there's some reason you're unable to use them?

emmett9001:
Potentially related traceback from a Parse.ly test machine
```
2016-07-12 20:57:07,316 - pykafka.producer - DEBUG - Sending 944 messages to broker 12
2016-07-12 20:57:07,801 - pykafka.producer - WARNING - Broker ue1d-kafka-beta1a.lan.cogtree.com:9092 disconnected. Retrying.
2016-07-12 20:57:07,801 [WARNING] pykafka.producer - Broker ue1d-kafka-beta1a.lan.cogtree.com:9092 disconnected. Retrying.
2016-07-12 20:57:07,801 - pykafka.cluster - DEBUG - Updating cluster, attempt 1/3
[...]

PR #594, opened 2016-08-19
don't wait for flush_ready during producer updates

emmett9001:
This pull request addresses an issue likely related to #498. It fixes a bug causing queue flushes on `Producer.update()` to block the main thread for seconds or more, depending on the maximum queue size configured. In Parsely's internal case, this bug can cause the main thread to block for around five minutes at a time.


emmett9001:
cc @jofusa, neither of us caught this when it was initially written.

jofusa:
good catch.

PR #615, opened 2016-11-04
Fix 'raise StopIteration' in generator (PEP 479)

zware:
This is a simple change to prevent `PendingDeprecationWarning`s on Python 3.5.  There is no change in semantics with any supported interpreter.


emmett9001:
Thanks for pointing this out, @zware!

zware:
Thanks for the merge!

PR #626, opened 2016-12-08
catch struct.error in producer

### LINKED ISSUE #598: unpack_from requires a buffer of at least `xxx` bytes
I am using current master code.

I have a consumer consuming message from one Kafka topic, when some events occurred, it triggers a pure Python producer to produce some messages to another topic.

Traceback:

``` python
  File "/home/brm_index/venv/local/lib/python2.7/site-packages/brm_index/workers/base.py", line 140, in index_bulk
    to_index = self.map_docs(docs, source)
  File "/home/brm_index/venv/local/lib/python2.7/site-packages/brm_index/workers/basic.py", line 134, in map_docs
    self.on_mapped(body)
  File "/home/brm_index/venv/local/lib/python2.7/site-packages/brm_index/workers/basic.py", line 218, in on_mapped
    k: body.get(k) for k in self.relation_keys
  File "/home/brm_index/venv/local/lib/python2.7/site-packages/brm_index/workers/base.py", line 348, in schedule_relation
    json_util.dumps(message, ensure_ascii=False).encode('utf-8'))
  File "/home/brm_index/venv/local/lib/python2.7/site-packages/pykafka/producer.py", line 350, in produce
  File "/home/brm_index/venv/local/lib/python2.7/site-packages/pykafka/producer.py", line 204, in _raise_worker_exceptions
    raise ex
error: unpack_from requires a buffer of at least 27949 bytes
```

It seems that the producer can't restore to normal state after the exception, causing the process hanging.

**PyKafka version**: @b2ca8eb
**Kafka version**: 0.9.0

emmett9001:
This pull request changes the `Producer` to retry produce requests when they result in `struct.error` as in [this issue](https://github.com/Parsely/pykafka/issues/598).

Fixes #598

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -21,6 +21,7 @@ __all__ = ["Producer"]
 from collections import deque
 import logging
 import platform
+import struct
 import sys
 import threading
 import traceback
@@ -449,8 +450,8 @@ class Producer(object):
                     to_retry.extend(
                         (mset, exc)
                         for mset in _get_partition_msgs(partition, req))
-        except SocketDisconnectedError as exc:
-            log.warning('Broker %s:%s disconnected. Retrying.',
+        except (SocketDisconnectedError, struct.error) as exc:
+            log.warning('Error encountered when producing to broker %s:%s. Retrying.',
                         owned_broker.broker.host,
                         owned_broker.broker.port)
             try:


emmett9001:
Since we don't know the root cause of this buffer size mismatch, this change does raise the risk that we'll retry infinitely on a case that really shouldn't be retried. The alternative is to just treat this exception as unhandleable and swallow it.

cc @msukmanowsky @kbourgoin @dan-blanchard

dan-blanchard:
> this change does raise the risk that we'll retry infinitely on a case that really shouldn't be retried.

Since we don't know what's causing these this makes me a bit nervous. Do we have any reason to believe retrying would fix this error?

Another alternative would be to keep track of what things you've retried and just prevent infinite retries in general, but that's a larger change.

emmett9001:
I suppose if it's a choice between infinitely retrying or risking the loss of some messages, I'd rather the latter.

Issue #630, opened 2016-12-13
deprecate SimpleConsumer's consumer_id parameter


emmett9001:
According to the Kafka protocol documentation, `consumer_id` is invalid on group requests from non-managed consumers.

> Note that when this API is used for a "simple consumer," which is not part of a consumer group, then the generationId must be set to -1 and the memberId must be empty (not null). Additionally, if there is an active consumer group with the same groupId, then the commit will be rejected (typically with an UNKNOWN_MEMBER_ID or ILLEGAL_GENERATION error).

That means that [the consumer_id parameter](https://github.com/Parsely/pykafka/blob/a9dd498392a2f78682719f44c56229cfff486c30/pykafka/simpleconsumer.py#L76) on `SimpleConsumer` should not actually be allowed. It should be deprecated in the next release and removed in the next major version.

The reason this parameter exists at all is because it's [used](https://github.com/Parsely/pykafka/blob/a9dd498392a2f78682719f44c56229cfff486c30/pykafka/balancedconsumer.py#L411) by `ManagedBalancedConsumer`. The interface should be adjusted to allow `ManagedBalancedConsumer` to set this value appropriately for its needs without making it part of the public interface.

Andrew Boot from the [google group thread](https://groups.google.com/forum/#!topic/pykafka-user/2cBD-nWP-4Y) has offered to work on a pull request for this.

bootandy:
How do you see the redesigned classes looking?

I would guess this:
OldSimpleConsumer will subclass NewSimpleConsumer. 
NewSimpleConsumer hides the deprecated params to stop people like me setting them :-) and ManagedBalanceConsumer can create an OldSimpleConsumer just like it does now.

emmett9001:
@bootandy as I look more deeply at this, I'm realizing a basic issue is that [this logic](https://github.com/Parsely/pykafka/blob/c3c5507f1106716af8a935ee2253c94f58751efb/pykafka/simpleconsumer.py#L197) in `SimpleConsumer.__init__` depends on `consumer_id` being set appropriately. This means that removing it from the kwargs list will require some rethinking of how the initial consumer setup happens. It's not entirely clear to me what the best way to handle this is - deprecating and later removing the kwarg would require a redesign, and keeping it around leads to the interface problem you encountered. It seems like bad practice to keep the kwarg and have it come with a warning not to use it, or to make the kwarg "private" (like `_consumer_id`). I wonder if there are examples of this kind of thing being handled in other open source projects.

[...]

PR #631, opened 2016-12-14
SimpleConsumer: add deprecated to docstring

bootandy:
SimpleConsumer: add deprecated to docstring.

SimpleConsumer constructor params: generation_id, consumer_id are marked as deprecated in docstring

Also extend docstring for UnknownMemberId to explain one cause of the error

https://github.com/Parsely/pykafka/issues/630

### DIFF
--- a/pykafka/client.py
+++ b/pykafka/client.py
@@ -23,7 +23,6 @@ from .handlers import ThreadingHandler, GEventHandler
 import logging
 
 from .cluster import Cluster
-from .handlers import ThreadingHandler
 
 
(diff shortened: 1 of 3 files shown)


emmett9001:
Thanks @bootandy, this is great if we decide to actually remove these parameters. See my [comment](https://github.com/Parsely/pykafka/issues/630#issuecomment-267414870) for why I'm not sure we want to do that. Also, if we do decide to merge this, it has an unrelated change in `client.py` that I think should be removed from this PR.

PR #632, opened 2016-12-15
run tests against 10.0.1

emmett9001:


emmett9001:
[Apparently](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md) `broker.version.fallback` is incompatible with librdkafka running against 0.10 brokers.

PR #634, opened 2016-12-17
Do not create request handlers after unsuccessful try to connect to Broker

smueller18:
Running pykafka without this modification, every unsuccessful try to connect to a broker is followed be the [creation of a request handler](https://github.com/smueller18/pykafka/blob/bugfix/endless_request_generation/pykafka/broker.py#L213) which won't be terminated automatically.
This fix prevents a creation of a request handler by throwing the socket exception [right before the handlers would be created](https://github.com/smueller18/pykafka/blob/bugfix/endless_request_generation/pykafka/broker.py#L212).


emmett9001:
Thanks for the contribution @smueller18. Merging this would cause issues like #543 to reappear, so I'm afraid we can't merge it as-is. I do agree, though, that it seems wrong that `BrokerConnection.connect()` can return without raising an error after having failed to create a connection. It should raise an exception, but issues like #543 require that all callers of `BrokerConnection.connect()` become aware of the possibility of this exception. The same applies to `Broker.connect()`, which should also raise an exception on a failed connection.

smueller18:
This pull request can be ignored because it is treated by #636, explained in #635.

Issue #635, opened 2016-12-19
raise connection errors from BrokerConnection.connect()


emmett9001:
[This change](https://github.com/Parsely/pykafka/pull/564) may have been ill-conceived, as pointed out in #634 by @smueller18. `BrokerConnection.connect()` failing to connect to a broker and returning without any indication of that error makes no sense to me - a proper interface for this function would have it raising the appropriate exception upon failure to connect. `BrokerConnection.connect()` and `Broker.connect()` should both be changed to raise connection exceptions to their callers, and the callers should be adjusted to expect these exceptions such that issues like #543 don't reoccur.

**PyKafka version**: https://github.com/Parsely/pykafka/commit/8f4617c42a3508d97d02e05a590e33482bf89b9a

emmett9001:
Closed with #636

PR #636, opened 2016-12-19
raise connection failure exceptions from BrokerConnection.connect()

emmett9001:
This pull request is a first attempt at fixing #635. It makes `BrokerConnection.connect()` raise exceptions on connection failure instead of silently passing. This also means that callers become responsible for handling this exception, which in most cases they apparently already are. In the few cases where they are not are fixed by this pull request.

@smueller18 please feel free to test this branch and let me know if you have suggestions or clarifications. Thanks.

### DIFF
--- a/pykafka/broker.py
+++ b/pykafka/broker.py
@@ -35,15 +35,13 @@ from .utils.compat import range, iteritems, get_bytes
 log = logging.getLogger(__name__)
 
 
-def _check_handler(handler_name="_req_handler"):
-    def decorator(fn):
-        """Ensures that the given handler is not None before calling fn"""
-        def wrapped(self, *args, **kwargs):
-            if getattr(self, handler_name) is None:
-                raise SocketDisconnectedError
-            return fn(self, *args, **kwargs)
-        return wrapped
-    return decorator
+def _check_handler(fn):
+    """Ensures that self._req_handler is not None before calling fn"""
+    def wrapped(self, *args, **kwargs):
+        if self._req_handler is None:
+            raise SocketDisconnectedError
+        return fn(self, *args, **kwargs)
+    return wrapped
 
 
 class Broker(object):
@@ -272,7 +270,7 @@ class Broker(object):
             self._req_handlers[connection_id] = handler
         return self._req_handlers[connection_id]
 
-    @_check_handler()
+    @_check_handler
     def fetch_messages(self,
                        partition_requests,
                        timeout=30000,
@@ -299,7 +297,7 @@ class Broker(object):
         # XXX - this call returns even with less than min_bytes of messages?
         return future.get(FetchResponse)
 
-    @_check_handler()
+    @_check_handler
     def produce_messages(self, produce_request):
         """Produce messages to a set of partitions.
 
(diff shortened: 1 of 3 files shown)


smueller18:
I tested the bugfix branch with following result:
If a broker cannot establish a connection [here](https://github.com/Parsely/pykafka/blob/bugfix/raise_connection_failures/pykafka/broker.py#L216), an exception is thrown. Therefore a `RequestHandler` is never being created and remains `None` for the current broker. But the exception is only logged [here](https://github.com/Parsely/pykafka/blob/bugfix/raise_connection_failures/pykafka/broker.py#L101) and not raised again. For that reason the method `Cluster._request_metadata()` still tries to send a request at function [`broker.request_metadata(topics)`](https://github.com/Parsely/pykafka/blob/bugfix/raise_connection_failures/pykafka/cluster.py#L249). But because the `RequestHandler` is `None` there will be an AttributeError [at this line](https://github.com/Parsely/pykafka/blob/bugfix/raise_connection_failures/pykafka/broker.py#L327): `AttributeError: 'NoneType' object has no attribute 'request'`.
Maybe checking for `self._req_handler is None` [at line 327](https://github.com/Parsely/pykafka/blob/bugfix/raise_connection_failures/pykafka/broker.py#L327) and raise a `SocketDisconnectedError` on `True` is a good option here instead of throwing a `AttributeError` to the main thread.

Here is the logging until the AttributeError:
```
Connected to pydev debugger (build 163.8233.8)
DEBUG:pykafka.cluster:Updating cluster, attempt 1/3
DEBUG:pykafka.connection:Connecting to localhost:9092
ERROR:pykafka.connection:Failed to connect to localhost:9092
WARNING:pykafka.broker:Failed to connect newly created broker for localhost:9092
ERROR:pykafka.cluster:Unable to connect to broker localhost:9092. Continuing.
ERROR:pykafka.cluster:'NoneType' object has no attribute 'request'
  File "/home/stephan/anaconda3/lib/python3.5/site-packages/pykafka-2.6.0.dev2-py3.5-linux-x86_64.egg/pykafka/cluster.py", line 249, in _request_metadata
    response = broker.request_metadata(topics)
  File "/home/stephan/anaconda3/lib/python3.5/site-packages/pykafka-2.6.0.dev2-py3.5-linux-x86_64.egg/pykafka/broker.py", line 327, in request_metadata
    future = self._req_handler.request(MetadataRequest(topics=topics))
AttributeError: 'NoneType' object has no attribute 'request'
```

emmett9001:
@smueller18 I like that solution. I don't think that `Broker.__init__` should raise the connection exception since an instantiated broker with no connection is still a valid one.

smueller18:
@emmett9001 sounds good to me

emmett9001:
@smueller18 Please let me know how the latest fix works for you. It should now result in all failed broker connections raising `SocketDisconnectedError`.

smueller18:
Yes, all failed broker connections ends up in a `SocketDisconnectedError`.
One thing though: In [cluster.py - _request_metadata()](https://github.com/Parsely/pykafka/blob/bugfix/raise_connection_failures/pykafka/cluster.py#L254), the stack trace of all errors is printed. It would be nicer not to print stack traces of `SocketDisconnectedError`.

emmett9001:
I agree that `log.exception` for every exception type there is a bit noisy. Want to open a separate PR for that change?

PR #638, opened 2016-12-28
fix socket recv error(SIGPIPE)

hallelujah-shih:


hallelujah-shih:
Brokers would fail to reconnect after the socket is reset by peer because the Exception 104 raised after the sending of requests is not going to be captured properly.

And since there's no way to make brokers reconnect mannually via client, the Exception 32 will be repeatedly raised till the process is killed.

emmett9001:
See the description of #640 for more information. @hallelujah-shih sorry for the duplicate PR, I read your issue before this pull request.

PR #633, opened 2016-12-15
allow NoBrokersAvailableError to bubble to main thread

emmett9001:
This pull request fixes #629 by allowing `NoBrokersAvailable` to be raised to the main thread. This exception was caught as part of #450, but doing so introduced the issue mentioned in #629.

PR #639, opened 2016-12-29
fix deadlock: #633, when queue is full

fxsjy:
The bug can be re-generated.
1)  start a python script to produce message infinitely ;
2)  stop Kafka service
3)  then, the python script will hang forever.

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -524,6 +524,10 @@ class OwnedBroker(object):
         if self._auto_start:
             self.start()
 
+    def cleanup(self):
+        if not self.slot_available.is_set():
+            self.slot_available.set()
+
     def start(self):
         def queue_reader():
             while self.running:
@@ -534,9 +538,8 @@ class OwnedBroker(object):
                 except Exception:
                     # surface all exceptions to the main thread
                     self.producer._worker_exception = sys.exc_info()
-                    if not self.slot_available.is_set():
-                        self.slot_available.set()
                     break
+            self.cleanup()
             log.info("Worker exited for broker %s:%s", self.broker.host,
                      self.broker.port)
         log.info("Starting new produce worker for broker %s", self.broker.id)
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -539,7 +539,7 @@ class OwnedBroker(object):
                 except Exception:
                     # surface all exceptions to the main thread
                     self.producer._worker_exception = sys.exc_info()
-                    if not self.running and not self.slot_available.is_set():
+                    if not self.slot_available.is_set():
                         self.slot_available.set()
                     break
             log.info("Worker exited for broker %s:%s", self.broker.host,
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -225,6 +225,7 @@ class Producer(object):
         was triggered
         """
         # only allow one thread to be updating the producer at a time
+
         with self._update_lock:
             if self._owned_brokers is not None:
                 for owned_broker in list(self._owned_brokers.values()):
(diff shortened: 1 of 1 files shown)


pykafka/producer.py
```diff
@@ -535,6 +539,8 @@ def queue_reader():
                 except Exception:
                     # surface all exceptions to the main thread
                     self.producer._worker_exception = sys.exc_info()
+                    if not self.running and not self.slot_available.is_set():
```
emmett9001: @fxsjy why is `not self.running` necessary here?

fxsjy: @emmett9001 , thanks for your review. I read the relevant code, and also consider  then "not self.running"   condition here is unneccessary.  I will fix it soon.

emmett9001:
@fxsjy Thanks for the information you've provided. The fundamental issue you're running into seems to be that `_produce` holds `_update_lock`, which becomes a problem when the producer's internal queue is full most of the time (that is, when `max_queued_messages` is low compared to the message production rate). When `_update` is [called in `_send request`](https://github.com/Parsely/pykafka/blob/647f174d53527c3d16249290c6f328a74639cc23/pykafka/producer.py#L449), it immediately [waits](https://github.com/Parsely/pykafka/blob/647f174d53527c3d16249290c6f328a74639cc23/pykafka/producer.py#L228) to acquire `_update_lock`. Meanwhile, the main thread has spent most of its time [holding `_update_lock`](https://github.com/Parsely/pykafka/blob/647f174d53527c3d16249290c6f328a74639cc23/pykafka/producer.py#L371) and [waiting for `slot_available`](https://github.com/Parsely/pykafka/blob/647f174d53527c3d16249290c6f328a74639cc23/pykafka/producer.py#L666), since the queue is usually full. The main thread is holding `_update_lock` and waiting on `slot_available`, and the worker thread is waiting on `_update_lock` and holding `slot_available` - ie a deadlock.

The worker thread's call to `_update` obviously needs to hold `update_lock`, so I think the most appealing way to fix this issue is to stop the main thread from holding the lock during `enqueue` calls, which you've taken a stab at here. The way you've gone about it renders the lock in `_produce` mostly useless, so I'm thinking it might make sense to remove the `with self._update_lock:` line entirely. I checked out my [notes](https://github.com/Parsely/pykafka/pull/595) from when this locking behavior was added and ran some local tests with the locking removed. I'm actually not able to replicate the issue mentioned in #595 when I remove the lock in `_produce`, which makes me wish I'd taken better notes!

Please revise this pull request to simply remove `with self._update_lock:` from `_produce` - all of the other changes in the diff seem unnecessary to me (please correct me if I'm overlooking something basic). If you agree that this change fixes the issue you're seeing, we may be able to merge it in and get it tested as a dev release. Thanks again for your contributions!

emmett9001:
See also comments on #633

fxsjy:
@emmett9001 ,  I thought that,  in Python if we make multiple threads to read and write a dict (_owned_brokers):_ will cause data corrupt. But, acctually, it looks like thread-safe, which doesn't like std::map in C++.
So, I agree that the line `with self._update_lock` can be removed.

pykafka/producer.py
```diff
@@ -535,6 +534,8 @@ def queue_reader():
                 except Exception:
                     # surface all exceptions to the main thread
                     self.producer._worker_exception = sys.exc_info()
+                    if not self.slot_available.is_set():
+                        self.slot_available.set()
```
emmett9001: I don't think these two lines are necessary anymore.

emmett9001: I think I'd be ok with moving this logic into a `OwnedBroker.cleanup()` function that is called not in the exception handler, but after the `while self.running` loop. This will make it clearer why we're messing with `slot_available` here and avoid potential confusion when we revisit this code. 
```python
def cleanup(self):
    if not self.slot_available.is_set():
        self.slot_available.set()
```
and
```python
while self.running:
    ....
    except Exception:
        ...
self.cleanup()
```
If you add those two things to this pull request, it should be fine to merge.

fxsjy: @emmett9001,  thanks for you advice. I will fix it soon.

emmett9001:
Thanks for the changes @fxsjy. In my local testing, a similar deadlock can happen when the number of brokers goes below the replication factor and the producer's queues are regularly full. Since this also happens on current master, I've opened a separate issue #641 to track that - any input would be appreciated.

PR #640, opened 2017-01-03
catch IOError from socket.sendall

### LINKED ISSUE #637: Broken pipe exception
Hi,
 My pykafka version 2.4.0, and kafka version is 0.9.0.0.
 Our system got so many "Broken pipe" exception:
```
[2016-08-11 10:40:40,918][140595995432768-MainThread][ERROR]<kafka_producer_client.py:103> send exception: [Errno 32] Broken pipe
 File "/home/itest/scorpio_job/common/kafka/kafka_producer_client.py", line 96, in send
 producer.produce(message)
 File "/usr/local/lib/python2.7/site-packages/pykafka/producer.py", line 341, in produce
 File "/usr/local/lib/python2.7/site-packages/pykafka/producer.py", line 202, in _raise_worker_exceptions
 raise ex
 error: [Errno 32] Broken pipe
```

I find a same issue done here! 
I did like you say:
```
try:
    consumer.consume()
except IOError:
    consumer.stop()
    consumer.start()
```
by the way I use asynchronous because of the large num.

but here is the question like you say:
1.sometimes it success if i do this
2.sometimes it will fail all the time, no matter how many times i try this.

but if I release KafkaClient,Producer object and rebuild the KafkaClient object , 'Broken pipe' problem all gone.but in this  situation,maybe I will lose my meesage that not be send.

Would you please to deal with this 'Broken pipe Exception', in that way,we will not lose the messages to be sended in you 'get_delivery_report' or any other Solution in asynchronous.
Thank you!^_^

emmett9001:
SocketDisconnectedError is a custom exception type to pykafka and thus can never be raised from a call to _socket.sendall. This commit attempts to fix #637 by catching the IOError that can actually come from sendall().

[This commit](https://github.com/Parsely/pykafka/commit/f2ad38bed3d7da41d25ada5e3bab15aafd85b050) seems to have not actually done anything.


emmett9001:
Duplicate of #638

Issue #641, opened 2017-01-06
Producer deadlocks when queues are full


emmett9001:
Steps to reproduce:

* Run a local 3-node kafka cluster, make a topic with 10 partitions and replication factor 2
* Run a producer producing messages to this topic in a `while True` loop with `max_queued_messages=5`
* Kill two of the kafka nodes
* Observe that the producer deadlocks after repeated retries

This may be related to https://github.com/Parsely/pykafka/pull/633#issuecomment-269578948

**PyKafka version**: https://github.com/Parsely/pykafka/commit/2cabc71b7def03e6a90dfbf767e34e6a8df54bbc
**Kafka version**: 0.9.0.1

fxsjy:
@emmett9001 ,  I also reproduced the bug you described in the issue .
I found that the root cause is:  `self._cluster.update()` did not throw the exception after 3 times of retries. 
Therefore, I add two lines of code in `cluster.py` to raise the exception when retries reach the limit.
https://github.com/Parsely/pykafka/pull/642

Also, I checked the effect of my patches by making two scenarios.
1) Kill two kafka nodes
2) Kill only-one kafka node

For the first scenario,  the producer script will end with an exception raised.
For the second scenario, we can see the "Updating attempt" logs, and the producer script works well.

emmett9001:
Thanks @fxsjy. Are you testing with a replication factor of 3?

fxsjy:
@emmett9001 ,  I have tested it with replication factor of 3.
The producer worked well when I killed two nodes. And, after I killed the last node, the producer script raise an exception.

emmett9001:
Possibly related to #673

emmett9001:
I can't replicate this issue locally on current master. Thanks again @fxsjy, and please post here if this resurfaces in your tests.

PR #642, opened 2017-01-07
raise exception after max retries

fxsjy:
This may fix the issue  #641.  I repeated the steps in the issue, and found that the producer did not deadlock, but raised an exceptions at retry 3 times.

### DIFF
--- a/pykafka/cluster.py
+++ b/pykafka/cluster.py
@@ -446,5 +446,7 @@ class Cluster(object):
             except LeaderNotAvailable:
                 log.warning("LeaderNotAvailable encountered. This may be "
                             "because one or more partitions have no available replicas.")
+                if i == self._max_connection_retries - 1:
+                    raise LeaderNotAvailable
             else:
                 break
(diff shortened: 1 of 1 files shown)


pykafka/cluster.py
```diff
@@ -446,5 +446,7 @@ def update(self):
             except LeaderNotAvailable:
                 log.warning("LeaderNotAvailable encountered. This may be "
                             "because one or more partitions have no available replicas.")
+                if i == self._max_connection_retries - 1:
+                    raise LeaderNotAvailable
```
emmett9001: This could be changed to simply `raise` since we're raising an exception of the same type we're catching.

emmett9001:
I'm still able to replicate the deadlock on my local cluster. In some cases, the exception is raised correctly, but in others it isn't and the deadlock still occurs.

emmett9001:
I've managed to fix this issue with a combination of @fxsjy's change here and [this branch](https://github.com/Parsely/pykafka/pull/648), which ensures that we don't deadlock on event variables used for thread synchronization. I'm not sure if that change is a great idea, but I can't immediately see the downside. It will probably make a whole class of potential issues go away.

Issue #646, opened 2017-01-12
consumer worker thread not exiting


FuFikCZ:
Hello,

I have problem with running code below. Threads aren't stopped after calling SimpleConsumer.stop(). Problem is that fetcher [simpleconsumer.py](https://github.com/Parsely/pykafka/blob/fb9d9191bc2467ecd0b546eba2e24219b35d4d46/pykafka/simpleconsumer.py#L384-L397) is waiting for `fetch()` to finish, but it won't finish due waiting for thread event. SimpleConsumer.stop() sets `self._running = False`, but thread still waits for thread event and won't exit. Solution may be to add call of event.set() in SimpleConsumer.stop().

```python
import threading
import logging
import sys
from pykafka import KafkaClient

log = logging.getLogger('pykafka')
log.setLevel(10)

ch = logging.StreamHandler(sys.stdout)
ch.setLevel(logging.DEBUG)
formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
ch.setFormatter(formatter)

log.addHandler(ch)

client = KafkaClient(zookeeper_hosts='zookeper')
for i in range(10):
    topic = client.topics['topic'.encode('utf-8')]

    consumer = topic.get_simple_consumer(
        auto_commit_enable=False,
        num_consumer_fetchers=1,
        queued_max_messages=500
    )
    consumer.start()

    for j in range(100):
        message = consumer.consume(block=True)
        if message is None:
            break

    consumer.commit_offsets()
    consumer.stop()

print(threading.active_count())
for t in threading.enumerate():
    print(t)
```

**PyKafka version**: 2.5.0
**Kafka version**: 2.11-0.10.1.0

emmett9001:
Thanks for pointing this out @FuFikCZ. It looks at first glance like this might be something that only appears at lower `queued_max_messages` values. You're only reading 100 messages per consumer before killing them - I wonder if a lower `fetch_message_max_bytes` value would fix this issue for you. It's possible that lowering that value to less than the size of 100 messages from your topic could allow some free space in the queue, stopping this deadlock.

All of that said, this still seems very much like a real bug in pykafka that needs to be addressed.

FuFikCZ:
@emmett9001 I have tried lowering `fetch_message_max_bytes`. It won't fix the problem, but number of "zombie" threads is lower. I have a lot of messages which can vary widely in sizes, so I can't rely on it :-( 

[...]

PR #647, opened 2017-01-12
FIX: SimpleConsumer.stop() not killing worker threads

FuFikCZ:
This may fix issue #646, but i am not sure, if fix won't break anything
else.


FuFikCZ:
closed, my solution does not work :(

PR #648, opened 2017-01-12
deadlock insurance

emmett9001:
This pull request changes calls to `wait()` on some thread synchronization variables to be non-blocking. This amounts to generalized insurance against deadlock scenarios in which these variables are involved. It will theoretically have a minor negative effect on CPU usage since the thread will wake up periodically, but this effect should be negligible, especially in situations where throughput is high enough that the blocking `wait()` would never take more than 5 seconds.

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -667,7 +667,9 @@ class OwnedBroker(object):
                 if len(self.queue) >= self.producer._max_queued_messages:
                     self.slot_available.clear()
             if self.producer._block_on_queue_full:
-                self.slot_available.wait()
+                while not self.slot_available.is_set():
+                    self.producer._cluster.handler.sleep()
+                    self.slot_available.wait(5)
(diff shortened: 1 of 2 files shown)


emmett9001:
I'd love an opinion on whether this is at all sensible, or if I should instead be working to make update operations fully isolated from produce/consume operations. cc @kbourgoin

fxsjy:
Has the problem been solved competely ?

emmett9001:
I have my doubts about whether this solution actually addresses the issue here. How has this pull request performed for you with respect to deadlocks you've encountered, @fxsjy?

fxsjy:
@emmett9001 , Yes, this PR works. But, I am not sure whether the modification has side-effect in other cases.

emmett9001:
@fxsjy My tests of this pull request have not been able to replicate any unwanted side effects.

PR #652, opened 2017-02-01
Downgrade to WARNING because it's always retried.

kbourgoin:
Hello Mr @emmett9001. This doesn't really need to show up in our logs, since it's just going to get retried anyway.


emmett9001:
@kbourgoin don't warnings also trigger sentries? I actually think `INFO` is appropriate here.

kbourgoin:
I can do INFO too. I had been silencing up to WARNING for this logger, but I'll downgrade it more.

PR #659, opened 2017-02-28
Fix docs so offset refers to broker not consumer group

jeffwidman:
It says consumer group, but it appears that it's actually returning broker offsets such as the highwater mark, old message still in the log, etc. It doesn't even take a consumer group as an argument.


emmett9001:
Thanks @jeffwidman

Issue #661, opened 2017-03-14
Native consumer hits error backoff delay after topic EOF in end-to-end testing.


maparent:
Good day!
I have been testing end-to-end performance of pykafka, similar to what is done in the [EndToEndLatency](https://github.com/apache/kafka/blob/trunk/core/src/main/scala/kafka/tools/EndToEndLatency.scala) test.
I have been very surprised that using rdkafka often made the latency considerably higher. Here is my test harness:

```python
from time import time
from random import random
from os import getpid

from pykafka import KafkaClient
from pykafka.common import OffsetType

TIMEOUT = 6000
TOPIC = 'test'
KEY = 'key'

client = KafkaClient(
    zookeeper_hosts='localhost:2181',
    broker_version='0.10.2',
    socket_timeout_ms=TIMEOUT)
topic = client.topics[TOPIC]


def produce_and_consume(key, value, producer, consumer):
    producer.produce(value, key)
    while True:
        t = time()
        msg = consumer.consume()
        delta = time() - t
        if msg.partition_key != key:
            continue
        if msg.value != value:
            continue
        return delta

if __name__ == '__main__':
    num_seen = 0
    pid = getpid()
    producer = topic.get_producer(
        True,
        sync=True,
        retry_backoff_ms=10,
        linger_ms=1)
    producer.start()
    for use_lib in (True, False):
        deltas = []
        consumer = topic.get_simple_consumer(
            "a_group", use_lib,
            auto_commit_enable=False,
            auto_offset_reset=OffsetType.LATEST,
            fetch_wait_max_ms=0,
            # fetch_error_backoff_ms=1,
            consumer_timeout_ms=TIMEOUT)

        for i in range(20):
            value = "%d_%d" % (pid, i)
            delta = produce_and_consume(KEY, value, producer, consumer)
            deltas.append(delta)
        print "lib=%d, n=%d, low=%f, high=%f, avg=%f" % (
            use_lib, i, min(deltas), max(deltas),
            sum(deltas) / len(deltas))
        consumer.commit_offsets()
        consumer.stop()
    producer.stop()
```
Here are typical results:
```
lib=1, n=19, low=0.499292, high=1.004887, avg=0.526834
lib=0, n=19, low=0.000024, high=0.009366, avg=0.001291
```

This was done with kafka 0.10.2, pykafka 2.5.0 or 2.6.0.dev2, on MacOS 10.12.3.

[...]

PR #663, opened 2017-03-16
allow fetch_error_backoff, only for rdkafka use.

maparent:
In response to https://github.com/Parsely/pykafka/issues/661

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -78,6 +78,7 @@ class BalancedConsumer(object):
                  auto_commit_interval_ms=60 * 1000,
                  queued_max_messages=2000,
                  fetch_min_bytes=1,
+                 fetch_error_backoff_ms=500,
                  fetch_wait_max_ms=100,
                  offsets_channel_backoff_ms=1000,
                  offsets_commit_max_retries=5,
@@ -124,6 +125,9 @@ class BalancedConsumer(object):
             server should return for a fetch request. If insufficient data is
             available, the request will block until sufficient data is available.
         :type fetch_min_bytes: int
+        :param fetch_error_backoff_ms: *UNUSED*.
+            See :class:`pykafka.simpleconsumer.SimpleConsumer`.
+        :type fetch_error_backoff_ms: int
         :param fetch_wait_max_ms: The maximum amount of time (in milliseconds)
             that the server will block before answering a fetch request if
             there isn't sufficient data to immediately satisfy `fetch_min_bytes`.
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -64,6 +64,7 @@ class SimpleConsumer(object):
                  auto_commit_interval_ms=60 * 1000,
                  queued_max_messages=2000,
                  fetch_min_bytes=1,
+                 fetch_error_backoff_ms=500,
                  fetch_wait_max_ms=100,
                  offsets_channel_backoff_ms=1000,
                  offsets_commit_max_retries=5,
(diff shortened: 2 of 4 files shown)


pykafka/balancedconsumer.py
```diff
@@ -124,6 +125,11 @@ def __init__(self,
             server should return for a fetch request. If insufficient data is
             available, the request will block until sufficient data is available.
         :type fetch_min_bytes: int
+        :param fetch_error_backoff_ms: The amount of data (in milliseconds) that
```
emmett9001: `amount of time`?

pykafka/simpleconsumer.py
```diff
@@ -64,6 +64,7 @@ def __init__(self,
                  auto_commit_interval_ms=60 * 1000,
                  queued_max_messages=2000,
                  fetch_min_bytes=1,
+                 fetch_error_backoff_ms=500,
```
emmett9001: I believe all of the changes to this file are unnecessary, since the `SimpleConsumer` never uses `fetch_error_backoff_ms`. It's ok for that kwarg to exist only for `RdKafkaSimpleConsumer`.

pykafka/balancedconsumer.py
maparent: oupse!

pykafka/simpleconsumer.py
maparent: I did as requested... and I'm having second thoughts. 
The discrepancy in parameters makes it harder to write code that just adapts to whichever class is chosen. My test harness becomes more complicated, and I fear it's typical. 
I propose reverting that change, but making it clear in the doc that the parameter is unused in all but the base case.
Thoughts?

maparent:
Sorry, I made a reply to your comment which was lost since I try to keep my change to a single commit. Hope you don't get it twice.

I did as requested, but I have misgivings.
Not having the parameter in the other classes makes it harder to write code that adapts to whichever class is desired. (My test example, for example, is now broken.)
I would suggest reinstating the parameter in the other classes; I could add documentation to the effect that
"This parameter is provided for a uniform API, but is only used by RdKafkaSimpleConsumer"
Thoughts?

pykafka/simpleconsumer.py
emmett9001: Yes, come to think of it it's probably a better idea to keep these interfaces fully synchronized. The problem is that you end up with a bunch of parameters that don't do anything for one or the other implementation, which is a pretty major API readability issue.

maparent: OK, restored the version with arguments everywhere.

maparent:
OK. no clue why the tests are failing in python 3.5 and not in python 3.4... but I somehow doubt it's my change. I get different results on my machine. (including an extra failure: test_zk_conn_lost, but the tests that block on travis work for me.) Are some of these tests flaky? Is the timeout too short? Can it be run again?

maparent:
I just squashed my changes to trigger travis, and it now passes. I suggest again some of the tests are flaky.

pykafka/balancedconsumer.py
```diff
@@ -124,6 +125,9 @@ def __init__(self,
             server should return for a fetch request. If insufficient data is
             available, the request will block until sufficient data is available.
         :type fetch_min_bytes: int
+        :param fetch_error_backoff_ms: *UNUSED*.
```
emmett9001: @johnistan Do you have an opinion on this? On one hand, there's a benefit to keeping the interfaces to the python and rdkafka consumers in sync. On the other, it's very confusing to first-time users of this API to have a kwarg that doesn't actually do anything.

PR #666, opened 2017-03-17
Enable Python 3.6 in Travis CI

messense:


emmett9001:
I'd rather not add another entry to the test matrix. I think it would make more sense to replace `3.5` with `3.6`.

messense:
Why? Python 3.6 has some breaking changes like `json.loads` accepts bytes while Python 3.5 don't. It should cover all the Python versions in test matrix.

PR #667, opened 2017-03-22
set and use produce.offset.report

maparent:
Allow to ask for and obtain the offset in the delivery_reports.
I'm not sure how to do this in the non-rdkafka case, pointers would be appreciated.

### DIFF
--- a/pykafka/rdkafka/producer.py
+++ b/pykafka/rdkafka/producer.py
@@ -184,7 +184,7 @@ class RdKafkaProducer(Producer):
             "message.timeout.ms": 1.2 * self._max_retries * (
                 self._ack_timeout_ms + self._retry_backoff_ms),
 
-            # "produce.offset.report"
+            "produce.offset.report": self._delivery_reports.queue is not None,
             # "partitioner"  # dealt with in pykafka
             # "opaque"
         }
--- a/pykafka/rdkafka/_rd_kafkamodule.c
+++ b/pykafka/rdkafka/_rd_kafkamodule.c
@@ -553,6 +553,10 @@ Producer_delivery_report_callback(rd_kafka_t *rk,
     /* Producer_produce sent *Message as msg_opaque == rkmessage->_private */
     PyObject *message = (PyObject *)rkmessage->_private;
     PyObject *put_func = (PyObject *)opaque;
+    if (rkmessage->offset != -1) {
+        PyObject* offset = PyLong_FromUnsignedLongLong(rkmessage->offset);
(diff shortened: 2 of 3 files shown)


maparent:
Running the exact same code, which should not even be used, many times (with noop git pushes) I get different parts of the travis build succeeding or failing randomly. This is a barrier to contributions, I suggest you look into it.
Regards,
Marc-Antoine

emmett9001:
This functionality [already exists](https://github.com/Parsely/pykafka/blob/95c2e6eaf4e338dca3da0e84e8274da06a8472db/pykafka/producer.py#L427) in the pure python producer - thanks for adding it to the rdkafka implementation. Some verification that a valid offset is returned should be added [here](https://github.com/Parsely/pykafka/blob/master/tests/pykafka/test_producer.py#L106) so that we can be sure this feature doesn't break in either implementation.

I don't think an additional kwarg is necessary to toggle this feature, since it's already always on in the pure python implementation.

I'm aware that the tests are a bit flaky. It's a longstanding problem that has improved a lot over time, but there's still some work to be done. In the meantime, manually re-running the test jobs that appear to fail for reasons unrelated to the PR at hand can help show what's a real issue and what's an ephemeral failure.

Thanks for your contributions, @maparent!

pykafka/producer.py
```diff
@@ -73,6 +73,7 @@ def __init__(self,
                  max_request_size=1000012,
                  sync=False,
                  delivery_reports=False,
+                 offset_report=False,
```
emmett9001: This can be removed - I don't think it needs to be configurable.

pykafka/rdkafka/producer.py
```diff
@@ -180,7 +181,7 @@ def _mk_rdkafka_config_lists(self):
             "message.timeout.ms": 1.2 * self._max_retries * (
                 self._ack_timeout_ms + self._retry_backoff_ms),
 
-            # "produce.offset.report"
+            "produce.offset.report": self._offset_report,
```
emmett9001: This can just be `True`

maparent:
Removed the parameters, and made the reports conditional to the reports queue. Added the test as requested. Let's see what Travis has to say!

maparent:
Still failed. Is there a way for me to re-run the failing test jobs, other than a spurious push?

emmett9001:
You can rerun individual test jobs by clicking "details" next to the Travis CI check at the bottom of a pull request. There, you can click into any job and click "Restart Job" in the top right corner.

Issue #599, opened 2016-09-05
unknown function in log for librdkafka related stuff


messense:
```
2016-09-05 03:25:35,010 ERROR Process-1145 Dummy-22 pykafka.rdkafka._rd_kafka.(unknown function):0     pykafka.rdkafka#producer-2 [FAIL] ns504248.ip-192-99-101.net:9092/2: Receive failed: Disconnected
2016-09-05 03:25:35,010 ERROR Process-1145 Dummy-22 pykafka.rdkafka._rd_kafka.(unknown function):0     pykafka.rdkafka#producer-2 [ERROR] ns504248.ip-192-99-101.net:9092/2: Receive failed: Disconnected
2016-09-05 03:25:35,019 ERROR Process-1145 Dummy-20 pykafka.rdkafka._rd_kafka.(unknown function):0     pykafka.rdkafka#producer-1 [FAIL] ns504248.ip-192-99-101.net:9092/2: Receive failed: Disconnected
2016-09-05 03:25:35,020 ERROR Process-1145 Dummy-20 pykafka.rdkafka._rd_kafka.(unknown function):0     pykafka.rdkafka#producer-1 [ERROR] ns504248.ip-192-99-101.net:9092/2: Receive failed: Disconnected
```

Just out of curiousity, can it show the real function name?

emmett9001:
Hm, not sure. Is there any additional context you can provide around these errors?

messense:
I happens when I try to producer some messages and producer timed out.

amitripshtos:
happens to me a lot as well, I  guess that sudden disconnection and everything is alright, just can not understand what happened :)

johnistan:
I am also seeing these messages. I think it is related to a reaper on the kafka broker. rdkafka dicussion here: https://github.com/edenhill/librdkafka/issues/437

emmett9001:
Resolved by #668

PR #668, opened 2017-03-23
avoid disconnection error of idle connections

maparent:


emmett9001:
Thanks @maparent, this is a great find! This fixes #599.

PR #682, opened 2017-05-16
fix broken version comparison used during ssl tests

emmett9001:


emmett9001:
See the travis runs on this branch - ssl connections are currently failing with `dh key is too small`.

Issue #538, opened 2016-05-03
Error discovering offset manager


dvenza:
Using Kafka 0.9.0.1 and pykafka from pip, on python3. The topic has 2 replicas and two partitions on a two broker cluster.

I don't know if it is related to #405 or to some other problem.
When I create a consumer with this code:
```
    consumer = topic.get_simple_consumer(consumer_group='pq_consumer'.encode('utf-8'))
    cursor = pq_conn.cursor()
    for message in consumer:
        if message is not None:
            do_stuff(message)
        consumer.commit_offsets()
```

I get this error:
```
2016-05-03 11:13:28,330 INFO pykafka.cluster (MainThread): Attempting to discover offset manager for consumer group 'b'pq_consumer''
2016-05-03 11:13:28,332 ERROR pykafka.cluster (MainThread): Error discovering offset manager.
2016-05-03 11:13:28,333 DEBUG pykafka.cluster (MainThread): Retrying offset manager discovery
2016-05-03 11:13:29,035 INFO pykafka.handlers (0: pykafka.RequestHandler.worker for 192.168.45.252:9092): RequestHandler worker: exiting cleanly
2016-05-03 11:13:29,325 INFO pykafka.handlers (3: pykafka.RequestHandler.worker for 192.168.45.252:9092): RequestHandler worker: exiting cleanly
2016-05-03 11:13:30,337 ERROR pykafka.cluster (MainThread): Error discovering offset manager.
2016-05-03 11:13:30,338 DEBUG pykafka.cluster (MainThread): Retrying offset manager discovery
2016-05-03 11:13:34,345 ERROR pykafka.cluster (MainThread): Error discovering offset manager.
Traceback (most recent call last):
  File "/home/ubuntu/zoe-logger/pq-consumer.py", line 21, in <module>
    main()
  File "/home/ubuntu/zoe-logger/pq_consumer/entrypoint.py", line 87, in main
    main_loop(kafka_topic, pq_conn)
  File "/home/ubuntu/zoe-logger/pq_consumer/entrypoint.py", line 32, in main_loop
    consumer = topic.get_simple_consumer(consumer_group='pq_consumer'.encode('utf-8'), auto_commit_enable=False)
  File "/home/ubuntu/pykafka/pykafka/topic.py", line 192, in get_simple_consumer
    **kwargs)
  File "/home/ubuntu/pykafka/pykafka/simpleconsumer.py", line 182, in __init__
    self._discover_group_coordinator()
[...]

Issue #526, opened 2016-04-15
Kafka 0.10.0 Release Plan


johnistan:
The 0.10.0 release is scheduled to be cut on April 22, 2016. 

https://cwiki.apache.org/confluence/display/KAFKA/Release+Plan+0.10.0

Here are some changes we should be tracking:
- KIP-32 Add timestamp to message (committed upstream)
  - [KIP](https://cwiki.apache.org/confluence/display/KAFKA/KIP-32+-+Add+timestamps+to+Kafka+message)
  - [Issue](https://issues.apache.org/jira/browse/KAFKA-2511)
- KIP-35 - Retrieving protocol version (in flux)
  - [KIP](https://cwiki.apache.org/confluence/display/KAFKA/KIP-35+-+Retrieving+protocol+version)
  - [Issue](https://issues.apache.org/jira/browse/KAFKA-3304)

emmett9001:
Thanks for centralizing this information, @jofusa - it's really helpful.

jianbin-wei:
how about stream?

emmett9001:
There are currently no plans for pykafka to support Kafka Streams

emmett9001:
See the [`MessageSet v1`](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-Messagesets) documentation for info about message timestamps.

KIP-35 is apparently going into `0.10.1.0`

emmett9001:
Relevant to 0.10 message format support https://github.com/Parsely/pykafka/issues/644

emmett9001:
https://github.com/Parsely/pykafka/pull/685 merged, master now supports message timestamps.

PR #685, opened 2017-05-20
Set and get the message timestamp, if the broker allows it.

maparent:


emmett9001:
@maparent This looks great at first glance. I'll have to test it out, but this is a very welcome pull request. Thanks for your contribution!

Related to #526

maparent:
Thank you!
BTW, my comment was a bit of a misnomer; I can get existing timestamps, and set the protocol number so librdkafka can set them itself, but I do not yet set it in non-rdkafka situation. I'm actually working on this, but I'd be curious to have an opinion: should I set the timestamp whenever the kafka version allows it?

(What I like about this option is the ability to set a timestamp different from now(), which is unfortunately impossible with librdkafka.)

An issue with that is that the protocol `X.pack_into` methods are not aware of the Producer, and hence of the library version. I'm thinking of adding a "protocol_version" argument to the message constructor (would default to 0).  Another option is to have a default value of None for the timestamp, and the producer would have to explicitly set `timestamp=utcnow()` if it wants to use protocol v1, but that seems more ad hoc. (Final option is to be introduce awareness of producer at the pack_into level, but that is a very heavy-handed change.)
Thoughts, preferences?

emmett9001:
@maparent I think adding `protocol_version` to `Message.__init__` is the cleanest way to handle this.

maparent:
Good. Do you want me to do this in this PR, or as a separate one? (Might not be today either way, but would be within a week at the most.)

emmett9001:
Doing it in this PR makes sense. Once that's in, I'll give this a detailed review and leave comments, if any.

emmett9001:
These changes look great so far - once this functionality has been added to the pure-python `Producer`, we should be able to get this merged.

maparent:
I've given it a go. Curious to see travis results.
I should add new tests, would like your opinion on what you think is needed.

maparent:
OK, many tests fail, I'll give it another look later today.

pykafka/protocol.py
```diff
@@ -218,21 +236,51 @@ def pack_into(self, buff, offset):
...
 
+    @property
+    def timestamp_d(self):
+        """Get the timestamp as a datetime, if valid"""
+        if self.timestamp not in (0, -1):
+            # Assuming a unix epoch
+            return datetime.utcfromtimestamp(self.timestamp / 1000.0)
+
+    @timestamp_d.setter
+    def timestamp_d(self, dt):
```
emmett9001: I think this could use a slightly more descriptive name, maybe just `timestamp_dt` would be more widely understandable.

pykafka/protocol.py
```diff
@@ -218,21 +236,51 @@ def pack_into(self, buff, offset):
...
         struct.pack_into(fmt, buff, offset + 4, *args)
         fmt_size = struct.calcsize(fmt)
         data = buffer(buff[(offset + 4):(offset + 4 + fmt_size)])
         crc = crc32(data) & 0xffffffff
         struct.pack_into('!I', buff, offset, crc)
 
+    @property
+    def timestamp_d(self):
+        """Get the timestamp as a datetime, if valid"""
+        if self.timestamp not in (0, -1):
```
emmett9001: This check could be changed to ensure that `self.timestamp` is greater than zero.

pykafka/protocol.py
```diff
@@ -218,21 +236,51 @@ def pack_into(self, buff, offset):
...
+            # Assuming a unix epoch
+            return datetime.utcfromtimestamp(self.timestamp / 1000.0)
+
+    @timestamp_d.setter
+    def timestamp_d(self, dt):
+        """Set the timestamp from a datetime object"""
+        self.timestamp = int(
+            1000 * (dt - datetime(1970, 1, 1)).total_seconds())
```
emmett9001: This logic can be simplified to the following:
```python
self.timestamp = time.mktime(dt.timetuple())
```

pykafka/producer.py
```diff
@@ -300,7 +303,7 @@ def stop_owned_brokers():
                 for queue_reader in queue_readers:
                     queue_reader.join()
 
-    def produce(self, message, partition_key=None):
+    def produce(self, message, partition_key=None, timestamp=None):
```
maparent: Sorry, I'd like to push for this: it's the reason I'm bothering with the whole thing at all. I want to migrate legacy data (including history) to kafka, so I need to be able to recreate the kafka queue as it would have been created at a certain time in the past.

pykafka/protocol.py
maparent: I tried that, and we lose the milliseconds, which are not included in the timetuple.

maparent:
I'll integrate changes when I reattack the test failures, possibly tomorrow, Thursday at latest. Thank you for the initial review.

pykafka/producer.py
emmett9001: Sounds good, I'm fine with adding this.

maparent:
Found and solved one stupid py3 issue, but I'm stymied; many tests that fail on jenkins work locally, while others that work on jenkins fail miserably on my local machine even on the master branch, so it's hard to see how I can make them pass on this branch.

Specifically: The `test_producer.py` tests pass locally (kafka 10.2, python 2.7 or 3.6)

The `test_consumer_lag` reliably fails with pykafka.exceptions.GroupCoordinatorNotAvailable, might be related to #538, but I did try to zap my zookeeper configuration to no avail.

When I do (on master)
py.test --pdb tests/pykafka/rdkafka/test_simple_consumer.py  -k 'test_consume_earliest or test_consume_latest'
the first test will fail with ConsumerStoppedException, and the second one will block; killing py.test leaves kafka zombies that I need to kill -9.

So... is there a way we can work on this together? 
Regards

maparent:
Ah, I found a silly error, so travis is comparatively happier. Still, what I said about having trouble running tests remains true.

maparent:
OK. I tried setting the kafka version to 0.10.1.1 in `tests/kafka_instance.py`, and `test_simple_consumer.py:test_consume_earliest` now also fails with a `GroupCoordinatorNotAvailable `, and `test_consume_latest` passes (no kafka zombies.)
Which means I could run the tests (python2.7), and compare master to my branch.

Many still fail that should not (below), but here are the failures specific to my branch:
```
test_simpleconsumer.py::TestSimpleConsumer::test_offset_resume
test_simpleconsumer.py::TestGEventSimpleConsumer::test_offset_resume
rdkafka/test_simple_consumer.py::TestRdKafkaSimpleConsumer::test_offset_commit_agrees
```
and, most important, `test_consume_latest` now hangs.
I'll work on those, but here are tests that fail for me on master, and which I cannot check:

```
test_balancedconsumer.py::BalancedConsumerIntegrationTests::test_consume_earliest
test_balancedconsumer.py::BalancedConsumerIntegrationTests::test_zk_conn_lost
test_balancedconsumer.py::BalancedConsumerGEventIntegrationTests::test_consume_earliest <- tests/pykafka/__init__.py
test_balancedconsumer.py::BalancedConsumerGEventIntegrationTests::test_zk_conn_lost <- tests/pykafka/__init__.py
test_producer.py::TestGEventProducer::test_async_produce
test_simpleconsumer.py::TestSimpleConsumer::test_consumer_lag
test_simpleconsumer.py::TestSimpleConsumer::test_offset_commit
test_simpleconsumer.py::TestGEventSimpleConsumer::test_consumer_lag
test_simpleconsumer.py::TestGEventSimpleConsumer::test_offset_commit
test_simpleconsumer.py::TestGEventSimpleConsumer::test_update_cluster
test_ssl.py::SslIntegrationTests::test_ca_only
test_ssl.py::SslIntegrationTests::test_client_cert
test_ssl.py::SslIntegrationTests::test_legacy_wrap_socket
rdkafka/test_simple_consumer.py::TestRdKafkaSimpleConsumer::test_consumer_lag <- tests/pykafka/test_simpleconsumer.py
rdkafka/test_simple_consumer.py::TestRdKafkaSimpleConsumer::test_offset_commit <- tests/pykafka/test_simpleconsumer.py
rdkafka/test_simple_consumer.py::RdkBalancedConsumerIntegrationTests::test_consume_earliest <- tests/pykafka/__init__.py
rdkafka/test_simple_consumer.py::RdkBalancedConsumerIntegrationTests::test_zk_conn_lost <- tests/pykafka/__init__.py
rdkafka/test_ssl.py::TestRdKafkaSsl::test_ca_only <- tests/pykafka/test_ssl.py
rdkafka/test_ssl.py::TestRdKafkaSsl::test_client_cert <- tests/pykafka/test_ssl.py
rdkafka/test_ssl.py::TestRdKafkaSsl::test_legacy_wrap_socket <- tests/pykafka/test_ssl.py
```

maparent:
OK. It seems that ``get_group_coordinator`` can take up to 40 seconds to stabilize; I just hiked the retry count in that specific case. For some reason, this success resolves the zombie process issue downstream. As for the tests which I said failed in the branch and not in master, I saw test_offset_resume fail or succeed in very arbitrary fashion on either branch. Again, it seems the issue is more with the testing harness than my code. But I'm curious if there are further ways to confirm that.

emmett9001:
@maparent Thanks for your effort around testing this change. Local testing is definitely spotty for pykafka, which is why my workflow for smaller changes like this usually involves pushing a branch and opening a PR to trigger a Travis build. It's rare that I run the full test suite locally. That said, I just attempted it and a lot of the tests did actually work for me. I'm getting a hang in master on `tests/pykafka/test_producer.py::ProducerIntegrationTests::test_async_produce_large_message`, which interestingly enough is the only consistently failing test I currently see in the Travis build for this branch.

One thing that might make your local testing run a bit more smoothly is to run them against an existing Kafka cluster instead of having them set up and tear down their own clusters. [Here](https://gist.github.com/emmett9001/8c44968f9d69ba697c2a339b8a872913) is the script I use to run my local Kafka cluster for testing - you can run it with `source kafka-mux 0.10.1.0`, assuming you have a Kafka 0.10.1.0 installation at `/opt/kafka-0.10.1.0`.

Ill dig into the `test_async_produce_large_message` failure and see if I can fix it.

emmett9001:
In general I tend to assume that the test harness is shaky (this is something we could really use community help with improving, by the way). I'll often end up re-running Travis builds several times until all jobs have passed, because many of the tests can fail occasionally for reasons unrelated to the changeset in a given PR.

maparent:
Thank you for this, I'll try to make it work in my environment.
I'll also look at test_async_produce_compression_large_message tomorrow.

emmett9001:
This [Travis build](https://travis-ci.org/Parsely/pykafka/jobs/240974684) I just ran proves that the failure in `test_async_produce_compression_large_message` is not dependent on the tests that run before it, since it fails in the same way when run on Travis as a single test. Unfortunately, it passes locally for me, and the error it gives on Travis is `-1`, Kafka's code for "unknown server error". @maparent does this command pass locally for you?
```
tox -e py27 -- -s tests/pykafka/test_producer.py::ProducerIntegrationTests::test_async_produce_compression_large_message
```

maparent:
I'm doing it without tox:
```py.test --pdb tests/pykafka/test_producer.py -k test_async_produce_compression_large_message```
and it does fail.

File: pykafka/cluster.py [hunk 5]
```diff
@@ -402,23 +402,24 @@ def get_group_coordinator(self, consumer_group):
         """
         log.info("Attempting to discover offset manager for consumer group '%s'",
                  consumer_group)
-        for i in range(self._max_connection_retries):
+        max_connection_retries = self._max_connection_retries + 5
```
emmett9001: If this is a consistent issue, I think I'd prefer to raise the default on `_max_connection_retries` than to disregard its configured value here.

maparent: But _max_connection_retries is used in other methods, and I do not know that those cases need more retries. Maybe create a separate _max_offset_mgr_connection_retries value?
This was admittedly hacky code, I wanted to see how travis would react.

maparent:
Wild speculation: Error 2 is CorruptRecordException("This message has failed its CRC checksum, exceeds the valid size, or is otherwise corrupt."). I wonder if the CRC becomes invalid after the compression process? Maybe it is then recalculated by the producer in the v0 protocol, but not in a valid way for a v1 message. Or maybe the compression of v1 is simply buggy. 
Have to go for now, will look again tomorrow.

maparent:
Solved the test_async_produce_compression_large_message issue.

emmett9001:
@maparent how did you manage to replicate the error in order to fix it? Also, do you have any understanding of why this might have been happening on Travis but not locally (for me at least)?

emmett9001:
I was asking about the former. I'm pretty sure my env var was set properly, though this of course suggests that it wasn't. Either way, huge kudos for resolving the issues - the tests all pass now! I'm doing a final review of the changeset and will merge assuming it all still looks good.

emmett9001:
Looks great, @maparent. Thanks for your responsiveness and support getting this much-requested feature into master!

Issue #684, opened 2017-05-19
Make gevent dependency opt-in?


messense:
Currently `pykafka` always requires to install `gevent` even if we don't use it. Would be great be able to opt-in.

Basically, `pip install pykafka` will not install `gevent`, while we can add a `extra_requires` param in `setup.py` to make `pip install pykafka[gevent]` install `pykafka` with `gevent`.

emmett9001:
Sure, this seems reasonable. Thanks for the suggestion @messense, and if you'd like to open a PR that implements this, that would be great.

PR #686, opened 2017-05-27
Make gevent dependency opt-in

messense:
Adresses #684 

I wonder if I should add a gevent-free test matrix to ensure it wouldn't break in the future.


emmett9001:
@messense A gevent-free entry in the test matrix would be useful, I agree.

emmett9001:
This looks great - thanks for your contribution!

PR #687, opened 2017-06-01
Add test for GEvent = None in get_simple_consumer

aleatha:
get_simple_consumer needs to test if GEvent is None before checking isinstance.


emmett9001:
Thanks @aleatha!

PR #689, opened 2017-06-02
Tweak Travis CI config

messense:
1. Make exisiting matrixes test without gevent
2. Added {py27,py36}-gevent matrixes to test with gevent

~~May require #687 to build successfully.~~ I have pulled that commit in.


messense:
Ok. Still needs to change `.travis.yml` and imports of some test cases, will do that later.

emmett9001:
Thanks @messense!

PR #693, opened 2017-06-12
message timestamp tests

emmett9001:
This pull request adds tests for message timestamps in the protocol.


maparent:
hmmm. saw the large message issue with you. I'm on the trail.

maparent:
Situation: the issue is in OwnedPartition.enqueue_messages. It does not know that the topic is compressed (maybe it isn't...) and so it expects meaningful message offsets, which it is not getting in the compressed message case. I suspect the solution is to ensure that the message set  knows its messages were compressed.

maparent:
BTW from the protocol text:

> When the producer is sending compressed messages, to avoid server side recompression, each compressed message should have offset starting from 0 and increasing by one for each inner message in the compressed message.

So I added a flag to Messages so the message_offset is not taken into account in ordering, and this test passes.
This probably has more far-reaching impact. I have no clue why this test passed earlier; I suspect the change of protocol changed the distribution of messages between brokers.

maparent:
I don't expect this to pass as such, please add https://github.com/maparent/pykafka/commit/2af4662dde7e326b9ebade42116f2c318d4d24b7

maparent:
(Sorry, I have to log off for the evening. Best of luck.)

maparent:
I saw tests failed, your commit is different from mine; offsets start at 1, mine still start at 0.
BTW: I'll be off next week, so best if we can resolve this today.

emmett9001:
@maparent This change is looking good to go, so I'm merging it via this PR. Thanks again for the contribution - I'll tag you in any additional changes that need to happen around message timestamps.

maparent:
Thank you very much!

PR #694, opened 2017-06-13
Timestamp fixes

maparent:
Fixes to the initial timstamp work: Set and use the FetchRequest api_version, other corrections.


emmett9001:
@maparent Thanks for looking into this. I've [updated some tests](https://github.com/Parsely/pykafka/pull/693) on my branch. You could rebase this branch against that one to get a more accurate picture of which tests still need to be fixed, but I'm not sure whether having two branches based on each other might cause issues. It looks like aside from the test fixes I've implemented in #693, some of the tests are still failing here. I'll look into why this might be tomorrow if you don't get to it first.

pykafka/protocol.py
```diff
@@ -169,6 +171,7 @@ class Message(Message, Serializable):
         "produce_attempt",
         "delivery_report_q",
         "protocol_version",
+        "compressed_offset",
```
emmett9001: Naming: `compressed_offset` might be misconstrued as "an offset that was compressed". Consider renaming this to something like `offset_was_compressed`.

pykafka/protocol.py
```diff
@@ -212,13 +217,13 @@ def __len__(self):
 
     @classmethod
     def decode(self, buff, msg_offset=-1, partition_id=-1):
-        (crc, protocol_version) = struct_helpers.unpack_from('iB', buff, 0)
-        offset = 5
+        (crc, protocol_version, attr) = struct_helpers.unpack_from('iBB', buff, 0)
```
emmett9001: Good catch, I totally missed this yesterday and it's probably what was driving me insane.

pykafka/protocol.py
```diff
@@ -769,14 +776,28 @@ class FetchResponse(Response):
           HighwaterMarkOffset => int64
           MessageSetSize => int32
     """
-    def __init__(self, buff):
+    api_version = 0
+
+    @staticmethod
+    def get_subclass(broker_protocol):
+        target_version = parse_version(broker_protocol)
```
emmett9001: Including a link here to [this documentation](https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-FetchResponse) might help disambiguate this logic for future readers.

pykafka/protocol.py
maparent: Yes, sorry about this.

pykafka/protocol.py
maparent: Moot with my latest commit.

maparent:
OK, i did a major overhaul. With compressed messages, we cannot know what the next_offset is; it depends if the message came from a compressed message or not. I encapsulated this logic in a MessageOffset class, and replaced the next_offset by last_offset_received in OwnedPartition. I hope you agree with how I went about things, but it's a big change, and tests now pass locally.

maparent:
More changes since last messages. I created my own branch https://github.com/maparent/pykafka/tree/test/message_timestamps rebasing yours and adjusting.

maparent:
OK... good and bad news: it seems I misinterpreted the documentation's statement that comprssed offsets would start with zero. So... I have to scrap today's work and understand better why things failed.

maparent:
Well... the last comment was wrong, there was a test snafu.
But as a result, I have a better handle on offset behaviour. The compressed offsets are 0-indexed with the new protocol and not the old one.
So we have two options:
1. Correct the offset so they start at the "right" place (currently implemented on my timestamp_contd branch)
2. Use complex offset object, as I started to do earlier today. I've corrected that code and now it works for me locally; I put it on a branch here: https://github.com/maparent/pykafka/tree/MessageOffset
I am torn, and not because of chucking code. The offset correction seems to work well locally (travis will tell), but I _feel_ it's too ad hoc. It seems to cover all existing cases, though.
The MessageOffset is more complex but feels more robust. Is it really? I'm not sure. I'd like to submit both to your judgement.

emmett9001:
@maparent Just so you know, I haven't forgotten about this. I'm planning on checking it out in detail before Friday.

pykafka/protocol.py
```diff
@@ -794,17 +814,45 @@ def _unpack_message_set(self, buff, partition_id=-1):
...
             elif message.compression_type == CompressionType.SNAPPY:
                 decompressed = compression.decode_snappy(message.value)
-                output += self._unpack_message_set(decompressed,
-                                                   partition_id=partition_id)
+                messages = self._unpack_message_set(decompressed,
+                                                    partition_id=partition_id)
+            if messages[-1].offset < message.offset:
+                # With protocol 1, offsets from compressed messages start at 0
+                assert messages[0].offset == 0
+                delta = message.offset - len(messages) + 1
```
emmett9001: I don't follow this math. From my understanding, messages nested in compressed messagesets come back with their offsets starting from 0. Thus if `message.offset` here were 50, I'd expect the first item in `messages` to have its offset set to 51 - the next offset. However, assuming that `len(messages)` here is 5, for example, this math results in `delta == 50 - 5 + 1 == 46`, meaning that below, the message offsets will start from 46 instead of 0. What am I missing here?

pykafka/protocol.py
```diff
@@ -794,17 +814,45 @@ def _unpack_message_set(self, buff, partition_id=-1):
...
+                for msg in messages:
+                    msg.offset += delta
```
emmett9001: Is it safe to artificially adjust the offset values of returned messages? I assume this would result in messages apparently appearing out-of-sequence as offsets jumped around. Imagine getting messages that came from a compressed message set and seeing that their offsets start from 50 and increase. Then after the messages from that messageset have been consumed, an uncompressed message set arrives and again starts returning offsets starting at 51 and increasing. Is that how these offsets work, or am I fundamentally not understanding what's going on here?

pykafka/protocol.py
maparent: This is extremely empirical. Suppose you start at offset 5, then receive 3 compressed messages. The compressed enclosing message will have offset 8, and the enclosed messages will have offsets (6, 7, 8) in protocol 0, or (0, 1, 2) in protocol 1. So I have to to add 5, which is 8-3, to the protocol-1 offsets to get back the 6, 7, 8 offsets, which is what I would be getting with uncompressed messages. Is it clearer?

pykafka/protocol.py
maparent: Same idea: if you start at 50, your first batch of 3 messages will have enclosing offset 53, (vs 51), and the next batch of 4 messages will have enclosing offset 57, so the enclosed offsets are monotonic.

emmett9001:
Whoops, I forgot that the [other PR](https://github.com/maparent/pykafka/pull/1) I opened for discussion was against your master branch, not the main one. We can continue that conversation there, but the PR will obviously need to be reopened against this repo at some point if we decide to pursue it.

pykafka/protocol.py
emmett9001: Oh, that makes sense. My incorrect assumption was that the containing message has an offset equal to `previous_offset + 1` instead of what it apparently actually is, which is `previous_offset + len(internal_messages)`.

By the way, your help in figuring out what Kafka is actually doing here is invaluable.

emmett9001:
Ok, after your clarification on how compressed message offsets actually work, I'm on board with scrapping https://github.com/maparent/pykafka/pull/1 in favor of this PR. I like keeping it simple, and the approach used here is more or less in line with the philosophy that we've used to build 0.8+ support so far. I believe the last step to merge this is to rebase it against https://github.com/Parsely/pykafka/pull/693 to include those tests and make sure they pass. Once we have that, I don't believe there are any additional blockers. Thanks again.

maparent:
I have done this rebase in https://github.com/maparent/pykafka/tree/test/message_timestamps

emmett9001:
I think it makes sense to combine both `maparent:timestamp_contd` and `test/message_timestamps` in this PR so we can have everything in one place. I also just now made a few updates to `test/message_timestamps`, so with another rebase of this branch we should be able to see how Travis handles things.

emmett9001:
I considered it and realized this PR setup will work fine. Travis is currently running on the latest in #693, which should include the current state of both that branch and this one.

maparent:
I added my rebase of your tests to this branch, so we get a travis verdict. there were issues in the other branch.

PR #700, opened 2017-07-14
Simplify cli usage

erik-stephens:
Can do `python3 -m pykafka.cli ...` without having to know where pykafka
installed to.

Before submitting with updated docs, wanted to gauge interest.  Not sure how popular this pattern is with the general community.


emmett9001:
Looks great to me! Thanks @erik-stephens.

PR #702, opened 2017-07-17
raise SocketDisconnectedError with host and port args

zhipcui:
when raise SocketDisconnectedError without host and port , we have no idea which broker has network issue.


emmett9001:
Thanks @zhipcui!

Issue #664, opened 2017-03-16
List Consumer Groups for Topic


johnistan:
Now that consumer groups and offsets are managed by kafka I have an need for an admin tool to report all existing consumers on a given topic.

I know the java cli tools has this command:

`bin/kafka-consumer-groups.sh  --list --bootstrap-server localhost:9092`

I don't see the corresponding command in pykafka. Am I missing a call or is this an enhancement thats needs a PR?

emmett9001:
The [kafka_tools](https://github.com/Parsely/pykafka/blob/master/pykafka/cli/kafka_tools.py) module has never been particularly feature-complete. A command that lists all existing consumer groups for a topic would be a welcome pull request.

emmett9001:
This CLI feature was generously contributed by @tanaysoni in #706.

PR #706, opened 2017-08-10
added cli to print consumer groups for a topic

tanaysoni:
Issue: #664
Added a command in `kafka-tools` to list consumer groups for a topic.

### DIFF
--- a/pykafka/cli/kafka_tools.py
+++ b/pykafka/cli/kafka_tools.py
@@ -116,6 +116,36 @@ def desc_topic(client, args):
     ))
 
 
+def print_consumer_groups(client, args):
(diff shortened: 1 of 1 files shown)


emmett9001:
Thanks for the contribution, @tanaysoni! The change itself looks great. I'm curious whether it worked as intended for you, though. When I ran it locally, I discovered a bug in the `ManagedBalancedConsumer` protocol implementation causing the topic name to be set incorrectly. I've fixed the bug in https://github.com/Parsely/pykafka/pull/707, and this change now works for me. It would be great if you could confirm whether or not you saw similar behavior in your own tests. Either way, thanks for contributing!

PR #712, opened 2017-08-24
Raise worker exception in RdKafkaSimpleConsumer

### LINKED ISSUE #704: RdKafkaSimpleConsumer ignores worker exceptions
I'm using RdKafkaSimpleConsumer with auto_commit enabled.
At some point autocommit worker thread receives some exception and dies.
Before exiting it sets:
`self._worker_exception = sys.exc_info()` 

This worker exception never reaches main thread and it continues to work like nothing happened, since nowhere self._raise_worker_exceptions() is called.

In difference with SimpleConsumer which calls self._raise_worker_exceptions() from consume() function, the overridden consume() function in RdKafkaSimpleConsumer does not raise worker exceptions.

Can be reproduced by manually raising some exception in autocommit worker thread.
Also simulated exception on SimpleConsumer, and it works as expected - consume() exits.

pykafka==2.6.0

prateek1306:
Fixes #704

### DIFF
--- a/pykafka/rdkafka/simple_consumer.py
+++ b/pykafka/rdkafka/simple_consumer.py
@@ -1,6 +1,7 @@
 from contextlib import contextmanager
 import logging
 from pkg_resources import parse_version
+import sys
 import time
 
 from pykafka.exceptions import RdKafkaStoppedException, ConsumerStoppedException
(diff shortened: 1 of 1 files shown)


prateek1306:
@emmett9001 Could you please review? Especially capturing the exception in `poll`. I am not sure if thats an oversight or intended.

emmett9001:
@prateek1306 Thanks for the contribution. This looks correct to me, and it's useful to have the oversight fixed. As for capturing the exception in `poll()`, I'm not exactly sure under what circumstances `rdk_handle.poll` will actually raise a relevant exception, so it's hard to say what this will do in practice. That said, catching and reraising exceptions that happen on worker threads is something we try to do across the entirety of the pykafka codebase, so I'm happy to add this catchall.

Issue #671, opened 2017-03-27
Rebalancing not work properly when Monkey patching


marcofvera:
**PyKafka version**:  2.5.0
**Kafka version**: 0.10.2

PyKafka can't balancing consumers when using `gevent` and `patch_all()` function.

Looking further it seems there is an issue when pykafka tries to acquire the lock to rebalance on `_rebalance` function in `BalancedConsumer` class.

Here is a code example:
```python
from gevent import monkey
monkey.patch_all()

import logging
logging.basicConfig(level=logging.INFO)

import gevent

from pykafka import KafkaClient
from pykafka.common import OffsetType


def consume(consumer):
    while True:
        message = consumer.consume()
        print('Message: %s' % message.value)


if __name__ == '__main__':
    client = KafkaClient(hosts='localhost:9092')
    topic = client.topics['test']

    consumer = topic.get_balanced_consumer(
        consumer_group='consumer',
        auto_commit_enable=True,
        auto_commit_interval_ms=1000,
        zookeeper_connect='localhost:2181',
        reset_offset_on_start=False,
        auto_offset_reset=OffsetType.LATEST,
        rebalance_max_retries=10
    )

    jobs = [gevent.spawn(consume, consumer)]
    gevent.joinall(jobs)
```
And here is my pip freeze output:
```
gevent==1.2.1
greenlet==0.4.12
kazoo==2.2.1
pykafka==2.5.0
six==1.10.0
tabulate==0.7.7
wsgiref==0.1.2
```

emmett9001:
I've managed to reproduce this bug locally. Adding `use_greenlets=True` to the `KafkaClient.__init__` call doesn't fix the problem.

marcofvera:
I'm sorry @emmett9001 I forgot to mention that I have already tried it.. 😞

marcofvera:
I added one by one all the "patches" and `monkey.patch_thread()` is the one that reproduce this bug.

tanaysoni:
@emmett9001 This seems to now work on my local after the merge of https://github.com/Parsely/pykafka/pull/718.

emmett9001:
@tanaysoni Yup, it works locally for me too. I had no idea these two issues were related - your fix was clearly very sorely needed!

carsonip:
It works for me too. Thanks.

marcofvera:
Thanks, guys! I'll try it out.

Issue #628, opened 2016-12-12
100% CPU when a consumer does not have any partitions to consume


amitripshtos:
Hey guys!
I found a weird bug and it's really easy to simulate it:
1. Create a topic with 1 partition
2. Open 3 python shells and use BalanceConsumer with the same group name for all shells
3. Start consuming until the .consume() is blocking
4. htop and see you CPUs dying

It's a problem since I'm not always aware of how much consumers I have and currently I have 100% CPUs on my consumers when they are just too much.

Thank you for your help! 

**PyKafka version**: 2.5.0
**Kafka version**: 0.9.1

emmett9001:
Hi @amitripshtos. It sounds like your program might be hitting [this line](https://github.com/Parsely/pykafka/blob/a9dd498392a2f78682719f44c56229cfff486c30/pykafka/simpleconsumer.py#L441) repeatedly and chewing up your CPU, but it's hard to be sure based on the information provided. What exactly do you mean by "until the .consume() is blocking"? Can you share some code that shows how you're calling consume? Also, does the topic you're consuming have any messages in it, or is it empty as part of the test?

amitripshtos:
Thanks for replying!
I'll explain better:
1. Create topic with one partition - no need for messages
2. Open 3 Python shells
3 create a balance consumers in each shell and start consuming

You will see that first shell CPU is fine, but two other having 100 precent CPU

That's the problem - consumers with no partitions assigned are killing the CPU

amitripshtos:
I tried playing with timeout settings and even in blocked=false it's happening

emmett9001:
Thanks for the clarification - I'll try to replicate this.

emmett9001:
I've replicated this issue locally, and it appears to be happening because `BalancedConsumer.consume()` [busywaits](https://github.com/Parsely/pykafka/blob/3ea3a6b585b6bd1fd8472ff3947eaa8a6fe3f962/pykafka/balancedconsumer.py#L723) when its internal `SimpleConsumer` doesn't return any messages and `block=True`. We can probably take advantage of the `SimpleConsumer`'s internal use of a semaphore to avoid busywaiting.

emmett9001:
@amitripshtos Please give #681 a try and see if it fixes the issue for you.

amitripshtos:
It's working, thanks!

PR #716, opened 2017-08-25
Periodically release rebalancing lock in balanced consumer

### LINKED ISSUE #701: balancedconsumer will get a PartitionOwnedError
1. start a kafka server and not include any data, create a topic with 3(any number greater than one)partition.
2. create a balancedconsumer, set `consumer_timeout_ms = -1`.
3. start the balancedconsumer client, and then start it again, here are two client named clientA, clientB.
4. after rebalance_max_retries*rebalance_backoff_ms ms, clientB will get a `PartitionOwnedError`.
5. i have debug the code, find the reason is `consume()` function will always acquire the `_rebalancing_lock` and not to release it, so `_rebalance()` will block at `with self._rebalancing_lock`.

it occur in both master and dev version.

prateek1306:
Fixes #701

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -733,6 +733,14 @@ class BalancedConsumer(object):
                 return False
             disp = (time.time() - self._last_message_time) * 1000.0
             return disp > self._consumer_timeout_ms
+
+        def non_blocking_message():
+            """Get one non blocking message from the consumer"""
+            # acquire the lock to ensure that we don't start trying to consume from
+            # a _consumer that might soon be replaced by an in-progress rebalance
+            with self._rebalancing_lock:
+                return self._consumer.consume(block=False)
+
         message = None
         self._last_message_time = time.time()
         while message is None and not consumer_timed_out():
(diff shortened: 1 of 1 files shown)


prateek1306:
This could possibly work for #671

pykafka/balancedconsumer.py
```diff
@@ -741,10 +749,11 @@ def consumer_timed_out():
                 self._raise_worker_exceptions()
                 self._internal_consumer_running.wait(self._consumer_timeout_ms / 1000)
             try:
-                # acquire the lock to ensure that we don't start trying to consume from
-                # a _consumer that might soon be replaced by an in-progress rebalance
-                with self._rebalancing_lock:
-                    message = self._consumer.consume(block=block)
+                if block:
+                    while not message:
```
emmett9001: I see what you're going for here, but if my read is correct, this amounts to busywaiting, which is something we want to avoid. This has been addressed in previous tickets https://github.com/Parsely/pykafka/issues/628, and I'd like to avoid reintroducing a condition under which it's possible to busywait in `consume()`.

emmett9001: A way to fix this issue without incurring high-CPU waiting could be to have the balanced consumer adjust `self._consumer._messages_arrived`, since it's the semaphore that's [used](https://github.com/Parsely/pykafka/blob/5d222bdcbdad7b23e0d252f9b4f2eeb23dd89817/pykafka/simpleconsumer.py#L440) to control how the `consume()` call blocks. I believe if we were to release that semaphore like we do [here](https://github.com/Parsely/pykafka/blob/5d222bdcbdad7b23e0d252f9b4f2eeb23dd89817/pykafka/simpleconsumer.py#L952) and add some logic to `SimpleConsumer.consume()` to let it know that the semaphore was released without a message actually having arrived, we'd fix the same issue you're attempting to with this PR.
My only concern is over how to do this in a way that maintains the sanity of the interfaces through which these two consumer types communicate with each other. Adjusting private members for purposes other than their intended uses is a road I'd rather not go down.

prateek1306: Right. One approach could be to have a timeout of 5s even for `block=True`
Anyways, this looks like a duplicate of #718

PR #718, opened 2017-09-01
added a new `rebalancing_in_progress_lock`

tanaysoni:
This new lock will revoke a blocking consumer when a rebalance is triggered.

### DIFF
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -754,6 +754,12 @@ class BalancedConsumer(object):
                 # a _consumer that might soon be replaced by an in-progress rebalance
                 with self._rebalancing_lock:
                     message = self._consumer.consume(block=block, unblock_event=self._rebalancing_in_progress)
+
+                # If Gevent is used, waiting to acquire _rebalancing lock introduces a race condition.
+                # This sleep would ensure that the _rebalance method acquires the _rebalancing_lock
+                # Issue: https://github.com/Parsely/pykafka/issues/671
+                if self._rebalancing_in_progress.is_set():
+                    time.sleep(.1)
             except (ConsumerStoppedException, AttributeError):
                 if not self._running:
                     raise ConsumerStoppedException
--- a/pykafka/balancedconsumer.py
+++ b/pykafka/balancedconsumer.py
@@ -235,8 +235,8 @@ class BalancedConsumer(object):
             raise ImportError("use_rdkafka cannot be used with gevent")
         self._use_rdkafka = rdkafka and use_rdkafka
 
-        self._rebalance_lock = cluster.handler.Lock()
-        self._rebalancing_in_progress_lock = cluster.handler.Lock()
+        self._rebalancing_lock = cluster.handler.Lock()
+        self._rebalancing_in_progress = self._cluster.handler.Event()
         self._internal_consumer_running = self._cluster.handler.Event()
         self._consumer = None
         self._consumer_id = get_bytes("{hostname}:{uuid}".format(
@@ -320,7 +320,7 @@ class BalancedConsumer(object):
         This method should be called as part of a graceful shutdown process.
         """
         log.debug("Stopping {}".format(self))
-        with self._rebalance_lock:
+        with self._rebalancing_lock:
             # We acquire the lock in order to prevent a race condition where a
             # rebalance that is already underway might re-register the zk
             # nodes that we remove here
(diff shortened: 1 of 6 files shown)


tanaysoni:
@emmett9001 I understand that the two locks are trying to achieve the same goal, but I couldn't figure out any way to leverage the existing `rebalancing_lock` to release a blocking consumer, when rebalance triggers. 

I have tested it on my local and it works fine when the `consumer_timeout_ms` is -1. However, a consumer will not rebalance if the timeout is larger than how long rebalance retries happen.

emmett9001:
Related to #701

emmett9001:
Thanks for this contribution, @tanaysoni. I think the fundamental thing you're getting at here, which is quite useful, is that `SimpleConsumer.consume()` needs some knowledge of whether a rebuild has been triggered between when it starts executing and when it finishes. The use of a lock for this, however, doesn't make a lot of sense to me. Since there's only one place in this code where you're acquiring your new lock, you're effectively using it as a boolean flag instead of a lock. For this case, I think a `self._cluster.handler.Event` is more appropriate. Instead of acquiring the lock, you can `set()` the event. `consume()` can check `my_event.is_set()` and return early if `True`.

tanaysoni:
Thank you for reviewing, @emmett9001. It does makes sense to use an `Event` rather than a `Lock`.

pykafka/simpleconsumer.py
```diff
@@ -420,7 +420,7 @@ def __iter__(self):
                 return
             yield message
 
-    def consume(self, block=True):
+    def consume(self, block=True, rebalance_event=None):
```
emmett9001: Using the name `rebalance` in the `SimpleConsumer` seems confusing to me, since the `SimpleConsumer` doesn't and shouldn't care that it's being used by a `BalancedConsumer`. A better name for this variable could be `unblock_event` - an event that, when set, causes `consume()` to return as soon as possible.

pykafka/balancedconsumer.py
```diff
@@ -741,10 +750,11 @@ def consumer_timed_out():
                 self._raise_worker_exceptions()
                 self._internal_consumer_running.wait(self._consumer_timeout_ms / 1000)
             try:
-                # acquire the lock to ensure that we don't start trying to consume from
-                # a _consumer that might soon be replaced by an in-progress rebalance
-                with self._rebalancing_lock:
-                    message = self._consumer.consume(block=block)
+                if not self._rebalancing_in_progress.is_set():
```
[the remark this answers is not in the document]
emmett9001: In my local tests, removing this line also causes [this test](https://github.com/Parsely/pykafka/blob/8fb6b47f522096a4c1256ea5e1cbfe17a1df3193/tests/pykafka/test_balancedconsumer.py#L67) to pass (it's currently failing in the [Travis build](https://travis-ci.org/Parsely/pykafka/builds/272368460?utm_source=github_status&utm_medium=notification)).

tanaysoni:
After making the changes,  the `test_consume_graceful_stop` is passing on my local.

As you suggested, I have added a test case for the `simpleconsumer`, but not sure if the similar test approach would work in the case of a `balancedconsumer`. The `consume()` call of a `balancedconsumer` does not returns when a `_rebalance()` is triggered.

emmett9001:
I like the way you've set up the new test. I think the other test we need is one that uses a `BalancedConsumer` to verify that the issue mentioned in #701 doesn't happen. Essentially, we need to check that two `BalancedConsumer`s running in parallel consuming the same topic do not encounter `PartitionOwnedError` when their timeouts are `-1`. We can do this with a threading approach similar to what you're doing in the `SimpleConsumer` test.

Also, take a look at the [Travis build](https://travis-ci.org/Parsely/pykafka/builds/272980194?utm_source=github_status&utm_medium=notification). There are still a few failing tests in `RdkBalancedConsumerIntegrationTests` that need to pass before this can be merged.

Thanks again for your contribution - this is a much-needed and appreciated fix!

tanaysoni:
@emmett9001 I'm assuming the individual test cases in `BalancedConsumerIntegrationTests` should be isolated, i.e., their order of execution should not matter?  I observed that `test_rebalance_unblock_event` fails if executed after `test_consume_latest`, both locally and on Travis.

On [reloading the class](https://github.com/Parsely/pykafka/pull/718/commits/01a4f196d4cfb827141f974a5a666f42e2230d05) before executing test, `test_rebalance_unblock_event` always passes irrespective of execution sequence. Possibly, something malicious is happening in `test_consume_latest`?

tests/pykafka/test_balancedconsumer.py
```diff
@@ -253,8 +253,8 @@ def test_rebalance_unblock_event(self):
         consumer_b_thread = threading.Thread(target=consumer_b.consume)
         consumer_b_thread.start()
 
-        consumer_a_thread.join(10)
-        consumer_b_thread.join(10)
+        consumer_a_thread.join(30)
```
tanaysoni: @emmett9001 What timeout would you consider appropriate for thread join? I kept it sufficiently large to be on the safer side, but at the cost of increase Travis build times.

emmett9001: I'd try to keep it as short as possible, 10s is probably a good starting point.

emmett9001:
@tanaysoni Yes, it's possible that there is some issue with tests affecting each other. The most straightforward workaround to avoid getting bogged down on this particular pull request is to put your test before `test_consume_latest`. We can figure out why the tests are affecting each other in a separate PR.

tests/pykafka/test_simpleconsumer.py
```diff
@@ -76,6 +77,29 @@ def test_consume(self):
...
+        consumer = topic.get_simple_consumer(use_rdkafka=self.USE_RDKAFKA, consumer_timeout_ms=-1)
+
+        # Consume all messages in the topic
+        count = 0
+        for _ in consumer:
+            count += 1
+            if count == self.total_msgs:
+                break
+
+        unblock_event = threading.Event()
```
emmett9001: For consistency, this should be `consumer._cluster.handler.Event()`

emmett9001:
@tanaysoni This is looking great now. Is there anything else you want to add before merge? I left a comment, but it's something minor that can be fixed without needing to re-run the tests. Thanks again!

tanaysoni:
@emmett9001 Thank you for reviewing. I think this is good to go!

emmett9001:
@tanaysoni Thanks for the contribution, I can't wait to try this out in production and see if it makes a difference for our issues at Parse.ly.

PR #723, opened 2017-09-13
raise when there is only one type of error present in fetch_offsets

### LINKED ISSUE #720: simpleconsumer does not raise errors in fetch_offsets.
simpleconsumer is not correctly raising KafkaException in the fetch_offsets method. This is due to the following test: 
 https://github.com/Parsely/pykafka/blob/master/pykafka/simpleconsumer.py#L597-L598
```
        if len(parts_by_error) > 1:
            raise KafkaException(parts_by_error)
```
When the same error code is returned for all partitions, such as GroupLoadInProgress.ERROR, then the length of parts_by_error will be 1, preventing the raising of KafkaException. This results in the subsequent reseting of the partition's offsets when messages are consumed.

emmett9001:
fixes #720

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -594,7 +594,7 @@ class SimpleConsumer(object):
             to_retry = [pair for err in itervalues(parts_by_error) for pair in err]
             reqs = [p.build_offset_fetch_request() for p, _ in to_retry]
 
-        if len(parts_by_error) > 1:
+        if len(parts_by_error) > 0:
             raise KafkaException(parts_by_error)
 
     def reset_offsets(self, partition_offsets=None):


pykafka/simpleconsumer.py
```diff
@@ -594,7 +594,7 @@ def _handle_success(parts):
             to_retry = [pair for err in itervalues(parts_by_error) for pair in err]
             reqs = [p.build_offset_fetch_request() for p, _ in to_retry]
 
-        if len(parts_by_error) > 1:
+        if len(parts_by_error) > 0:
             raise KafkaException(parts_by_error)
```
iciclespider: Will len(parts_by_error) ever be 0?  Isn't the only valid exit from this method the return on line 586? I was thinking this should be coded simply as the following without any len(parts_by_error) check:
```
        raise KafkaException(parts_by_error)
```

emmett9001: I believe you're right - I'll simplify this.

emmett9001: Oops, the one condition I overlooked that actually requires this check is the one tested in [`test_extra_consumer`](https://github.com/Parsely/pykafka/blob/4935c4208cbdabffbcd3e5fb7d85e58fcbe911d2/tests/pykafka/test_balancedconsumer.py#L182). In cases where there are more consumers than partitions on the topic, the "extra" consumer will fall out of this loop without ever having a successful offset fetch, but also without ever encountering a retryable error. To support this case, I believe this check needs to remain.

iciclespider: I believe that introduces two issues:

1. In the case of such an extra consumer, this test fails https://github.com/Parsely/pykafka/blob/master/pykafka/simpleconsumer.py#L589 causing such consumers to retry for self._offsets_fetch_max_retries times before finally falling out of that loop.
1. I find it confusing that here https://github.com/Parsely/pykafka/blob/master/pykafka/simpleconsumer.py#L590 `success_responses` is returned, whereas if it falls out of the loop and there are no error, then this method does not return a value, but just exits.

emmett9001:
@iciclespider These seem like valid issues, here's a PR that addresses them https://github.com/Parsely/pykafka/pull/729

Issue #709, opened 2017-08-21
lz4 Support


sghaskell:
Is support for lz4 compression on the roadmap and if so, when do you anticipate a release?

emmett9001:
Yes, pykafka should definitely support all of Kafka's supported compression methods. There is no timeline for it at the moment, but it's something I'm interested in adding when I next have some availability to write a big-ish feature. In the meantime, any community member reading this who's interested in contributing lz4 compression support should feel free to open a pull request for review.

sghaskell:
Thanks for the quick reply @emmett9001. I'll look into it and see if I can't get something working.

emmett9001:
See [`utils/compression.py`](https://github.com/Parsely/pykafka/blob/master/pykafka/utils/compression.py) for a template of how new compression formats should be added.

sghaskell:
I've got working code for producing and consuming lz4 compressed data utilizing python-lz4. @emmett9001 can you help me understand how PyKafka determines what message protocol to use? Is this configurable by the end user or determined via some other mechanism? Can you produce messages in PyKafka using protocol versions v1 or v2?

Kafka 0.8.2 and 0.9 clients had a bug with calculating the frame checksum for lz4 compressed data - see [KAFKA-1493](https://issues.apache.org/jira/browse/KAFKA-1493) . It was fixed in 0.10.0 but the behavior is to still calculate the broken checksum for any v0 produced messages for compatibility. 

See [KIP-57](https://cwiki.apache.org/confluence/display/KAFKA/KIP-57+-+Interoperable+LZ4+Framing) for more details on the issue.

It seems that every message I send from a PyKafka producer uses protocol version v0 which results in the incorrectly calculated checksum. I'd like to test with v1/v2 message format if possible. Thanks for your help.

emmett9001:
[...]

PR #735, opened 2017-10-18
lz4 Support

sghaskell:
@emmett9001 Sorry this took so long, but here is the work I've done to add support for lz4. It requires [lz4 Bindings for Python](https://pypi.python.org/pypi/lz4). The methods [encode_lz4_old_kafka](https://github.com/Parsely/pykafka/compare/master...sghaskell:feature/lz4-support?expand=1#diff-be2d15aff9900ab3baca0523701a468eR213) and [decode_lz4_old_kafka](https://github.com/Parsely/pykafka/compare/master...sghaskell:feature/lz4-support?expand=1#diff-be2d15aff9900ab3baca0523701a468eR244) have been borrowed from [kafka-python codec.py](https://github.com/dpkp/kafka-python/blob/a00f9ead161e8b05ac953b460950e42fa0e0b7d6/kafka/codec.py). They exist to work around the bug I discuss [here](https://github.com/Parsely/pykafka/issues/709#issuecomment-324156080). In order to compress without error I had to set [block_mode=1](https://github.com/Parsely/pykafka/compare/master...sghaskell:feature/lz4-support?expand=1#diff-be2d15aff9900ab3baca0523701a468eR192) for [independent blockmode](http://python-lz4.readthedocs.io/en/latest/lz4.frame.html?highlight=block_mode#lz4.frame.compress). 

There is little to no error handling and I have not added any tests for the code.

I was able to successfully produce and consume lz4 v0 and v1 messages in my limited testing. Have a look and give it a try.


emmett9001:
@sghaskell Thanks, I'll check this out in the coming days. Can you enable me to push commits to your branch so we can both contribute to this PR?

sghaskell:
@emmett9001 I just added you as a collaborator.

pykafka/utils/compression.py
```diff
@@ -29,11 +29,32 @@
 except ImportError:
     snappy = None
 
+try:
+    import lz4.frame as lz4
+except ImportError:
+    lz4 = None
+
+#try:
+#    import lz4f
```
emmett9001: Can you explain what's going on here? Why might we need `lz4f` instead of `lz4`?

sghaskell: Looking deeper into [kafka-python codec.py](https://github.com/dpkp/kafka-python/blob/a00f9ead161e8b05ac953b460950e42fa0e0b7d6/kafka/codec.py#L205-L216) it looks like it's there to support interoperable lz4 framing.

See [KIP-57](http://grokbase.com/t/kafka/dev/164sdz8cvb/kip-57-interoperable-lz4-framing)

emmett9001: What's your intention with having it commented here? Is it a placeholder? I don't see it being used anywhere other than the check for the presence of an `lz4` library.

pykafka/utils/compression.py
```diff
@@ -166,3 +187,93 @@ def _detect_xerial_stream(buff):
         header = struct.unpack('!' + _XERIAL_V1_FORMAT, bytes(buff)[:16])
         return header == _XERIAL_V1_HEADER
     return False
+
+
+if lz4:
+    encode_lz4 = lz4.compress  # pylint: disable-msg=no-member
```
emmett9001: @sghaskell The tests I wrote are passing just fine without setting the `block_mode` on this call. Am I missing something?

emmett9001:
@sghaskell I've gotten the code to a point where I think it's mergeable assuming the tests pass. I'm still working on getting Travis to run the tests (it appears to have broken somehow...) but is there anything you notice here that should be different? I've mirrored kafka-python's implementation pretty closely.

pykafka/utils/compression.py
sghaskell: I would imagine any simple tests you write will pass without setting block_mode. I found that I had to set it when producing messages due to a lack of support for dependent-block compression.

See [KIP-57](https://cwiki.apache.org/confluence/display/KAFKA/KIP-57+-+Interoperable+LZ4+Framing)

* The only flag left unsupported is dependent-block compression (LZ4 Stream API), which our implementation does not currently support.

emmett9001: I wonder how kafka-python manages this https://github.com/dpkp/kafka-python/blob/a3995b38210d3700855d5b438529a5387c095569/kafka/codec.py#L198 cc @dpkp

emmett9001: I've actually managed to get this working against both 0.8.2 and 0.10.1 without setting `block_mode`.

sghaskell: Awesome! It must have been user error on my part then. Glad to hear it works!

PR #736, opened 2017-10-20
Use longer sleep in SimpleConsumer.fetcher

zware:
This reverts commit 43ee0ec55636bfc6a84d47f1c127f3d20215c6f3, for which I could not find a rationale.

In our project, we've found that this incredibly short sleep causes high (~20%) CPU usage at idle.  There are also a couple of `sleep()` calls in simpleconsumer.py (lines 441 and 802) that don't appear to do anything but burn a bit of CPU, though removing them is not included in this PR.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -400,7 +400,7 @@ class SimpleConsumer(object):
                     if not self._running:
                         break
                     self.fetch()
-                    self._cluster.handler.sleep(.0001)
+                    self._cluster.handler.sleep(.01)
                 except ReferenceError:
                     break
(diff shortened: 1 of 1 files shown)


emmett9001:
The other `sleep()` calls you're noticing are necessary for the components to function properly when running under gevent.

zware:
@emmett9001 Thanks for the merge and the explanation!  Any plans for when the next release will be?

emmett9001:
@zware we're moving toward another round of updates in Parse.ly's production environment. I like to wait until a dev version is verified in our system before making it official, so my guess is that 2.7.0 will become officially stable within the next month.

Issue #745, opened 2017-11-21
fetch_offsets unconditionally resets the offsets


bsideup:
Hi,

In pykafka v2.6.0, call to `fetch_offsets` will populate `partition_offsets_to_reset` and consumer will attempt to reset the offsets.
This is very unpleasant side-effect because we use pykafka to calculate consumer lag and except it to operate in read-only mode.

Is it possible to add some flag to make sure that `fetch_offsets` will not reset the offsets?

emmett9001:
Thanks @bsideup, this sounds like a really useful feature for the read-only use case.

bsideup:
@emmett9001 until it's fixed, my current workaround is:
```python
def noop_reset_offsets(self, partition_offsets=None):
    return

pykafka.SimpleConsumer.reset_offsets = noop_reset_offsets
```

yes, Monkey Patching :D
Ugly, hacky, but doesn't break anything in case of a read-only usage

PR #746, opened 2017-11-27
add a consumer param to prevent offset resets while consuming

tanaysoni:
Issue: #745

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -74,7 +74,8 @@ class SimpleConsumer(object):
                  reset_offset_on_start=False,
                  compacted_topic=False,
                  generation_id=-1,
-                 consumer_id=b''):
+                 consumer_id=b'',
+                 reset_offsets_on_consume=True):
         """Create a SimpleConsumer.
 
         Settings and default values are taken from the Scala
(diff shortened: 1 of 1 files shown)


tanaysoni:
Thank you @emmett9001 for reviewing. It'd be great to leverage the existing `reset_offset_on_start` kwarg for this use case, but when the kwarg is set to `False`, the `fetch_offset()` method would be called [here](https://github.com/Parsely/pykafka/blob/60cdff36524e221b458541864c61f6dadbb066f4/pykafka/simpleconsumer.py#L259). This may lead to offset reset attempts. 

My impression is that `reset_offset_on_start` is supposed to prevent resets on immediate start of a consumer, but subsequent resets patterns are not impacted. We might potentially lose this functionality if we use this param to prevent any `offset_reset` calls.

tanaysoni:
Thank you @emmett9001 for reviewing. I agree it'd be good to leverage an existing kwarg, but wouldn't that shadow the existing functionality of `reset_offset_on_start`? My impression is that currently the param only determines if offset reset is to be done immediately upon consumer start, without impacting any subsequent offset resets. However, for the read-only consume use case, we would like to prevent any offset reset attempts throughout the lifetime of a consumer.

emmett9001:
@tanaysoni My thinking is that since `fetch_offsets` is only called once (during `start()`), that it makes sense to overload `reset_offset_on_start` to toggle the resetting behavior of `fetch_offsets`. Other calls to `fetch_offsets` made by client code  don't necessarily have to be affected by this setting. This could be accomplished with a new argument `reset=False` on `fetch_offsets`.

Issue #722, opened 2017-09-13
Support CreateTopics / DeleteTopics APIs


messense:
https://kafka.apache.org/protocol#The_Messages_CreateTopics

It would be awesome to be able to create/delete topics in pykafka.

emmett9001:
@messense I agree pykafka should support the topic creation and deletion APIs. Do you know when support for these APIs was added to kafka?

messense:
I think it was added in Kafka 0.11, a new Java AdminClient was added too.

https://cwiki.apache.org/confluence/display/KAFKA/KIP-4+-+Command+line+and+centralized+administrative+operations

nikhilbiyani:
@emmett9001 I was able to create a topic with kafka broker 0.11.0 with the following setup:

- Assuming that 'mytopic' does not exist
- Create a producer from topic in the following way:
```
client = KafkaClient(hosts='localhost:9092', broker_version="0.11.0")
topic = client.topics['mytopic']
p = topic.get_producer()
```
- Produce some message by the producer:
```
for msg in range(0,10):
        p.produce('%i message' % msg, partition_key='key/garbage', timestamp=time.time()*1000)
p.stop()
```

- The topic was automatically created and the message were properly written.

NOTE:
While running all of this, I get one warning from pykafka:
```
No handlers could be found for logger "pykafka.cluster"
```

emmett9001:
@nikhilbiyani What you're noticing is pykafka's support for automatic topic creation, which is distinct from use of the topic create/delete API. The API provides a rich interface to topic settings as they're being created and deleted, where the autocreation interface simply uses broker defaults to create a topic for which a metadata request is sent.

emmett9001:
This is blocked by #755

emmett9001:
#750 is now unblocked and needs a bit more work, but should be mergeable in the next few weeks.

PR #750, opened 2017-12-06
Topics APIs

emmett9001:
This pull request addresses #722 by implementing protocol support for the create/delete topics APIs. It also adds an interface to the `cli` module that uses these APIs.


emmett9001:
It looks like these are the first API requests pykafka is implementing that need to talk to the "controller broker". The controller broker is assigned at the cluster level, and is distinct from the "group coordinator broker", which differs per consumer group. Pykafka currently doesn't include a way to discover the controller broker, so it will need to be added. The ID of the controller broker is returned by v1-5 of the `MetadataResponse` (at the time of this writing, pykafka only supports v0.

emmett9001:
Another option for handling lack of knowledge about the controller broker is to simply send the request to all brokers in the cluster sequentially and handle any error code 41's that occur (error 41 indicates that a request requiring communication with the controller was sent to a non-controller broker).

Issue #732, opened 2017-10-10
Pure-Python component speed optimization


emmett9001:
As mentioned in his blogpost [here](http://matthewrocklin.com/blog/work/2017/10/10/kafka-python), @mrocklin noticed that there are some potential easy wins for throughput optimization in the pure python consumer and producer. We should investigate potential changes that can lead to significant increases in message processing speed.

emmett9001:
@mrocklin did you have specific ideas about improvements we could make in this area?

mrocklin:
I found that profiling gave some nice inputs on where things were slow.  I'm running the following.

```python
from pykafka import KafkaClient
kafka_client = KafkaClient(hosts='127.0.0.1:9092')

topic = kafka_client.topics[b'bar']

producer = topic.get_producer(use_rdkafka=True)

msg = b'0' * 1000  # 1kB message

for i in range(100000):
    producer.produce(msg)
```

```
python -m cProfile -o out.prof myscript.py

pip install snakeviz
snakeviz out.prof
```

About a quarter of the time is spent in `random.choice` in `partitioners.py`.  Another decent chunk is spent in protocol.py setting a timestamp.  There is another large chunk in `producer.py::produce` that is hard to observe well because it's not in function calls.  I did a tiny bit of micro-optimization and shaved off around 10-15% from the total time

```diff
diff --git a/pykafka/producer.py b/pykafka/producer.py
index b3aa5e8..2aff29a 100644
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -317,13 +317,13 @@ class Producer(object):
         :return: The :class:`pykafka.protocol.Message` instance that was
             added to the internal message queue
         """
-        if not (isinstance(partition_key, bytes) or partition_key is None):
+        if partition_key is not None and type(partition_key) is not bytes:
             raise TypeError("Producer.produce accepts a bytes object as partition_key, "
                             "but it got '%s'", type(partition_key))
-        if not (isinstance(message, bytes) or message is None):
+        if message is not None and type(message) is not bytes:
             raise TypeError("Producer.produce accepts a bytes object as message, but it "
                             "got '%s'", type(message))
-        if self._protocol_version < 1 and timestamp:
+        if timestamp is not None and self._protocol_version < 1:
             raise RuntimeError("Producer.produce got a timestamp with protocol 0")
         if not self._running:
             raise ProducerStoppedException()
```

(isinstance checks can be surprisingly painful when done in inner loops)

PR #751, opened 2017-12-06
[WIP] Producer speed enhancements

emmett9001:
This pull request addresses #732 by making some small optimizations to the producer's main code paths.


pykafka/partitioners.py
```diff
@@ -39,6 +34,20 @@ def __call__(self, partitions, key=None):
                                   ' partitioner implementation')
 
 
+class RandomPartitioner(BasePartitioner):
```
dan-blanchard: Since this isn't actually random anymore, I think a more accurate name would be better. Maybe `RoundRobinPartitioner`?

Also the docstring should be updated to indicate that this is not actually a random partition.

PR #753, opened 2017-12-10
Update simpleconsumer.py

iciclespider:
Prevent the `log.error("Error fetching offsets for topic '%s' (errors: %s)"` message from being logged `_offsets_fetch_max_retries` times when there are no offests to be fetched by this consumer.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -588,7 +588,7 @@ class SimpleConsumer(object):
 
             success_responses.extend([(op.partition.id, r)
                                       for op, r in parts_by_error.get(0, [])])
-            if len(parts_by_error) == 1 and 0 in parts_by_error:
+            if len(reqs) == 0 or (len(parts_by_error) == 1 and 0 in parts_by_error):
                 return success_responses
             log.error("Error fetching offsets for topic '%s' (errors: %s)",
                       self._topic.name,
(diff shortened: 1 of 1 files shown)


pykafka/simpleconsumer.py
```diff
@@ -600,9 +600,7 @@ def _handle_success(parts):
             to_retry = [pair for err in itervalues(parts_by_error) for pair in err]
             reqs = [p.build_offset_fetch_request() for p, _ in to_retry]
 
-        if len(parts_by_error) > 0:
-            raise KafkaException(parts_by_error)
-        return success_responses
+        raise KafkaException(parts_by_error)
```
emmett9001: This change doesn't seem safe to me. These three lines are here as a defensive measure against cases we didn't think about. Keeping these lines unchanged seems to me like it wouldn't affect your logging change.

emmett9001:
Thanks for the contribution! I've requested a small change, after which this should be good to merge.

pykafka/simpleconsumer.py
iciclespider: The only way out of the above loop is to retry issues `self._offsets_fetch_max_retries` number of times. Why would it ever be valid to continue retrying something `self._offsets_fetch_max_retries` number of times in order to finally fall out of the loop?

The code should not repeatedly retry what is a non-error condition.

What this change does is within the loop check for the valid condition of zero reqs needed, due to having no partitions assigned to this consumer, and cleaning return from within the loop.

If the code does get to the end of the maximum number of retries and falls out of the loop, then it is an error. Why else would `self._offsets_fetch_max_retries` have been attempted?

emmett9001: That's a good point, and after reading the code a bit more closely I agree with your assesment.

Issue #754, opened 2017-12-12
Support automatic api version discovery


emmett9001:
Right now, we have a bunch of information about the supported versions of various APIs hardcoded into pykafka. This information should, when possible, be dynamically generated via the process outlined [here](https://kafka.apache.org/protocol#api_versions) (the `ApiVersionsRequest`).

emmett9001:
Blocks #755

PR #756, opened 2017-12-13
API Version Discovery

emmett9001:
This pull request implements support for automatic API version discovery as outlined in #754. It requests API version information at the beginning of `Cluster.update()` and stores this information for use throughout the lifetime of the producer and consumer. This information is used by the components to decide with protocol version to use for every request they make. This happens via an `ApiVersionAware` mixin class on the base `Request` and `Response` classes that implements a `get_version_impl(api_versions)` which returns the class implementing the highest request/response version supported by both the broker and the client.

When running against Kafka >= 0.10, this removes the need for pykafka or client code to explicitly check broker version numbers when making requests anywhere beyond the initial `ApiVersionsRequest` that happens immediately upon client initialization. Previously pykafka had been hardcoding broker version numbers based on external knowledge of support for various apis.


emmett9001:
Note: When making a request, check out the broker's supported api versions for that request. Also check out pykafka's supported api versions for the request and associated response. Make the request with the highest api version supported by pykafka's request and response that is lte the broker's max version.

emmett9001:
This is ready to merge! I'm going to let it sit for a minute and give it a once-over later since I wrote it in less than 24 hours, but I'm guessing not much will change.

PR #765, opened 2018-01-11
add api_version to OffsetFetchRequest/OffsetFetchResponse

sorcky:
According to [kafka protocol description](https://kafka.apache.org/protocol#The_Messages_OffsetFetch)
OffsetFetch Request (version 2) fetches offsets for all _topics_ if topics array parameter is `Null`.
We use this functionality to get fetch offsets for all topics/partitions for particular consumer group, including offsets for unassigned/not_curently_consumed partitions.
Tested with kafka 0.11.0.


pykafka/protocol.py
```diff
@@ -1550,6 +1570,12 @@ class OffsetFetchResponse(Response):
             Metadata => string
             ErrorCode => int16
     """
+    API_KEY = 9
+
+    @classmethod
+    def get_versions(cls):
+        return {0: OffsetFetchResponse, 1: OffsetFetchResponse, 2: OffsetFetchResponse}
```
emmett9001: `OffsetFetchResponse` v2 adds a response-level `error_code` parameter that should be handled by a new subclass.

pykafka/protocol.py
```diff
@@ -1532,6 +1540,18 @@ def get_bytes(self):
         return output
 
 
+class OffsetFetchRequestV1(OffsetFetchRequest):
+    API_VERSION = 1
+
+
+class OffsetFetchRequestV2(OffsetFetchRequest):
```
emmett9001: This should inherit from `OffsetFetchRequestV1`

pykafka/protocol.py
emmett9001: This can probably be handled similarly to how it's done in the `MetadataResponse` [subclasses](https://github.com/Parsely/pykafka/blob/5b99db27107502f6be3411df672a85eca609d6cf/pykafka/protocol.py#L686)

sorcky:
Not sure whether we need class `OffsetFetchResponseV1` and/or constant `API_VERSION`.
It works without this stuff.
So it is there just for code consistency.

emmett9001:
These changes look good. Unfortunately it looks like all of the [test jobs](https://travis-ci.org/Parsely/pykafka/builds/328006037?utm_source=github_status&utm_medium=notification) are failing. Pykafka has been carrying around a peculiar [bugfix](https://github.com/Parsely/pykafka/blob/master/pykafka/protocol.py#L1516) for a while that caused version 0 of this request to never be used, instead always using v1. To maintain this behavior, [this line](https://github.com/Parsely/pykafka/blob/814263f5cbf3d4a5caf061b3faafd0d2785cfd15/pykafka/protocol.py#L2415) and [this line](https://github.com/Parsely/pykafka/blob/814263f5cbf3d4a5caf061b3faafd0d2785cfd15/pykafka/protocol.py#L2434) should both be changed to the following:
```python
9: ApiVersionsSpec(9, 0, 1),
```
I've verified that this change causes all tests to pass except for [this one](https://github.com/Parsely/pykafka/blob/814263f5cbf3d4a5caf061b3faafd0d2785cfd15/tests/pykafka/test_protocol.py#L902). To fix this test, you can print `str(msg)` [here](https://github.com/Parsely/pykafka/blob/814263f5cbf3d4a5caf061b3faafd0d2785cfd15/tests/pykafka/test_protocol.py#L905) and copy the resulting bytestring into the `assertEqual` call. The string should be broken up like the existing one with linebreaks and indentation to make the different sections clearly visible. 

The protocol tests need a new class `TestOffsetCommitFetchAPIV2` with two functions `test_offset_fetch_request` and `test_offset_fetch_response` that validate the new v2 request/response. These new tests aren't required for this PR to be merged, but if they're not added here I'll add them after merge.

Again, if you'd like to add me as a contributor to this pull request, I'd be happy to make the tests pass, but please feel free to make the changes yourself.

sorcky:
Thank you for the thorough explanations.
I've set len(topics)=-1 [here](https://github.com/Parsely/pykafka/pull/765/commits/b4f71ce9162d394379b7b64857dd5ef232b041dc#diff-981e32d9f180f47e763f019725fb8570R965) to test the V2 difference.

emmett9001:
Thanks for working through the back-and-forth to get this merged! Looks great now.

PR #799, opened 2018-05-07
stop/start the linger/flush cycle based on queue emptiness

### LINKED ISSUE #788: Producer linger_ms=0 blocks gevent
**PyKafka version**: 2.8.0dev2 (master) (but any would do)
**Kafka version**: 0.10+, 1.0+ (but any would do)

I am using Gevent handler.
To flush messages right away, I set `linger_ms=0` in producer. When there is no message, it keeps flushing in queue_reader and effectively blocks the whole program. I can imagine that the threaded version will not really "block" but a high CPU usage may be observed.

I realize that to flush when there is at least 1 message, `min_queued_messages=1` should be used. 

However, the inline comment says:

> linger_ms=0 indicates no lingering.

without indicating the consequences, which may be misleading. If there is no plan to change the flush loop, I suggest adding a note of the consequences of `linger_ms=0`.

emmett9001:
If the queue is empty after linger_ms of waiting, wait until there is at least one message in the queue to start lingering again, or until a configurable timeout. The default behavior mirrors current master.

Fixes #788

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -570,6 +570,7 @@ class OwnedBroker(object):
         self.broker = broker
         self.lock = self.producer._cluster.handler.RLock()
         self.flush_ready = self.producer._cluster.handler.Event()
+        self.has_message = self.producer._cluster.handler.Event()
         self.slot_available = self.producer._cluster.handler.Event()
         self.queue = deque()
         self.messages_pending = 0
@@ -603,6 +604,9 @@ class OwnedBroker(object):
             queue_reader, name=name)
 
     def stop(self):
+        with self.lock:
+            if not self.has_message.is_set():
+                self.has_message.set()
         self.running = False
 
     def increment_messages_pending(self, amnt):
@@ -630,6 +634,8 @@ class OwnedBroker(object):
             if len(self.queue) >= self.producer._min_queued_messages:
                 if not self.flush_ready.is_set():
                     self.flush_ready.set()
+            if not self.has_message.is_set():
+                self.has_message.set()
 
     def flush(self, linger_ms, max_request_size, release_pending=False, wait=True):
         """Pop messages from the end of the queue
@@ -650,7 +656,9 @@ class OwnedBroker(object):
         :type wait: bool
         """
         if wait:
-            self._wait_for_flush_ready(linger_ms)
+            flush_is_ready = self._wait_for_flush_ready(linger_ms)
+        if not flush_is_ready:
+            return []
         with self.lock:
             batch = []
             batch_size_in_bytes = 0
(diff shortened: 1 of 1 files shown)


emmett9001:
@carsonip Any input on this change is welcome.

carsonip:
Sorry that I didn't reply on time. The code looks fine to me but I have a few questions.

1. As what you've mentioned in the comment, the default value of `queue_empty_timeout_ms=0` is worrying. I'm not sure of the exact impact on CPU there. I'll do some experiments when I'm free.
2. I'm also not sure why anyone wants a flush when there is nothing in the queue. Any reason for this? Or are we doing this only because we need it to behave like before? (i.e. to avoid a "breaking" change)
3. I suppose for average users using the default option, they will experience a longer delay before the flush than in the old code. The 5-second wait is for sure (rather than an expected wait of 5*1000/2 ms). Do we have to document that there is a delay in case users think it is a bug / regression?

I'll try the code soon. Thanks!

emmett9001:
Thanks for the feedback - it's super helpful! Responses below:

1. The default value of 0 simply mirrors the previous behavior. If there's enough call for changing that default, I think it will make sense to do so.
2. Yes, I'm very interested in avoiding breaking changes, since pykafka is used as a core infrastructure piece in many systems. I also know that at least for workloads like @parsely's, the producer queues are in practice *never* empty, so this pull request is a complete noop for it and similar workloads. Still, as mentioned above, I'm open to changing the default if I hear enough community desire for that change. It's not clear to me how often pykafka is used for workloads similar to Parse.ly's, and how often it's used in situations where the queue can fully empty.
3. I don't know where a longer delay has been introduced. By default, the timing behavior of queue flushes should exactly match previous behavior. When the queue empty timeout is adjusted, flush time will increase only when the queue was previously completely empty. Furthermore, it increases in a predictable way relative to the user-defined setting. I don't think there needs to be regressoin documentation beyond what's added in this PR, but please tell me if you think I'm overlooking something.

carsonip:
1. Talking about the performance, do you have any code for benchmarking the throughput / CPU usage of pykafka so that we can see performance gain/regression of this PR?
3. Yes I am talking about the special case when the queue was previously completely empty. It is an edge case anyway. The need of mentioning the delay in the doc is not only about this PR. When I first picked up code from my colleagues, I was unaware of pykafka doing the batching by default, which caused a delay. In my system, the response time is important. I think a sentence in the readme mentioning this default behavior will be helpful to other devs like me.

Thanks.

emmett9001:
See [this document](http://activisiongamescience.github.io/2016/06/15/Kafka-Client-Benchmarking/) from @johnistan for ideas on how to do producer benchmarking. I'm still not completely clear on how you'd like the documentation to change - maybe it would clarify things for you to open a pull request indicating the change you'd like to see to documentation. Thanks again for your input.

carsonip:
Sure. I'll create a PR soon.

PR #809, opened 2018-05-24
Document about the linger in producer

carsonip:
Document about the linger behavior as mentioned in #799


emmett9001:
Added the same information in a more integrated way here https://github.com/Parsely/pykafka/commit/858554029830e15cfa6d15df002d1772672d8ee0

carsonip:
Looks good. Thanks!

Issue #816, opened 2018-06-12
client CPU 100% after brokers restart


lokesh-b:
**PyKafka version**: 2.5.0 (Also happens in 2.7.0)
**Kafka version**: 0.10.2.1

Steps to Reproduce:

1. Start Kafka server with multiple brokers say 5 and 15 partitions 
2. Set replication factor of a test topic to maybe 3
3. Start python client with gunicorn
4. Send messages to test topic (Maybe greater than 10 CPM)
4. Do rolling restart of brokers
5. Watch out for the CPU percentage of gunicorn workers
6. Repeat step 4. At some point you will see python CPU usage of one or more workers reach 100%

Below screenshot shows you that _produce method is in loop forever because it doesn't have the right owned brokers list.
![image](https://user-images.githubusercontent.com/5994573/41320640-6f1c216e-6e55-11e8-9e65-2e101b7dcacb.png)

Screenshot which shows the cpu usage of the worker:
![pykafka2](https://user-images.githubusercontent.com/5994573/41320699-b0198094-6e55-11e8-9bf0-10d2819e8e6d.png)

Screenshot which shows the leader ID is right but the owned_brokers dict is not updated.
![image](https://user-images.githubusercontent.com/5994573/41320778-fb89ec4e-6e55-11e8-8a91-ec3c9bdbda51.png)

Please advice on what can be done here. Thanks

emmett9001:
Thanks @lokesh-b, this looks like a real issue.

lokesh-b:
@emmett9001 Could you please review https://github.com/Parsely/pykafka/pull/819 and let me know if this can help in fixing this issue. Thanks.

PR #819, opened 2018-06-19
Issue-816 : Fix for CPU issue seen in _produce because of metadata …

lokesh-b:
…update failure

This PR should fix https://github.com/Parsely/pykafka/issues/816. 

@emmett9001  Please review and let me if this can fix the problem for me.

### DIFF
--- a/pykafka/producer.py
+++ b/pykafka/producer.py
@@ -434,17 +434,25 @@ class Producer(object):
 
     def _produce(self, message):
         """Enqueue a message for the relevant broker
-
+        Attempts to update metadata in response to missing brokers.
         :param message: Message with valid `partition_id`, ready to be sent
         :type message: `pykafka.protocol.Message`
         """
         success = False
+        retry = 0
         while not success:
             leader_id = self._topic.partitions[message.partition_id].leader.id
             if leader_id in self._owned_brokers:
                 self._owned_brokers[leader_id].enqueue(message)
                 success = True
             else:
+                retry += 1
+                if retry < 10:
+                    log.debug("Failed to enqueue produced message. Updating metdata.")
+                    self._update()
+                else:
+                    raise ProduceFailureError("Message could not be enqueued due to missing broker "
+                                              "metadata for broker {}".format(leader_id))
                 success = False
 
     def _mark_as_delivered(self, owned_broker, message_batch, req):


pykafka/producer.py
```diff
@@ -434,17 +434,27 @@ def get_delivery_report(self, block=True, timeout=None):
...
+        max_retries = self._max_retries ** 2
         while not success:
             leader_id = self._topic.partitions[message.partition_id].leader.id
             if leader_id in self._owned_brokers:
                 self._owned_brokers[leader_id].enqueue(message)
                 success = True
             else:
+                retry += 1
+                if retry % self._max_retries == 0:
+                    log.debug("Retries exceeded limit. Updating metadata again.")
```
emmett9001: Debug message should read: `"Failed to enqueue produced message. Updating metdata."`

pykafka/producer.py
```diff
@@ -434,17 +434,27 @@ def get_delivery_report(self, block=True, timeout=None):
...
+                    self._update()
+                elif retry > max_retries:
```
emmett9001: This loop doesn't need to use `max_retries` at all. It's preferable to set the number of retries here to an arbitrary number - 10 should work.

pykafka/producer.py
```diff
@@ -434,17 +434,27 @@ def get_delivery_report(self, block=True, timeout=None):
...
+        retry = 0
+        max_retries = self._max_retries ** 2
         while not success:
             leader_id = self._topic.partitions[message.partition_id].leader.id
             if leader_id in self._owned_brokers:
                 self._owned_brokers[leader_id].enqueue(message)
                 success = True
             else:
+                retry += 1
+                if retry % self._max_retries == 0:
```
emmett9001: This can call `_update` on every retry, not only on selected ones.

pykafka/producer.py
```diff
@@ -434,17 +434,27 @@ def get_delivery_report(self, block=True, timeout=None):
...
+                    log.debug("Retries exceeded limit. Updating metadata again.")
+                    self._update()
+                elif retry > max_retries:
+                    raise KafkaException("Retries exceeded max limit")
```
emmett9001: This could be more descriptive: `"Message could not be enqueued due to missing broker metadata for broker {}".format(leader_id)`

emmett9001: The exception should be a [`ProduceFailureError`](https://github.com/Parsely/pykafka/blob/d94bc46d238cd32b00dae2aead49310f455e1005/pykafka/exceptions.py#L51)

pykafka/producer.py
```diff
@@ -434,17 +434,27 @@ def get_delivery_report(self, block=True, timeout=None):
 
     def _produce(self, message):
         """Enqueue a message for the relevant broker
-
+		Incase of  'success = False':
```
emmett9001: Wording can be improved: `Attempts to update metadata in response to missing brokers.`

Issue #823, opened 2018-06-27
ReferenceError: weakly-referenced object no longer exists


pri22296:
Hi, I'm using kafka as a queue for web crawling. I noticed that some times some spiders just hang and stop crawling. I noticed the following exception from pykafka each time the spider hangs. Any help would be appreciated.

```
Exception in thread 35: pykafka.SimpleConsumer.autocommiter:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/priyam.singh/venv/local/lib/python2.7/site-packages/pykafka/simpleconsumer.py", line 387, in autocommitter
    self.cleanup()
ReferenceError: weakly-referenced object no longer exists
```

**PyKafka version**: 2.7.0

emmett9001:
Thanks for filing this report, @pri22296. I'll need a bit more example code to properly diagnose the issue. Can you post some code that shows how you're using pykafka?

From the look of it, you might be instantiating a consumer outside of a thread and then attempting to use it in a thread.

pri22296:
Hi, I'm using pykafka within scrapy framework which itself is single threaded and I'm not spawning any new threads. This is how I am instantiating the consumer
```
consumer = topic.get_balanced_consumer(consumer_timeout_ms=25,
                                       auto_commit_enable=True,
                                       consumer_group=self.consumer_group)
```

emmett9001:
You might be attempting to use the `consumer` in a context where it's already been garbage collected. This would explain the weakref error, since the autocommitter thread is trying to operate on an object that no longer exists. Is there anywhere in your code where you might be manually cleaning up the consumer instance without first calling `stop()` on it?

emmett9001:
It's also worth looking through https://github.com/Parsely/pykafka/issues/422 to see if any of the examples listed there look similar to your issue.

pri22296:
This only happens when I'm using the auto_commi optiont. If I commit manually on regular intervals, this issue does not occur. My code is supposed to run 'forever'. So I'm never cleaning up any consumer instances.

vikt0rs:
this issue is the reason we don't use the latest pykafka version :(

[...]

PR #827, opened 2018-07-05
call cleanup() in fetch instead of autocommitter

emmett9001:
This pull request addresses #823 by moving the call to `cleanup()` from the `SimpleConsumer`'s autocommitter thread to its fetcher thread. This ensures that `cleanup` is called even when `auto_commit_enable=False`. The call is also wrapped in a `ReferenceError` check to avoid the issue mentioned in #823.

### DIFF
--- a/pykafka/simpleconsumer.py
+++ b/pykafka/simpleconsumer.py
@@ -421,7 +421,6 @@ class SimpleConsumer(object):
                     # surface all exceptions to the main thread
                     self._worker_exception = sys.exc_info()
                     break
-            self.cleanup()
             log.debug("Autocommitter thread exiting")
         log.debug("Starting autocommitter thread")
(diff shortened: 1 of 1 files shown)


stuarteberg:
Even after this change, I am still seeing a `ReferenceError` when my `SimpleConsumer` is destroyed:

```pytb
ERROR [2018-11-01 10:48:03,665] weakly-referenced object no longer exists
  File "/miniforge/envs/flyem-forge/lib/python3.6/site-packages/pykafka/simpleconsumer.py", line 449, in fetcher
    self.cleanup()
ReferenceError: weakly-referenced object no longer exists
```

This happens because `SimpleConsumer.__del__()` calls `stop()`, but that doesn't synchronously wait for fetch workers to exit -- it merely sets `self._running = False` and proceeds.  When the worker `fetcher()` resumes, `self` is no longer a valid reference.

This means that `cleanup()` may never be called -- is that a problem?  If 

I can avoid it if I briefly `sleep` before

PR #844, opened 2018-07-26
[WIP] Support RecordBatch

emmett9001:
This pull request adds support for `RecordBatch` message formatting to pykafka. This includes supporting the encoding and decoding of [`varints`](https://developers.google.com/protocol-buffers/docs/encoding#varints) via the `struct_helpers` interface.

### DIFF
--- a/pykafka/protocol/message.py
+++ b/pykafka/protocol/message.py
@@ -170,6 +170,52 @@ class Message(Message, Serializable):
             raise RuntimeError()
 
 
+class Record(Message):
+    """
+    Specification::
+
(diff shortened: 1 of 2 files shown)


emmett9001:
The protocol logic for `RecordBatch` is implemented and untested. The next step is to figure out how to get Kafka to start speaking in `RecordBatch` instead of `MessageSet` so I can run some tests to work out the protocol implementation bugs. Once that's done, the next step will be to make changes to the producer and consumer to have them use `RecordBatch` when appropriate.

PR #871, opened 2018-09-20
implement __contains__ on TopicDict 

emmett9001:
This pull request fixes a bug on `TopicDict` causing `__contains__` to function differently from `__getattr__`. In python3, the following was possible before this fix:
```python
In [1]: from pykafka import KafkaClient                                               
                                                                                                                  
In [2]: client = KafkaClient(hosts="localhost:9092")
                                                            
In [3]: client.topics['testtopic_replicated']                              
Out[3]: <pykafka.topic.Topic at 0x7f81bde9ffd0 (name=b'testtopic_replicated')>
                                      
In [4]: 'testtopic_replicated' in client.topics                      
Out[4]: False                                                    
```
Using the same string with the `in` operator and during direct attribute access yielded different results. This pull request fixes that issue by performing the same string conversion in `__contains__` that's performed in `__getattr__`.

### DIFF
--- a/pykafka/cluster.py
+++ b/pykafka/cluster.py
@@ -54,6 +54,13 @@ class TopicDict(dict):
     def values(self):
         return [self[key] for key in self]
 
+    def __contains__(self, key):
+        try:
+            key = get_string(key).encode('ascii')
+        except UnicodeEncodeError:
+            return False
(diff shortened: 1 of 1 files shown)


emmett9001:
Hold off on merging this until `2.8.0` is released, which should be early next week.

PR #905, opened 2019-01-07
Print consumer lags will show more infos

JIHENGWANG:
Will show all partitions and all consumer ids that listen to the particular partitions.

### DIFF
--- a/pykafka/cli/kafka_tools.py
+++ b/pykafka/cli/kafka_tools.py
@@ -71,7 +71,7 @@ def fetch_consumer_lag(client, topic, host, consumer_group):
     current_offsets = consumer.fetch_offsets()
     consumer_id_dict = {}
     for p_id, _ in current_offsets:
-        print ("{} {}".format(p_id, kz_patitions[p_id]))
+        #print ("{} {}".format(p_id, kz_patitions[p_id]))
         if bool(kz_patitions):
             consumer_id, _ = kz_client.get('{}/{}'.format(topic_path, kz_patitions[p_id]))
         else:
--- a/pykafka/cli/kafka_tools.py
+++ b/pykafka/cli/kafka_tools.py
@@ -41,32 +41,32 @@ def fetch_offsets(client, topic, offset):
         return topic.fetch_offset_limits(offset)
 
 
-def fetch_consumer_lag(client, topic, host, consumer_group):
+def fetch_consumer_lag(client, topic, zookeeper_host, consumer_group):
     """Get raw lag data for a topic/consumer group.
 
     :param client: KafkaClient connected to the cluster.
     :type client:  :class:`pykafka.KafkaClient`
     :param topic:  Name of the topic.
     :type topic:  :class:`pykafka.topic.Topic`
-    :param host: Host of cluster
+    :param zookeeper_host: Host of zookeeper
     :param consumer_group: Name of the consumer group to fetch lag for.
     :type consumer_groups: :class:`str`
     :returns: dict of {partition_id: (latest_offset, consumer_offset)}
     """
     latest_offsets = fetch_offsets(client, topic, 'latest')
     from kazoo.client import KazooClient
-    zookeeper_host = host.replace("9092", "2181")
-    kz_client = KazooClient(hosts=zookeeper_host)
-    kz_client.start()
     try:
+        kz_client = KazooClient(hosts=zookeeper_host)
+        kz_client.start()
         topic_path = '/consumers/{}/owners/{}'.format(consumer_group, topic.name)
         kz_patitions = kz_client.get_children(topic_path)
         kz_patitions = sorted(kz_patitions)
     except:
+        kz_client = None
         kz_patitions = None
-    
+
     consumer = topic.get_simple_consumer(consumer_group=consumer_group,
-                                         auto_start=False, 
+                                         auto_start=False,
                                          reset_offset_on_fetch=False)
     current_offsets = consumer.fetch_offsets()
     consumer_id_dict = {}
@@ -77,7 +77,8 @@ def fetch_consumer_lag(client, topic, host, consumer_group):
         else:
             consumer_id = None
         consumer_id_dict[p_id] = consumer_id
-    kz_client.stop()
+    if bool(kz_client):
+        kz_client.stop()
     return {p_id: (latest_offsets[p_id].offset[0], res.offset,
             consumer_id_dict[p_id], ) for p_id, res in current_offsets}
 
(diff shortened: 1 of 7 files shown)


pykafka/cli/kafka_tools.py
```diff
@@ -54,11 +55,16 @@ def fetch_consumer_lag(client, topic, consumer_group):
...
     consumer = topic.get_simple_consumer(consumer_group=consumer_group,
-                                         auto_start=False, 
+                                         auto_start=False,
                                          reset_offset_on_fetch=False)
     current_offsets = consumer.fetch_offsets()
-    return {p_id: (latest_offsets[p_id].offset[0], res.offset)
-            for p_id, res in current_offsets}
+    consumer_id_dict = {}
+    for p_id, stat in current_offsets:
+        info = ast.literal_eval(stat[1])
```
emmett9001: Though the offset metadata that pykafka creates looks like a python dictionary, it's actually a [JSON string](https://github.com/Parsely/pykafka/blob/ede795a90ac8d3212238a33cd6ef7e2514592436/pykafka/simpleconsumer.py#L887). Thus, the correct way to decode here is `json.loads(stat[1])`.

Also, since the `stat` tuple here is a `namedtuple`, it's preferable to reference the metadata field by name instead of subscript: `stat.metadata`.

pykafka/cli/kafka_tools.py
```diff
@@ -54,11 +55,16 @@ def fetch_consumer_lag(client, topic, consumer_group):
...
+        consumer_id_dict[p_id] = [info['consumer_id'], info['hostname']]
+    return {p_id: (latest_offsets[p_id].offset[0], res.offset,
+                   consumer_id_dict[p_id][0],
+                   consumer_id_dict[p_id][1], ) for p_id, res in current_offsets}
```
emmett9001: This iteration over `current_offsets` is redundant with the one you've added above. Only one pass over `current_offsets` is necessary to retrieve the relevant info, so there should only be one `for` in this logic.

pykafka/cli/kafka_tools.py
```diff
@@ -54,11 +55,16 @@ def fetch_consumer_lag(client, topic, consumer_group):
...
-                                         auto_start=False, 
+                                         auto_start=False,
                                          reset_offset_on_fetch=False)
     current_offsets = consumer.fetch_offsets()
-    return {p_id: (latest_offsets[p_id].offset[0], res.offset)
-            for p_id, res in current_offsets}
+    consumer_id_dict = {}
+    for p_id, stat in current_offsets:
+        info = ast.literal_eval(stat[1])
+        consumer_id_dict[p_id] = [info['consumer_id'], info['hostname']]
```
emmett9001: This line works only if the offset metadata stored in Kafka was put there by pykafka. Since this tools module is intended to be generically usable with any Kafka installation, the logic here needs to tolerate cases where the metadata was not written by pykafka. If we fail at either parsing the metadata as JSON or finding the expected `consumer_id` or `hostname` fields therein, we can simply pass with the assumption that the metadata was not written by pykafka.

pykafka/cli/kafka_tools.py
JIHENGWANG: Thank you for your comment. I did some debug, and I found that stat.metadata is actually not JSON string. It's actually a byte, so json.loads(stat.metadata) will give me an error. I still need to decode the byte to string, and do json.loads. Like json.loads(stat.metadata.decode())

pykafka/cli/kafka_tools.py
JIHENGWANG: Hi, I did some changes to make it only one for loop

pykafka/cli/kafka_tools.py
```diff
@@ -54,12 +55,23 @@ def fetch_consumer_lag(client, topic, consumer_group):
...
-            for p_id, res in current_offsets}
-
+    pid_dict = {}
+    for p_id, stat in current_offsets:
+        # Check if metadata stored into current_offsets
+        consumer_id = None
+        hostname = None
+        if bool(stat.metadata):
+            info = json.loads(stat.metadata.decode())
+            consumer_id = info['consumer_id'] if 'consumer_id' in info else None
```
emmett9001: A more concise way to write this is
```python
consumer_id = info.get('consumer_id')
```

pykafka/cli/kafka_tools.py
```diff
@@ -54,12 +55,23 @@ def fetch_consumer_lag(client, topic, consumer_group):
...
+            hostname = info['hostname'] if 'hostname' in info else None
+        # Fill data in to the dict
```
emmett9001: Unnecessary comment

pykafka/cli/kafka_tools.py
```diff
@@ -54,12 +55,23 @@ def fetch_consumer_lag(client, topic, consumer_group):
...
-
+    pid_dict = {}
+    for p_id, stat in current_offsets:
+        # Check if metadata stored into current_offsets
+        consumer_id = None
+        hostname = None
+        if bool(stat.metadata):
+            info = json.loads(stat.metadata.decode())
+            consumer_id = info['consumer_id'] if 'consumer_id' in info else None
+            hostname = info['hostname'] if 'hostname' in info else None
```
emmett9001: Same comment as above

pykafka/cli/kafka_tools.py
```diff
@@ -54,12 +55,23 @@ def fetch_consumer_lag(client, topic, consumer_group):
...
     current_offsets = consumer.fetch_offsets()
-    return {p_id: (latest_offsets[p_id].offset[0], res.offset)
-            for p_id, res in current_offsets}
-
+    pid_dict = {}
+    for p_id, stat in current_offsets:
+        # Check if metadata stored into current_offsets
+        consumer_id = None
+        hostname = None
+        if bool(stat.metadata):
```
emmett9001: This `bool` cast is unnecessary since the `if` operator casts its argument to a boolean.

pykafka/cli/kafka_tools.py
```diff
@@ -54,12 +55,23 @@ def fetch_consumer_lag(client, topic, consumer_group):
...
-                                         auto_start=False, 
+                                         auto_start=False,
                                          reset_offset_on_fetch=False)
     current_offsets = consumer.fetch_offsets()
-    return {p_id: (latest_offsets[p_id].offset[0], res.offset)
-            for p_id, res in current_offsets}
-
+    pid_dict = {}
+    for p_id, stat in current_offsets:
+        # Check if metadata stored into current_offsets
```
emmett9001: Unnecessary comment

pykafka/cli/kafka_tools.py
JIHENGWANG: Fixed

pykafka/cli/kafka_tools.py
JIHENGWANG: Fixed

pykafka/cli/kafka_tools.py
JIHENGWANG: Fixed

PR #928, opened 2019-03-17
Fix tests for librdkafka 0.11.3+

carsonip:
Limit queued.max.messages.kbytes according to librdkafka cc43f4.

This will fix rdkafka tests on librdkafka 0.11.3+.
```
E       RdKafkaException: Configuration property "queued.max.messages.kbytes" value 102400000 is outside allowed range 1..2097151
```

### DIFF
--- a/pykafka/rdkafka/simple_consumer.py
+++ b/pykafka/rdkafka/simple_consumer.py
@@ -248,9 +248,10 @@ class RdKafkaSimpleConsumer(SimpleConsumer):
             # queued.max.messages.kbytes so for now we infer the implied
             # maximum (which, with default settings, is ~2GB per partition):
             "queued.min.messages": self._queued_max_messages,
-            "queued.max.messages.kbytes": str(
+            "queued.max.messages.kbytes": str(min(
                 self._queued_max_messages
-                * self._fetch_message_max_bytes // 1024),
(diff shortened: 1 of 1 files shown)


pykafka/rdkafka/simple_consumer.py
```diff
@@ -248,9 +248,10 @@ def _mk_rdkafka_config_lists(self):
             # queued.max.messages.kbytes so for now we infer the implied
             # maximum (which, with default settings, is ~2GB per partition):
             "queued.min.messages": self._queued_max_messages,
-            "queued.max.messages.kbytes": str(
+            "queued.max.messages.kbytes": str(min(
                 self._queued_max_messages
-                * self._fetch_message_max_bytes // 1024),
+                * self._fetch_message_max_bytes // 1024,
+                2097151)),  # queued.max.messages.kbytes is 1..2097151 according to librdkafka cc43f4
```
emmett9001: The information about why this number specifically matters would be better stored in a variable name than this comment:
```python
LIBRDKAFKA_MAX_ALLOWED_QUEUED_MAX_MESSAGE_BYTES = 2097151  # see librdkafka commit cc43f4
```

carsonip: Right. Magic number is bad.

PR #936, opened 2019-04-15
Improve stability in cases of new broker additions to cluster

saich:
- Handle topic creation when a partition has in-sync replicas that current broker configuration doesn't know
- Update cluster's metadata when a partition's leader is not listed in currently known brokers


saich:
@emmett9001 Appreciate if you can review and let me know your thoughts.

This fixes a couple of issues we encountered in production with producers during a maintenance activity on Kafka cluster.

emmett9001:
These changes both look reasonable to me

PR #939, opened 2019-04-26
Socket reconnection on retries of JoinGroup, SyncGroup, LeaveGroup

saich:
Today, if `JoinGroup` sees a socket disconnection, it will retry again without reconnecting the connection which results in subsequent retries all to fail.

This change will ensure that we attempt to reconnect the connection if see a disconnected connection while selecting a unique `RequestHandler` to use for this request.

Alternative approaches that can be considered:

- Reconnect automatically in `RequestHandler.request` before calling `self.shared.requests.put`
- Reconnect automatically in `BrokerConnection.request` if required

### DIFF
--- a/pykafka/broker.py
+++ b/pykafka/broker.py
@@ -294,7 +294,16 @@ class Broker(object):
             handler = RequestHandler(self._handler, conn)
             handler.start()
             self._req_handlers[connection_id] = handler
-        return self._req_handlers[connection_id]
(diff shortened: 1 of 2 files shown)


emmett9001:
Thanks for the contribution. I think your first alternate suggestion of putting this functionality in `RequestHandler` might be a better way to maintain encapsulation.

PR #946, opened 2019-05-22
fix docs - removed mention NoBrokersAvailableError

morgangt:
I removed mention about old class, `NoBrokersAvailableError` - this is very confusing


morgangt:
If everything goes quickly, I will continue to improve the documentation in the future

emmett9001:
Thanks, @morgangt!

PR #972, opened 2019-10-07
Feature SASL SCRAM support

### LINKED ISSUE #651: How to connect to IBM Bluemix Message Hub/kafka
I would like to use pykafka to connect to IBM Bluemix Message Hub with is a Kafka implementation.  There is a kafka-python example in this thread.  https://stackoverflow.com/questions/35294949/can-i-call-the-bluemix-message-hub-service-from-python/40325181#40325181

I also found this for confluent-kafka:

A new Python client

Confluent, the company founded by the creators of Kafka, has released a Python client (https://github.com/confluentinc/confluent-kafka-python). It is built as a thin bindings layer on top of librdkafka and works for both Python 2 and 3.

To use with Message Hub, simply pass these properties in when creating a Producer/Consumer:

    'ssl.ca.location': '/etc/ssl/certs/',
    'sasl.mechanisms': 'PLAIN',
    'sasl.username': '<USER>',
    'sasl.password': '<PASSWORD>',
    'security.protocol': 'sasl_ssl'

Note: The value of ‘ssl.ca.location depends’ on your operating system. For example:

    Ubuntu: /etc/ssl/certs/
    RedHat: /etc/pki/tls/cert.pem
    OS X: select system root certificates from Keychain Access and export as .pem on the filesystem.

Is there a way to connect using pykafka?  If so, how would I do it?

swenzel:
This pull request adds support for modular SASL authentication mechanisms with two mechanisms (PLAIN, SCRAM) already implemented.

closes #651


swenzel:
Ditched kafka 0.8 because there were too many issues getting it to run. It neither supports SSL nor SASL, the configs differ significantly and there seems to be an issue with its zk client. I thought it's more useful to test kafka 2 instead.

Switched to librdkafka 0.11.5 because 0.9.5 doesn't support the SCRAM mechanisms. In the end I didn't choose a 1.x version since it introduces another configuration issue that I didn't want to solve.

I went far out of scope fixing several issues with CI and the tests which took me a lot longer than actually implementing the feature.  
I've really put a lot of work into it but now is the point where I say it's enough. I'm open to refactoring certain things or to revert a few changes if you don't want them. But I won't continue fixing CI or other tests that do not belong to the code that I changed. Especially if it's for keeping python 2 or kafka 0 compatibility.

hfjn:
Could someone please just take a look at this PR? I think @swenzel has done some great work here which really pushes pykafka forward at should be a high priority to work on.

garrettthomaskth:
Hi, I would also like to see this merged! :D

Neustradamus:
Any news?

DoZX:
Look forward to this feature.

PR #1026, opened 2020-12-18
[Do no merge] Support zstd compression

sergii-tsymbal-exa:
Current implementation is incorrect. Fails on real kafka


sergii-tsymbal-exa:
Does not work as expected

